PR #4, opened 2021-08-26
Add global header/footer block architecture

iandunn:
refactor of https://github.com/WordPress/wporg-repo-tools/pull/9

interconnected with https://github.com/WordPress/wporg-news-2021/pull/20

see https://github.com/WordPress/wporg-news-2021/issues/6

@coreymckrill , what do you think about this?

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace WordPress_org\MU_Plugins\Global_Header_Footer;
+namespace WordPressdotorg\MU_Plugins\Global_Header_Footer;
 
 defined( 'WPINC' ) || die();
 
@@ -12,12 +12,12 @@ function register_assets() {
 	// don't want this visible in Inserter. need to create ticket for that?
 
 	register_block_type(
-		'wordpress-org/global-header',
+		'wporg/global-header',
 		array( 'render_callback' => __NAMESPACE__ . '\render_global_header' )
 	);
 
 	register_block_type(
-		'wordpress-org/global-footer',
+		'wporg/global-footer',
 		array( 'render_callback' => __NAMESPACE__ . '\render_global_footer' )
 	);
 }
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace WordPress_org\MU_Plugins\Global_Header_Footer;
+
+defined( 'WPINC' ) || die();
+
+add_action( 'init', __NAMESPACE__ . '\register_assets', 9 );
+// why 9? if can be 10, then callers could be 9
+
+
+function register_assets() {
+	// don't want this visible in Inserter. need to create ticket for that?
+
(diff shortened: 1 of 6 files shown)


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -0,0 +1,63 @@
...
+
+	register_block_type(
+		'wordpress-org/global-header',
+		array( 'render_callback' => __NAMESPACE__ . '\render_global_header' )
+	);
+
+	register_block_type(
+		'wordpress-org/global-footer',
+		array( 'render_callback' => __NAMESPACE__ . '\render_global_footer' )
+	);
```
coreymckrill: I think ideally this would use `register_block_type_from_metadata` instead, and point to `block.json` files. That's the preferred method these days, I believe.

coreymckrill (approved):

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace WordPress_org\MU_Plugins\Global_Header_Footer;
```
coreymckrill: I've been trying to standardize on `WordPressdotorg` for the root namespace, because that's what I've seen in use in the codebase already. Does `WordPress_org` already exist too?

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -0,0 +1,63 @@
...
+
+add_action( 'init', __NAMESPACE__ . '\register_assets', 9 );
+// why 9? if can be 10, then callers could be 9
+
+
+function register_assets() {
+	// don't want this visible in Inserter. need to create ticket for that?
+
+	register_block_type(
+		'wordpress-org/global-header',
```
coreymckrill: Given that the text domain and the repo prefix for our WordPress.org stuff is `wporg`, it seems like it would be more consistent to use that for the block namespace too...

mu-plugins/blocks/global-header-footer/blocks.php
iandunn: No, I just think `WordPressdotorg` is bad because it all runs together. I'm fine using it for consistency though.

mu-plugins/blocks/global-header-footer/blocks.php
iandunn: I can see that for a static block, or even a more complex dynamic block.

For this, though, the `render_callback` would still have to be in `$args` AFAICT, and the name is the only param we'd use. Creating a separate file that'd only have 1 param in it seems kinda excessive in this context.

It's not a big deal if you feel strongly, though. If we do, it looks like [`register_block_type( __DIR__ )` is a more convenient wrapper around `register_block_type_from_metadata()`](https://make.wordpress.org/core/2021/06/23/block-api-enhancements-in-wordpress-5-8/).

coreymckrill: Ah good point, I didn't think about the fact that block.json probably doesn't even support the `render_callback` property. Probably not worth changing then, at this point.

iandunn: 👍🏻

PR #5, opened 2021-09-07
Refine header styles

iandunn:
see https://github.com/WordPress/wporg-news-2021/issues/6

interconnected with https://github.com/WordPress/wporg-news-2021/pull/28


iandunn:
This doesn't include the search or mobile menu, those will be in future PRs. It's also not trying to be pixel perfect yet, just getting the bulk of it done.

One problem I ran into is that adding Patterns and Openverse to the header made it too big to fit the design breakpoints. I tested out a submenu for the "resource" type links, and I think it works pretty well:

Screen Shot 2021-09-24 at 1 33 38 PM

I'm curious to hear what others think, though. cc @beafialho

beafialho:
> One problem I ran into is that adding Patterns and Openverse to the header made it too big to fit the design breakpoints. I tested out a submenu for the "resource" type links, and I think it works pretty well:

I like this solution @iandunn. Regarding the term "Add-ons", my only comment is I think it makes more sense to label it "Resources". The term Add-ons might be mistaken by some users not as familiar with WP as referring only to tools you can **add** to your site, whereas Resources feels broader with all the items you need to **build** your site.

iandunn:
That's a good point, we definitely want to communicate both _add_ and _build_.

Do you think people will make the connection between those actions and "Resources", though? If I saw it, I'd assume it refers more to things like documentation or support.

Is there a word that's more descriptive than both "Addons" and "Resources"?

beafialho:
> If I saw it, I'd assume it refers more to things like documentation or support.

That's also a good point. I've noticed that the current "Support" page links to Documentation, so maybe the label would be "Support" and Documentation, Forums and Guidelines would be its submenus?

There is also "Tools" instead of "Resources".

iandunn:
> maybe the label would be "Support" and Documentation, Forums and Guidelines would be its submenus?

I agree that'd be ideal. IIRC, though, the forums are intentionally a little hard to find, because the support team was overwhelmed by folks who didn't try to read docs first. The idea is to encourage folks to "self-serve" a bit before asking for other people's time.

> There is also "Tools" instead of "Resources".

🤔  I think that's a good one for plugins, but doesn't work as well for Themes, Patterns, Blocks, and Images/Openverse.

Maybe it'd be good to have a discussion on the designp2, and xpost to a few other teams, to get more perspectives on it? I think it'll be hard to come up with a good word with just a few people.

The News site isn't going to launch fo awhile, so we have plenty of time.

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -13,44 +13,108 @@
...
+
+	<!-- wp:image {"width":160,"height":24,"className":"site-header__wporg-logo-full"} -->
+	<figure class="wp-block-image is-resized site-header__wporg-logo-full">
+		<a href="https://wordpress.org/">
+			<img src="https://wordpress.org/style/images/wporg-logo.svg?3" alt="" width="160" height="24" />
+		</a>
+	</figure>
+	<!-- /wp:image -->
+
+	<!-- wp:group {"tagName":"div","className":"site-header__search-container"} -->
```
coreymckrill: I think you can omit `"tagName":"div"` since that's the default tag for a group block.

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -13,44 +13,108 @@
...
-<!-- /wp:button -->
-
-<!-- wp:button {"className":"site-header__get_wordpress"} -->
-<div class="wp-block-button site-header__get_wordpress"><a class="wp-block-button__link" href="https://wordpress.org/download/">Get WordPress</a></div>
-<!-- /wp:button --></div>
-<!-- /wp:buttons --></header>
-<!-- /wp:group -->
+<!-- wp:group {"tagName":"header","className":"site-header"} -->
+<header class="wp-block-group site-header">
+	<!-- wp:image {"width":27,"height":27,"className":"site-header__wporg-logo-mark"} -->
```
coreymckrill, line 19: Might be worth adding a comment here about why there are two logos, similar to the comment below about two Get WP buttons...

coreymckrill (approved):

mu-plugins/blocks/global-header-footer/header.php
iandunn: 👍🏻

mu-plugins/blocks/global-header-footer/header.php
iandunn: 👍🏻

iandunn:
i applied corey's suggestions in 2be4821

i went ahead and merged this w/ the add-on menu, b/c it's easier when we don't have lots of open interconnected prs. i'll  switch to the `...` overflow menu in the next pr.

PR #6, opened 2021-09-15
Header/Footer: Modularize CSS files and process with PostCSS.

iandunn:
while working on https://github.com/WordPress/wporg-news-2021/pull/28, the single `style.css` file has become unwieldy, and at this point the the benefits of preprocessing now outweigh the pain of tooling IMO.

it's easier to separate the files before that PR is merged

PostCSS has most of the features of SASS, but is more future-proof.


coreymckrill:
What do you think makes PostCSS more future-proof? From a package dependency perspective, it looks like this PostCSS config has 8 dependencies, whereas the Sass setup in `wporg-news-2021` has 2...

iandunn:
my understanding is that postcss is basically polyfilling future official css features, while sass is a custom thing.

e.g., w/ a sass plugin you can use variables inside media queries, but postcss instead [supports custom media queries](https://github.com/postcss/postcss-custom-media) 

also informed by this discussion: https://github.com/WordPress/gutenberg/issues/14801#issuecomment-480087247

ryelle:
passer-by comment: PostCSS is pretty minimal, and you only opt into the features you need, that's why there are more deps for a PostCSS config. Usually you can find PostCSS plugins following official spec proposals, so at some point when browsers support that spec, you could remove the plugin and everything will keep working.

If that's a feature you want, you'll need to check each plugin though - I see you've got `postcss-nested`, which follows how Sass does nesting, but isn't the spec version. The plugin using the actual spec format is [`postcss-nesting`](https://github.com/csstools/postcss-nesting).

It's also fairly straightforward to write small PostCSS plugins (maybe? I've written a few at this point so I'm probably biased), which can be used as helpers to do things like cleanup, format values, etc. Definitely an edge case but [it's come in handy before](https://github.com/ryelle/postcss-palettize-colors). Those can be in the same codebase as the config, and used without needing to release a npm plugin or anything.

iandunn:
great catch, switched to `postcss-nesting` in aca800f

coreymckrill (approved):
Thanks, this all makes sense. 👍 

At some point it might be worth migrating `wporg-news-2021` to PostCSS also, though I don't think it would be a high priority right now.

iandunn:
agreed, i opened WordPress/wporg-parent-2021#1 to track that, but am ambivalent about when to do it

PR #7, opened 2021-09-22
Header/Footer: Refine footer styles

iandunn:

Depends a little bit on https://github.com/WordPress/wporg-news-2021/pull/28, but should be mostly fine without

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/common.pcss
@@ -5,12 +5,47 @@
  * https://github.com/WordPress/wporg-news-2021/issues/13 isn't done yet.
  */
 
-.wp-block-group.site-header,
-.wp-block-group.site-footer,
-.wp-block-navigation:not(.has-background) .wp-block-navigation__responsive-container.is-menu-open {
-	background-color: var(--wp--preset--color--dark-strokes-grey);
+/*
+ * Variables, mixins, etc
+ *
+ * Breakpoints should be synced with `wporg-parent-2021` (or `wporg-news-2021` until parent exists).
+ */
+@custom-media --tablet (min-width: 876px);
+@custom-media --desktop (min-width: 1152px);
+@custom-media --desktop-wide (min-width: 1320px);
+
+:root {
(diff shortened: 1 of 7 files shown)


mu-plugins/blocks/global-header-footer/postcss/common.pcss
```diff
@@ -5,12 +5,47 @@
...
  */
 
+/*
+ * Variables, mixins, etc
+ *
+ * Breakpoints should be synced with `wporg-parent-2021` (or `wporg-news-2021` until parent exists).
+ */
+@custom-media --tablet (min-width: 876px);
+@custom-media --desktop (min-width: 1152px);
+@custom-media --desktop-wide (min-width: 1320px);
```
coreymckrill, line 15: One of the things that the News theme inherited from Blockbase was the usage of Gutenberg's [Base Styles](https://github.com/WordPress/gutenberg/tree/trunk/packages/base-styles) package. Specifically there are a few places we're using that package's defined [breakpoints](https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_breakpoints.scss#L6-L13). Not necessarily required here, but it might be nice to normalize our breakpoints one way or the other...

coreymckrill (approved):

mu-plugins/blocks/global-header-footer/postcss/common.pcss
iandunn: I agree, but it seems like the design has pretty specific requirements, at least for the header.

A "[media queryless](https://css-tricks.com/how-to-make-a-media-query-less-card-component/)" approach might be an alternative. I looked at that briefly, but IIRC it didn't seem practical in this case (or I just ran out of time/energy).

iandunn: I went ahead and merged this, because juggling multiple open/draft/local PRs is a pain, but I'm happy to continue discussing this and iterate.

iandunn: Maybe they should be renamed to `menu-mobile`, `menu-tablet`, etc, since they're built around that. Then we could use the base breakpoints for more generic stuff?

PR #8, opened 2021-09-29
Header: Enable responsive menu and style it.

iandunn:
Also creates the "overflow" menu, because the design doesn't want to wrap to a new line when there isn't room.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/header.php
+++ b/mu-plugins/blocks/global-header-footer/header.php
@@ -35,20 +35,26 @@ if ( ! $is_fse_theme ) {
 	<!-- wp:group {"className":"site-header__search-container"} -->
 	<div class="wp-block-group site-header__search-container">
 		<!-- wp:html -->
-		<a class="site-header__open-search" href="https://wordpress.org/search/">
-			<span class="screen-reader-text">Open Search</span>
-
+		<button
+			aria-haspopup="true"
+			aria-expanded="false"
+			aria-label="Open search"
+			class="site-header__open-search"
+		>
 			<img
 				src="<?php echo esc_url( plugins_url( '/images/search.svg', __FILE__ ) ); ?>"
 				alt=""
 				width="18"
 				height="17"
 			/>
-		</a>
+		</button>
 
-		<button class="site-header__close-search">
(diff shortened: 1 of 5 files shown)


coreymckrill (commented):
Kudos on the overflow menu, that looks like it was tricky to style.

Here are a few quirks I noticed in Firefox:
* Mobile menu icon only has two bars instead of three.
* When the responsive mobile menu is open, it covers up the rest of the header where the WP icon and search button are. Additionally the `X` icon to close the menu has some weird padding.
    * Maybe related to the mobile admin bar that gets pinned to the top when logged in?
* The search icon doesn't do anything, but I'm guessing that'll come in a separate PR?

iandunn:
> Mobile menu icon only has two bars instead of three.

That seemed weird to me too, but that's the image the Navigation block uses. I thought about overriding it w/ a custom one, but decided it wasn't worth it. Maybe we can circle back and do a "pixel perfect" sprint once the essential things are done?


> The search icon doesn't do anything, but I'm guessing that'll come in a separate PR?

Yeah

----

I'll look into the other comments, thanks!

iandunn:
> When the responsive mobile menu is open, it covers up the rest of the header ...  the X icon to close the menu has some weird padding.

🤔 

I haven't been able to reproduce either of those so far, in FF or Chrome. Which _theme_ branch are you testing it with?

coreymckrill:
Just tried it again in FF using the "iPhone 12 mini" preset for the viewport size. Latest trunk for the theme branch, latest commit for this PR branch. Now the `X` close icon isn't even visible.

| Menu closed | Menu open|
| --- | --- |
| Screen Shot 2021-10-01 at 14 30 31 | Screen Shot 2021-10-01 at 14 30 37 |

iandunn:
huh, i don't have that preset (ff dev edition `93.0b9 (64-bit)`). i tried w/ other iphone presets and still can't repro it 🤔

coreymckrill:
Oh, I'm on normal FF, not the dev edition...

iandunn:
I still can't reproduce that, so I opened #10 to dig into it more. I'll merge this so that the rest of it doesn't have to be juggled across active branches.

PR #9, opened 2021-09-29
Add functional search to header

iandunn:

Builds on top of https://github.com/WordPress/wporg-mu-plugins/pull/8 

Should roughly match the mockup, but not be pixel perfect. On mobile, click the icon to open menu; on desktop hover over it. searches should go to w.org/search to cover all sites.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/header.php
+++ b/mu-plugins/blocks/global-header-footer/header.php
@@ -32,40 +32,51 @@ if ( ! $is_fse_theme ) {
 	</figure>
 	<!-- /wp:image -->
 
-	<!-- wp:group {"className":"site-header__search-container"} -->
-	<div class="wp-block-group site-header__search-container">
-		<!-- wp:html -->
-		<button
-			aria-haspopup="true"
-			aria-expanded="false"
-			aria-label="Open search"
-			class="site-header__open-search"
-		>
-			<img
-				src="<?php echo esc_url( plugins_url( '/images/search.svg', __FILE__ ) ); ?>"
-				alt=""
-				width="18"
-				height="17"
-			/>
-		</button>
+	<!--
+		The search block is inside a navigation submenu, because that provides the exact functionality the design
+		calls for. It also provides a consistent experience with the primary navigation menu, with respect to
+		keyboard navigation, ARIA states, etc. It also saves having to write custom code for all the interactions.
+	-->
+	<!-- wp:navigation {"orientation":"vertical","className":"site-header__search","isResponsive":true} -->
+		<!-- wp:navigation-link {"label":"Search","url":"#","kind":"custom","isTopLevelLink":false} -->
+			<!-- wp:html -->
+			<!--
+				This markup is forked from the `wp:search` block. The only reason we're not using that, is because the
+				`action` URL can't be customized.
 
-		<button
-			aria-haspopup="false"
-			aria-expanded="false"
-			aria-label="Open search"
-			class="site-header__close-search"
-		>
-			<img
-				src="<?php echo esc_url( plugins_url( '/images/close.svg', __FILE__ ) ); ?>"
-				alt=""
-				width="21"
-				height="21"
-			/>
-		</button>
-		<!-- /wp:html -->
+				@link https://github.com/WordPress/gutenberg/issues/35572
 
-		<!-- wp:search {"className":"site-header__search-form","label":"Search","placeholder":"Search WordPress.org...","buttonText":"Submit search"} /-->
-	</div> <!-- /wp:group -->
+				The only things that changed were:
+
+				1) The `s` parameter is renamed to `search`, because `do-search.php` requires that.
+				2) The instance ID was changed to `99`, to make it likely to be unique.
+
+				If that issue is ever resolved, we should be able to replace this with the Search block, without having
+				to change any CSS.
+			-->
+			<form
+				role="search"
+				method="get"
+				action="https://wordpress.org/search/do-search.php"
+				class="wp-block-search__button-outside wp-block-search__text-button site-header__search-form wp-block-search"
+			>
+				<label for="wp-block-search__input-99" class="wp-block-search__label">Search</label>
+				<div class="wp-block-search__inside-wrapper">
+					<input
+						type="search"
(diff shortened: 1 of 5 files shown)


mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -67,6 +68,15 @@ class="site-header__close-search"
...
 
+	<!-- The search block is inside a navigation submenu, because that provides the exact functionality the design
+	     calls for. It also provides a consistent experience with the primary navigation menu, with respect to
+	     keyboard navigation, ARIA states, etc. -->
+	<!-- wp:navigation {"orientation":"horizontal","className":"site-header__search_TODO","isResponsive":true} -->
+		<!-- wp:navigation-link {"label":"Search Test","url":"#","kind":"custom","isTopLevelLink":false} -->
+			<!-- wp:search {"className":"site-header__search-form","label":"Search","placeholder":"Search WordPress.org...","buttonText":"Submit search"} /-->
+		<!-- /wp:navigation-link -->
+	<!-- /wp:navigation -->
+
```
iandunn, line 80: @coreymckrill what do you think of this? it's a bit unconventional, but seems better than writing all the custom interactions from scratch, or forking G components.

My only concern is that it could be a bit fragile. It would break if there's some point in the future where G decides to restrict the types of `innerblocks` that are allowed.

https://github.com/WordPress/gutenberg/blob/771b4564ebfdcdd2913b4e30e8d9d6aeaf607e67/packages/block-library/src/navigation-link/index.php#L229-L239

coreymckrill: 🤔 I agree it seems kind of fragile, but I understand the benefits you're describing.

Could it work if the search block was inside `wp:navigation` but not inside a `wp:navigation-link`? That seems like it would be more semantically acceptable, and less likely to run afoul of future `innerblocks` restrictions...

iandunn: I was wondering about that too, it's on my list of things to try out. I think the submenu will be needed for the desktop interaction, but it might work w/out it.

Worst case, this may be the easiest way to implement the design, and then if it breaks in the future we could easily change the search icon to a simple link to https://wordpress.org/search/ while we spend the time refactoring a permanent solution.

iandunn: It looks like removing the wrapping `wp:navigation-link` would work, but we'd have to add some significant styles to counteract what it's trying to do.

It feels like that'd be swimming upstream, and would make it _more_ fragile, rather than less. I think the likelihood of `innerblocks` being restricted for `navigation-link` but not `navigation` is low, so IMO it's best to keep it they way it currently is.

iandunn:
This is blocked on https://github.com/WordPress/gutenberg/issues/35572

If that gets traction, then I think we're good to go ahead w/ the `wp:search` block. If not, then I think we'll have to build a custom form.

mu-plugins/blocks/global-header-footer/header.php
iandunn: Doh, it doesn't look like I can change the "Open menu" `aria-label` to say "Open search"; at least not without an ugly hack :/

I think I'll look into reusing the lower level `micromodal` etc, but I'm worried that might be too fragile when things change in the future.

iandunn: Looks like there's already an issue for that: https://github.com/WordPress/gutenberg/issues/24369

So maybe we could go ahead with this approach, and then add the attributes once that issue is resolved? Or would that be too confusing in the meantime for folks using screen readers? 🤔

coreymckrill: > go ahead with this approach, and then add the attributes once that issue is resolved

This seems like a good strategy to me.

iandunn: 👍🏻

coreymckrill (commented):
Code looks fine, and well-documented.

Here are a few alignment issues I see in Firefox, running trunk for `wporg-news-2021` and this branch for `wporg-mu-plugins`. I realize this PR is not meant to be pixel-perfect, so feel free to disregard if these will be fixed later.

There's a very narrow window where the menu items wrap to a second line before they get hidden by the ellipses:

line-wrap

Also in that screenshot, you can see that the handle of the magnifier icon gets cut off.

On narrower screens, the right edge of the magnifier icon gets cut off:

trimmed-magnifier

When opening the mobile search menu item, the input label is doubled up:

double-search

iandunn:
> narrow window where the menu items wrap to a second line before they get hidden by the ellipses

Huh, I don't get that on mine. You've got 4bf2721f3dde054a86336cc9d0626505090ffa02 pulled, right?

If so, is increasing `@custom-media --desktop-wide` to `1450px` enough to account for whatever small variation is causing that?
<br/>


> the handle of the magnifier icon gets cut off.

I don't see it much on my machine, but d826d31 should hopefully fix it
<br/>


> opening the mobile search menu item, the input label is doubled up

huh, that's explicitly hidden, and it works on my system:

https://github.com/WordPress/wporg-mu-plugins/blob/d826d31c751c15f7805df2a37bfdf4b686918a9b/mu-plugins/blocks/global-header-footer/postcss/header/search.pcss#L31-L33

when you inspect that element, is the above rule absent? or is it overridden by something?

iandunn:
I'm gonna merge this to avoid juggling branches, but I'll iterate based on your responses to my previous comment.

coreymckrill:
> Huh, I don't get that on mine. You've got 4bf2721 pulled, right?

I've got the `search` branch checked out, which does appear to have that commit...

> If so, is increasing @custom-media --desktop-wide to 1450px enough to account for whatever small variation is causing that?

The Openverse link stops wrapping to a second line at 1478px.

> when you inspect that element, is the above rule absent? or is it overridden by something?

🤔 When I inspect that element, the classname is `.wp-block-navigation-link__label` rather than `.wp-block-navigation-item__label`. This is with WP 5.9-alpha-51906 and GB 11.7.0.

iandunn:
🤔  huh, I'll open a new issue and look into it next week 👍🏻

PR #20, opened 2021-11-30
Global nav refinements

MaggieCabrera:
This PR introduces refinements to the global header defined over at https://github.com/WordPress/wporg-news-2021/issues/86

So far this includes:

- [x] The active menu link (News in this case) should be **bold** and highlighted through color `#7B90FF`
- [x] The "Get WordPress" button should have a `2px` border radius and its text should be **bold**
- [x] Search icon should be centered to its container

**On mobile:**

- [x] Let's decrease the text size of the submenu items
- [x] "Close" icon should be inside its container
- [x] "Close" section and the menu's background should be a darker shade specified in Figma `#1C2024`

TO DO (these affect directly the block's functionality):

- [x] Arrows in Global Nav Bar should match with Local Nav Bar's
- [ ] The header responsiveness isn't ideal at the moment, because the links jump straight into the smallest version of the `…` nav, when they should go in there gradually. Here's what we could use to improve that: https://css-tricks.com/the-priority-navigation-pattern/

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
@@ -55,13 +55,16 @@
 
 			&.current-menu-item {
 				border-left: var(--active-menu-item-border-height) solid #7B90FF;
-				color: #7B90FF;
 				font-weight: bold;
 
 				@media (--tablet) {
 					padding-bottom: calc(37px - var(--active-menu-item-border-height));
 					border-bottom: var(--active-menu-item-border-height) solid #7B90FF;
 					border-left: none;
+				}
+
+				& a {
+					color: #7B90FF;
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


MaggieCabrera:
The last item will be addressed on a separate PR

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
```diff
@@ -109,11 +124,27 @@
 	& .wp-block-navigation-item {
 		padding-bottom: 0;
 	}
+
+	& .wp-block-navigation__submenu-icon {
+		& svg {
+			display: none;
+		}
+		background-image: url("data:image/svg+xml,%3Csvg width='10' height='7' viewBox='0 0 10 7' fill='none' stroke='white' stroke-width='1.2' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M0.416667 1.33325L5 5.49992L9.58331 1.33325'%3E%3C/path%3E%3C/svg%3E");
```
iandunn, line 132: What are the benefits of putting this inline? I don't like it at first glance because it's not obvious what the contents are, and it can't be reused, but I'm sure there are some benefits too.

PR #23, opened 2021-12-03
Footer font should be smaller

tellyworth:
Per figma. See https://github.com/WordPress/wporg-news-2021/issues/101#issuecomment-985065056


MaggieCabrera (approved):
Works for me!
Screenshot 2021-12-03 at 09 16 48

PR #27, opened 2021-12-08
Remove footer links as per design

scruffian:
As raised in https://github.com/WordPress/wporg-news-2021/issues/101 we should remove these links from the footer.
Screenshot 2021-12-08 at 09 22 32


danieldudzic:
Links seem to be correct. @mikachan and I spotted an issue with the margin between the links and the footer logos being different in Figma. Not sure if this deserves a separate Issue:

Screenshot_08_12_2021__13_57

ryelle (approved):

PR #30, opened 2021-12-08
Header Navigation: Update styles to match design

ryelle:
I think this is all that's left to wrap up https://github.com/WordPress/wporg-news-2021/issues/86, updating the submenu text size and spacing, and fixing the X close buttons on search & the menu.

| Before | After |
|--------|------|
| before-menu | after-menu |
| before-menu-with-ab | after-menu-with-ab |
| Screen Shot 2021-12-08 at 15 33 30 | after-search-with-ab |

PR #32, opened 2021-12-09
Navigation menu: Fix alignment, background color, and font size.

ryelle:
See https://github.com/WordPress/wporg-mu-plugins/pull/30#issuecomment-989728416, this should address all the listed issues. It also updates the font size on the large-screen header to match the design (well, it uses `--wp--preset--font-size--small`, which is 14px, not 13px — can use 13px directly if it's important).

> The X button should be centered to its container, it's slightly aligned to the left here

This is to account for the scrollbar, the whole nav is shifted slightly. I don't want to change this, since I don't fully understand why it was done this way.

small-closed
small-menu
small-search

large-closed
large-search

Screen Shot 2021-12-09 at 13 46 25

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -31,10 +31,12 @@
 		};
 
 		/**
-		 * Get the visible list items (filtering on display hides the extra "Get WordPress" link).
+		 * Get the top-level list items.
 		 */
 		this.getListItems = function () {
-			return this.wrapper.querySelectorAll( 'li:not(.global-header__mobile-get-wordpress)' );
+			return this.wrapper.querySelectorAll(
+				'.wp-block-navigation__container > li:not(.global-header__mobile-get-wordpress)'
+			);
 		};
 
 		/**
@@ -61,10 +63,9 @@
 		 * Hide menu items that exceed the container's width
 		 */
 		this.hideExtraItems = function () {
-			let totalWidth = 0;
 			//not pretty! but it's not likely to change
 			const dotMenuWidth = 83;
-			totalWidth += dotMenuWidth;
+			let totalWidth = dotMenuWidth;
 			this.resetMenu();
 
 			for ( let i = 0, len = this.itemsWidths.length; i < len; i++ ) {
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -3,147 +3,161 @@
  *
  * Applies a priority navigation pattern to the header menu.
  * https://css-tricks.com/the-priority-navigation-pattern/
- * 
+ *
  */
- ( function () {
-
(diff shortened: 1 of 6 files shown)


beafialho:
I may not have expressed myself correctly but the menu links on mobile should be `21px`. On desktop they should be `13px`.

@ryelle if you notice here, the dark grey is not quite the same in the icon and screen backgrounds. The correct color value is `#1C2024`

145456490-1881175f-0a27-402b-906c-3528a6a43c00
145456491-a8a1e47b-4a1c-4207-8d97-839e853131cf

ryelle:
> I may not have expressed myself correctly but the menu links on mobile should be 21px. On desktop they should be 13px.

No, I understood — but I'm using the existing custom properties, so it's 20px on mobile, and 14px on desktop. Are you asking me to change it to 21/13?

Fixed the icon background color (next time could you circle the issue in your screenshot? I usually keep my monitor brightness down, and really can't differentiate #1c2024 from black).

beafialho:
> No, I understood — but I'm using the existing custom properties, so it's 20px on mobile, and 14px on desktop. Are you asking me to change it to 21/13?

Yes, please.

> Fixed the icon background color (next time could you circle the issue in your screenshot? I usually keep my monitor brightness down, and really can't differentiate #1c2024 from black).

Sure, I can do that 👍

ryelle:
Fixed the font sizes on the menus.

small
Screen Shot 2021-12-10 at 11 28 16

I also fixed what I think was the issue mentioned in https://github.com/WordPress/wporg-news-2021/issues/120, 

> The rule should be: if there is enough space for all the menu links, it shouldn’t collapse into the ... Seems like the margin is too big on the right, it can be just 25px like in its left margin

The last item would still account for the space needed for the … menu item, and collapse if it's not there. But if the last item can fit, we won't need the … item. So now all the menu items can fit at 1280px wide, and it collapses down from there.

Screen Shot 2021-12-10 at 11 28 48

Screen Shot 2021-12-10 at 11 29 18

ryelle:
Actually, I had the script a little bit wrong still — it was also tallying the width of the submenu items, which added an extra gap at the end of the menu. Now the menu fills the whole space available, and only starts hiding elements at 1243px wide:

Screen Shot 2021-12-10 at 11 38 37

iandunn:
> This is to account for the scrollbar, the whole nav is shifted slightly. I don't want to change this, since I don't fully understand why it was done this way.

Yeah, it was because the scrollbar covers part of that space (at least in some browser/OS combinations), making it _look_ like it's not centered.

beafialho:
Thank you @ryelle! 

My only comment based on the screenshots: the font sizes for menu items on mobile are correct `21px`, but the submenus should be `15px` as specified in the mockups. This font size change should only affect the menu on mobile, not the desktop menu.

Captura de ecrã 2021-12-15, às 15 34 42

ryelle:
Updated the font size to 15px in https://github.com/WordPress/wporg-mu-plugins/commit/ea333c269b36494ec9f7e92b3598faccddf244e1:

beafialho:
LGTM 👍

PR #39, opened 2021-12-17
Remove lines.

tellyworth:
See https://github.com/WordPress/wporg-news-2021/issues/132

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
@@ -18,7 +18,6 @@
 	& .global-header__desktop-get-wordpress-container {
 		display: none;
 		order: 4;
-		border-left: 1px solid var(--wp--preset--color--darker-grey);
 		padding-left: var(--wp--style--block-gap);
 
 		@media (--tablet) {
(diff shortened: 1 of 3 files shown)


beafialho:
@tellyworth can you add a screenshot so I can review the changes?

iandunn (approved):
LGTM from a code perspective

tellyworth:
Before:
Screen Shot 2021-12-20 at 4 55 49 pm
After:
Screen Shot 2021-12-20 at 4 56 54 pm

I'm not sure what's up with the vertical alignment on my local install, but as you can see that's separate from this PR and not caused by it.

beafialho:
Thank you @tellyworth.

> I'm not sure what's up with the vertical alignment on my local install

Could it be the header's height, which should be `90px`? My only other comment is the word "Search" next to the magnifying glass icon shouldn't be there.

iandunn:
I don't see the alignment issue locally, so it may be something specific to your env, or it could be a specific browser etc. IMO it shouldn't block merging this, unless you wanna combine the two.

tellyworth:
I'm still getting a header that's taller (105.6px). But after some nuke-updating, here are cleaner screenshots.

Before:
Screen Shot 2021-12-21 at 6 41 22 pm

After:
Screen Shot 2021-12-21 at 6 42 17 pm

beafialho:
This looks good to me 👍

PR #61, opened 2022-01-07
Styles: Update google fonts to use API v2

ryelle:
The style URL for Open Sans is registered in WP core, but is not actually used there anymore. It's registered using the older, `fonts.googleapis.com/css` path, while the current version is `fonts.googleapis.com/css2`. See the info in [this update post](https://developers.google.com/fonts/docs/css2).

In updating the header, we could also update this API, which will give us better performance - at the very least by using `display=swap`. I've also added the preconnect resource hints for the header.

Note that if/when the whole of wp.org is using the new design, this will become obsolete, since the new design does not use Open Sans, and the fonts registered by `theme.json` already use this v2 API.

Fixes [#5610-meta](https://meta.trac.wordpress.org/ticket/5610).

To test:
- On any site, look for the Open Sans call in the source code or network tab, it should load from `css2`
- There should be no visual difference
- Load a few different locales, Open Sans should be loaded correctly for each subset (cyrillic, greek, etc)


iandunn (approved):
LGTM 👍🏻

Issue #42, opened 2021-12-22
Port old `header.php` features etc


iandunn:
port any needed "features" from old header

### `header.php`

- [x] skip to content links. maybe adjust to use [block theme standard](https://github.com/WordPress/gutenberg/blob/trunk/docs/how-to-guides/themes/block-theme-overview.md#accessibility). remove duplicate markup in themes that adds it (e.g., `wporg-learn-2020/header.php`).
- [x] ~`content-type` header?~ Taken care of by `WP::send_headers()`
- [x] `no-index` when accessed directly?
- [ ] ~text direction~ No, i18n menu, other i18n features
- [x] `<meta>` tags - do via hook so it affects classic and FSE
  - [x] `<meta name="viewport">`
  - [x] ~`<meta name="referrer" content="always">`~ Invalid Header.
  - [x] `<meta charset="utf-8" />`
- [x] `<link>` tags - do via hook so it affects classic and FSE
  - [x] `<link rel="shortcut icon">`
- [x] `<script>` tags - do via hook so it affects classic and FSE
  - [x] `body.no-js` => `body.js`
  - [x] ~Googletagmanager~ Handled in plugin.
  - [x] ~Always enqueue jQuery. Needed anywhere??~ No.
- [x] `<style>` tags - do via hook so it affects classic and FSE
  - [x] ~Survey Banner~ #93
- [x] codex-specific stuff?
- [x] `<body id="wordpress-org"` - is this needed? might be hacky to add to FSE body, so maybe port to class or something?
- [x] ~annual survey - it'd be nice to do this in a cleaner way, and maybe use native localization instead of hardcoding~ #93
- [x] header "options" like title , wrapper items, `$headadd`, etc. there may be a cleaner way to do it
  - [x] `$pagetitle` global in Themes, Rosetta, GlotPress.
  - [x] `$pagetitle` global in Planet
  - [x] `$headadd`
  - [x] `$wporg_global_header_options` (sans #93)
- [x] download mobile
- [x] google analytics
- [x] ~`X-Olaf`~ No.
- [x] ~rosetta `get_header_announcements()` ?~ #93
- [x] anything else from `load_custom_header()`?  (because it's no longer used)

### `functions-global.php`
- [x] capital P - is this for menu items that rosetta/etc admins create? might be cleaner to filter `wp_get_nav_menu_items()` or whatever, rather than buffering in this file
- [x] 301 redirect - still needed?
- [x] menu generation not needed anymore?
[...]

Issue #38, opened 2021-08-12
Implement global header/footer design across w.org network


iandunn:
After it's implemented for News in WordPress/wporg-news-2021#6, we'll need to roll the new header/footer out to all sites in the network. That can/should probably happen before News launches.

There are a lot of potential improvements we can make under the hood, to make the header easier to maintain across the various systems that implement it. It'd be nice to do at least some of that here, but we'll have to weigh the time/benefit tradeoffs for each one.

### Pain Points

> This file is used by wordpress.org static files including index-wporg.php. It is also used by the codex [and GlotPress, Trac, etc], based on a cron job set up to fetch header.php?wpcodex=1.
> 
> And, a lot of WordPress and bbPress themes include it.
> 
> Changes require global testing!

* It’s a pain to include the header in local environments, since you have to curl `w.org/header.php` instead of having a dynamic file. It's easier to fork it, but those quickly become out of date.
* Changes have to be manually copied to the hardcoded Genshi templates for Trac
* It’s hard to test updates to Trac and the Codex before deploying.
* Testing the changes across all sites is tedious, and it’s easy to miss something
* `wp4.css` is massive, and changes have unintended consequences
* footer strings have to be copied to `pub/wporg/footer.php` for localization

### Potential solutions

We could:

- [x] Open source the files. We should do this if nothing else. Maybe make the global header a separate repo with completely self-contained styles, or just commit to `meta.svn`
- [x] I think the styles would be easier to maintain, and less fragile, if they were component-ized in the new design, rather than remaining a part of the wp4.css maze. That could still be used by the other themes/systems until they’re migrated to the new parent. We could remove the header/footer styles from `wp4.css` and move them to some shareable component.
- [x] create a REST API endpoint to get the HTML for the header/footer. that seems like a more modern version of how some sites will `curl https://wordpress.org/header.php` to import it.
- [ ] To make global testing easier and more reliable, we could setup e2e screenshot comparisons or something. They might have to be post-deploy given the lack of dev environments for Trac etc, though.
- [x] Is there a reason Trac doesn’t `<iframe>` it, or curl-and-cache it like the Codex? Could the Codex `<iframe>` it for simplicity? probably because submenus/modals can't expand beyond fixed height of the iframe

### Refine existing code

- [x] #44 - high
- [x] #29  - high
- [x] #46 - high
- [x] #35 - high
- [x] #52 - high
- [x] #42 - medium
- [x] #22 - medium
- [x] #57 - medium
[...]

PR #62, opened 2022-01-10
Add Codex/Planet required bits to header

dd32:
See #42 & #38

Unfortunately the Codex requires wp4 styles and isn't super pretty with just removing that and dropping in the new header/footer.

as it is with the new header/footer
no_styles

as it is with wp4 sttyles after https://github.com/WordPress/wporg-mu-plugins/commit/cc68eb9886612a24917a35b35fb91d629af30755 - Just adding codex-wp4 didn't help here either.
with_styles

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -618,7 +618,11 @@ function get_download_url() {
  */
 function rest_render_global_footer( $request ) {
 
-	$markup = render_global_footer();
+	/*
+	 * Render the header but discard the markup, so that any header styles/scripts
+	 * required are then available for output in the footer.
+	 */
+	render_global_header();
 
 	// Serve the request as HTML
 	add_filter( 'rest_pre_serve_request', function( $served, $result ) {
@@ -629,7 +633,7 @@ function rest_render_global_footer( $request ) {
 		return true;
 	}, 10, 2 );
 
-	return $markup;
+	return render_global_footer();
 }
 
 /**
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -88,6 +88,18 @@ function register_routes() {
 
 	register_rest_route(
 		'global-header-footer/v1',
+		'header/planet',
+		array(
+			array(
+				'methods'  => WP_REST_Server::READABLE,
+				'callback' => __NAMESPACE__ . '\rest_render_planet_global_header',
+				'permission_callback' => '__return_true',
+			),
+		)
+	);
+
+	register_rest_route(
+		'global-header-footer/v1',
 		'footer',
 		array(
 			array(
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -183,6 +195,60 @@ function restore_inner_group_container() {
...
+
+	$markup = render_global_header();
+
+	// Serve the request as HTML
+	add_filter( 'rest_pre_serve_request', function( $served, $result ) {
+		header( 'Content-Type: text/html' );
+
+		echo $result->get_data();
+
+		return $served = true;
```
iandunn: Why this instead of just `return true` ?

Won't `$foo = true` always evaluate to `true`?

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -183,6 +195,60 @@ function restore_inner_group_container() {
...
+	}, 100 );
+
+	add_filter( 'body_class', function( $class ) {
+		return [
+			'wporg-responsive',
+			'wporg-codex'
+		];
+	} );
+
+	wp_enqueue_style( 'wp4', 'https://s.w.org/style/wp4.css', array(), 95 );
```
iandunn: I think this is unnecessary after https://github.com/WordPress/wporg-mu-plugins/commit/5cde8ecadd5bfb9a53b1c3cc5e2b7ef6f5c52132

iandunn (approved):
Left a few small things, but overall LGTM 👍🏻

iandunn:
Er, I meant the code looks good, but if the screenshots are current then of course we'll need to do more to fix the styles

mu-plugins/blocks/global-header-footer/blocks.php
dd32: Yes. It's more a coding style I've been in the habit of in filters to make it clear to myself that I'm handling the filter correctly.

Happy to switch to return true directly.

```suggestion
		return true;
```

iandunn: Ah, I see 👍🏻  I'm not sure it'll be obvious to others, but I don't feel strongly

dd32: GitHub doesn't show it clearly in this thread view, but I committed the above suggestion shortly after commenting. https://github.com/WordPress/wporg-mu-plugins/pull/62/commits/d70f783ed1f91723f332a4c00c962fa724bc24fa

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -513,6 +573,27 @@ function get_download_url() {
...
+
+	$markup = render_global_footer();
+
+	// Serve the request as HTML
+	add_filter( 'rest_pre_serve_request', function( $served, $result ) {
+		header( 'Content-Type: text/html' );
+
+		echo $result->get_data();
+
+		return $served = true;
```
dd32:
```suggestion
		return true;
```

mu-plugins/blocks/global-header-footer/blocks.php
dd32: That's now handled properly after 9c4143cb0671d583b3cdc55a2a4355a196799730 - https://github.com/WordPress/wporg-mu-plugins/commit/5cde8ecadd5bfb9a53b1c3cc5e2b7ef6f5c52132 wasn't adding the support during rest requests.

dd32:
This has ended up with some code that I don't like in it.

https://github.com/WordPress/wporg-mu-plugins/pull/62/files#diff-460e50b82495cab49b3ce3a702572bec919fae2cbec691937f8abbe7de6e2c33R255-R262

As we're loading `wordpress.org/wp-json` the `pub/wporg-main` theme is loaded, that brings with it a bunch of styles that don't make sense to load (and break layouts).

There's also this:
https://github.com/WordPress/wporg-mu-plugins/pull/62/files#diff-460e50b82495cab49b3ce3a702572bec919fae2cbec691937f8abbe7de6e2c33R180-R183

it feels like this is uglier than it needs to be, and that maybe we can have these endpoints served by a site with no theme.

dd32:
Codex:
screencapture-planet-wordpress-org-Main-Page-2022-01-13-15_34_05
screencapture-planet-wordpress-org-Main-Page-2022-01-13-15_33_49

Planet:
Screen Shot 2022-01-13 at 3 20 06 pm

Screen Shot 2022-01-13 at 3 21 08 pm

iandunn:
>  have these endpoints served by a site with no theme

That sounds like it could work, if it won't remove some of the scripts/styles that we _do_ want installed.

Issue #35, opened 2021-12-16
Assign `active-menu-item` dynamically


iandunn:
part of #38

right now it's hardcoded, but we'll need it to be set based on the current site

my first thought is something like this:

```php
// in `render_global_header()`, after all menu items have been added
$menu_items = set_current_item_class( $menu_items );
```

...and then that function would do something like:

```php
foreach ( $menu_items as $item ) {
	if ( $current_url === $item->url ) {
		$item->classes[] = 'current-menu-item';
	}
}
```

The logic will need to be more complex than that, to account for child pages etc. There may be some existing Core or `functions-global.php` code we can adapt.

dd32:
Fixed via #68

PR #68, opened 2022-01-11
Dynamically assign `active-menu-item` to the menu

dd32:
This is probably not the most efficient way, but is more dynamic than what we currently use on WordPress.org (which is mostly a tightly coupled if-branching setup).

Some style changes were required to make submenu active highlighting work more correctly, see below screenshots of the after effect.

https://user-images.githubusercontent.com/767313/148885968-309a0e84-397a-415a-8755-2a3860002911.mov

And for the reasoning for the CSS rotation:

Screen Shot 2022-01-11 at 3 20 20 pm

See #35

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -624,8 +624,30 @@ function get_menu_url_for_current_page( $menu_items ) {
 		return 'https://make.wordpress.org/';
 	}
 
+	if ( 'developer.wordpress.prg' === $_SERVER['HTTP_HOST'] ) {
+		// DevHub doesn't exist within the menu.
+		return '';
+	}
+
+	// Temporary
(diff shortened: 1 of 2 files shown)


iandunn (approved):
This looks good 👍🏻 

The only thing I might change is that `get_menu_url_for_current_page()` feels a bit complicated, but I don't have any ideas for simplifying it. So I'd go ahead and merge 👍🏻

dd32:
> The only thing I might change is that `get_menu_url_for_current_page()` feels a bit complicated

@iandunn I also felt that, which is why I extracted the logic from `set_current_item_class()` directly to start with (and it was simpler then..).

I'd love to simplify this in future, I just wasn't able to figure out anything more reliable given the menu doesn't know if it's pointing to `get_permalink()`, `home_url('/')`, or a combination of them. Might be able to add some flags into the menu array to start with, but then Rosetta dynamic menu's kicks in and throws a spanner in there :)

iandunn:
👍🏻

Issue #3, opened 2021-08-24
Should we add existing open-source mu-plugins from `meta.svn`?


iandunn:
_cc @WordPress/wordpress-org-meta-team for input. See https://github.com/WordPress/wporg-repo-tools/pull/9 for the origins of this repo_

----

🤔 , should we add [the existing open-source mu-plugins from `meta.svn`](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub) here?

If we don't, then it seems odd to have some of them in one place, and others in a different place. 

But if we do, then all Meta committers will need to use the GitHub -> SVN workflow; if not, things would get out of sync, and we might accidentally overwrite things.

I guess that problem could happen with either approach. The sync script can bail if it detects that situation, but there'll still be manual work necessary to clean it up.

Maybe we should wait until _all_ of `meta.svn/sites/trunk/wordpress.org` migrates to GH? If so, maybe now is a good time to do that? Scope creep, though...

coreymckrill:
Per [this comment](https://github.com/WordPress/wporg-repo-tools/pull/9#issuecomment-904074604), I was thinking that we'd gradually migrate the _non_ open-source mu-plugins from `dotorg.svn` here, as we determine that they are safe to open source. The ones in `pub` from `meta.svn` would probably be my lowest priority, since they're already open.

Ideally, I think, we eventually wouldn't need a `pub` folder at all within `mu-plugins` (maybe a `private` instead with a few things that can't be open sourced?) But that would require moving stuff out of `meta.svn` and then that would potentially break some svn externals...

PR #70, opened 2022-01-11
Add a Skip-to mu-plugin

dd32:
This PR adds a skip-to function based on the WordPress block theme skip-to link, but targeted towards classic themes.

This allows us to remove some of the deprecated header functions.

See #42 and https://github.com/WordPress/wordpress.org/pull/49 for an implementation example


iandunn (commented):
It'd be simpler if we could do something like `echo do_blocks( '<!-- wp:skip-to ...`, but I'm guessing there's a reason why that doesn't work in this context?

skip-to/skip-to.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	function skip_tag( $selector ) {
+		$target = ( '#' === substr( $selector, 0, 1 ) ? $selector : '' );
+		$tabindex = ( $target ? '' : 'tabindex="-1"' ); // Will be removed once the target is set.
+		printf(
+			'<a id="wporg-skip-link" %s class="skip-link screen-reader-text" href="%s" data-selector="%s">%s</a>'."\n",
+			$tabindex,
+			esc_attr( $target ),
+			esc_attr( $selector ),
+			__( 'Skip to content', 'wporg' )
```
ryelle: Since the `selector` can be customized, is there ever a case where this might not skip to content, but to another section? Or there might be more than one skip on a page? If so, this text should also be a parameter so it can be controlled.

dd32: The customization of the skip destination was primarily to avoid having to change all of the existing themes structure, while removing the header dependancies.

I considered that, and decided against supporting multiple skip-links per page, for this at least. Similarly, it looks like the Core skip functionality also doesn't support multiple skiplinks, only linking to the singular individual `<main>` element.

In https://github.com/WordPress/wordpress.org/pull/49/files#diff-b468acd1a227cd79ee142be67224a169761ccddb47f283c5d384523d67fb0c1a you can see an example of where there were previously two skip to links, that I've condensed to one.

In that case, the links didn't make sense as they were.
 - Start of document -> Site Header
 - Line before Site Header -> Content

In that case, you'd have to `Tab, Select (Skiplink #1, go to site header), Shift+Tab (Back one), Select (Skiplink #2, go to content)`. The only way to reach the second one was to go through the full header, and then skip over the much smaller "site header".

dd32:
> It'd be simpler if we could do something like `echo do_blocks( '<!-- wp:skip-to ...`

It would be. However, as far as I know, there isn't a block for this.

Instead `the_block_template_skip_link()` is hooked directly to wp_footer():
https://github.com/WordPress/wordpress-develop/blob/f7d2a2ee9d003633e6c729c0835cd4addd23f9b3/src/wp-includes/theme-templates.php#L101-L205

It works fine for it's intended use-case, block themes, but isn't written in a way to support classic themes on w.org.

Even if I could convince WordPress to output the JS/CSS it generates into all themes, it wouldn't help for the myriad of classic themes that are not built using `<main>` as their content tag. 

This isn't intended on being a "final solution", it's a stop-gap for WordPress.org specifically until all themes are using `<main>` and are block-themes, so that we don't need to alter all themes structures just to swap the header.

skip-to/skip-to.php
ryelle: > The customization of the skip destination was primarily to avoid having to change all of the existing themes structure, while removing the header dependancies.

Makes sense 👍🏻 

> In that case, the links didn't make sense as they were.

Agreed, I'd call that a bug you fixed 🙂

iandunn:
That makes sense 👍🏻

skip-to/skip-to.php
```diff
@@ -0,0 +1,141 @@
+<?php
```
ryelle: I think this folder should be in the [`mu-plugins` folder](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/), if I'm understanding the repo structure correctly

ryelle:
I tested this on Learn with part of https://github.com/WordPress/wordpress.org/pull/49 applied, and it worked great. I just have the one question/comment about file structure, but once the folder is moved I think this can merge  👍🏻 

It'll also need to be added to `_loader.php` when syncing to SVN.

skip-to/skip-to.php
iandunn: Good catch, yeah, `mu-plugins` is what's synced to `pub-sync` on dotorg.

dd32: Ah... Yeah good catch!

The structure of this repo doesn't allow this repo to be developed on sandboxes, so I screwed up in the copy-to-git process.

iandunn: yeah, that is annoying 🙁 

related: https://github.com/WordPress/wporg-mu-plugins/issues/3

Issue #57, opened 2022-01-05
Mobile menu and search: inconsistent behaviour and visual refinement


beafialho:
Currently there are a few inconsistencies with the mobile menu that are critical for launch:

https://user-images.githubusercontent.com/65220155/148218743-fc3b0f9a-2544-467a-abe1-baf05bdcf07f.mp4

- [ ] If I have the menu open and click the Search icon, the search field doesn't show up. The way it should behave is: if I click a button, it should override the other.

- [ ] When the menu is open, its icon should then be replaced by the `Close x` icon.

Here's the intended behaviour:

https://user-images.githubusercontent.com/65220155/148219311-6ff042cf-0190-44d3-bf80-75e788901766.mp4

- [x] The search field should have `2px` radius

- [x] On Desktop, there's a tiny space between the icon and the search box that shouldn't be there:
<img width="375" alt="Captura de ecrã 2022-01-05, às 12 43 14" src="https://user-images.githubusercontent.com/65220155/148219501-dacb1c37-352b-4cba-bbce-5d7fbfe4881f.png">

iandunn:
> have the menu open and click the Search icon

👍🏻   we probably need to add some JS to hook into the open actions for both menus, and close the other one. I'm not sure if Gutenberg provides hooks for those, but hopefully.


> ​The search field should have 2px radius

Do you mean a `border-radius` on the `input` field? That seems like a valid change, but I don't understand why it'd be critical for the launch. Can you help me understand that?

beafialho:
> Do you mean a border-radius on the input field? That seems like a valid change, but I don't understand why it'd be critical for the launch. Can you help me understand that?

That's right, a border-radius on the input field is critical. In a redesign [details matter](https://uxmyths.com/post/712451083/myth-if-your-design-is-good-small-details-dont-matter), otherwise it looks unfinished and unpolished.

iandunn:
I agree that *some* small things matter, but more so with UX impacts like that article cites. 

If your standard for "critical" includes having everything *visually* polished, that works for me, we just need to pick a launch date that reflects that.

kellychoffman:
Bringing in additional designers to address this.

ryelle:
> If I have the menu open and click the Search icon, the search field doesn't show up. The way it should behave is: if I click a button, it should override the other.

[...]

Issue #74, opened 2022-01-11
mobile, patterns/learn: Header & footer use Open Sans instead of Inter


iandunn:
mobile, patterns/learn: search "search" font and spacing different from news-test.
maybe classic theme styles overriding?
maybe header styles added to news theme instead of this repo? font looks like being overridden by learn theme b/c targets any `label` tag, probably update learn css to be more targeted

ryelle:
The root styles to use Inter are set in News, but are not carried over to this project. (I don't know why I keep thinking it's system fonts 😅 )

ryelle:
Spacing will be fixed by #79

PR #79, opened 2022-01-12
Improve visuals for search in the header block

### LINKED ISSUE #72: desktop, patterns: hover on search icon, blue line between menu and search box. happens on most classic themes, if not all

### LINKED ISSUE #183: There are two magnifying glass icons in the `Search` field, there should only be one on the right with `29px` padding
| Test Site | Mockup |
| --- | ----------- |
|  IMG_8738 |  IMG_8745 |

MaggieCabrera:
I started working on the search on mobile to fix the Safari issue but I ended up tweaking some of the styles to bring it closer to the designs. I hope that's ok!

This is how it looks now on desktop:

Site | Figma
--- | ---
Screenshot 2022-01-12 at 10 41 09 | Screenshot 2022-01-12 at 10 20 55

And on mobile (Figma on the left for reference):

Screenshot 2022-01-12 at 10 46 25

This closes https://github.com/WordPress/wporg-news-2021/issues/183 and resolves a couple of the issues in https://github.com/WordPress/wporg-mu-plugins/issues/57:

- [x] The search field should have 2px radius
- [x] On Desktop, there's a tiny space between the icon and the search box that shouldn't be there:


beafialho:
Thanks @MaggieCabrera!

Could you please check that the color of "Search WP.org..." text in the search bar is `#6A6A6A` to make sure it has enough contrast for accessibility? Seems slightly lighter on mobile in the screenshot.

MaggieCabrera:
@beafialho the color for the text on the input is set to #23282d but the browser usually has a little lower opacity when the text is a placeholder. Can you give me the colors for both the text and the placeholder that you'd want?

beafialho:
Sure, you can see the colors [here](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=4659%3A22595).

The text on the input should be `#6A6A6A` but I'm not sure what other text color you're referring to?

MaggieCabrera:
So, the text that you see `Search WP.org...` is the placeholder, has lower opacity than when the user actually inserts a text into the input. The opacity of the placeholder is defined by the browser (not all do it the same way, the gif shows FF) unless we define it ourselves, but it's usually different so the user knows if the text had been inserted by the user or not:
Screen Capture on 2022-01-12 at 11-27-37

beafialho:
Ahh of course, I wasn't getting there. Thanks for clarifying. The placeholder text should be `#6A6A6A` and the input text should be `#1E1E1E`.

Other than this, it all looks good to me 👍

MaggieCabrera:
This is how it looks like now:
Screen Capture on 2022-01-12 at 11-41-04

Thanks for the feedback @beafialho !

beafialho:
Looks perfect, thank you!

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -114,7 +114,7 @@ class="wp-block-search__button-outside wp-block-search__text-button global-heade
 						class="wp-block-search__input"
 						name="search"
 						value=""
-						placeholder="<?php echo esc_attr_x( 'Search WordPress.org...', 'input field placeholder', 'wporg' ); ?>"
+						placeholder="<?php echo esc_attr_x( 'Search WP.org...', 'input field placeholder', 'wporg' ); ?>"
```
iandunn, line 117: @beafialho , is this to shorten the length of the placeholder, or something else?

It seems incongruent from a user perspective, since only insiders are familiar with that domain alias.

mu-plugins/blocks/global-header-footer/postcss/header/search.pcss
```diff
@@ -91,27 +96,49 @@
...
 			width: 100%;
-			padding-right: 35px; /* Prevent user input from running into icon. */
+			padding-right: 50px; /* Prevent user input from running into icon. */
 			border: none;
+			border-radius: 2px;
 			background-color: var(--wp--preset--color--white);
-			color: var(--wp--preset--color--dark-grey);
+			color: var(--wp--preset--color--black);
 			font-size: inherit;
+			-webkit-appearance: none;
```
iandunn, line 107: Is this what gets rid of Safari-mobile's extra glass icon? If so, a comment seems useful to make that obvious.

iandunn (approved):
LGTM 👍🏻 

Left a few questions, but they're not blockers

mu-plugins/blocks/global-header-footer/header.php
beafialho: @iandunn it is. I don't feel strongly about it though. We can leave `Search WordPress.org` as placeholder text.

ryelle:
I've pushed a few changes to sync the styles on the classic themes better, an example using Learn:

| Before | After |
|--------|------|
| mobile-search-learn-before | mobile-search-learn |

I'm not totally sure why the search bar won't go full width - and why fixing the font made it smaller - but I don't think we need to solve that in this PR (unless it's obvious to someone else 🙂 )

For comparison, News:
mobile-search-news

ryelle:
I reverted the font change, tracking that in #74

ryelle:
Okay, after reverting the font change and merging in `trunk`, the search bar is the right size in both classic and block themes. 🤷🏻 

I think this is good to merge now.

PR #80, opened 2022-01-12
Header: Output just the variables generated by `theme.json` from the News site

### LINKED ISSUE #66: showcase content styles messed up
The only issue I can see is the site title font:

Sandbox:
Screen Shot 2022-01-10 at 4 15 41 PM

Production"
Screen Shot 2022-01-10 at 4 15 19 PM

ryelle:
Updates the classic header to output the custom properties set by the News theme. This removes the network option method, in favor of directly calling `wp_get_global_stylesheet` on News.

It likely still needs to be cached, but I wanted a gut-check on the approach first :)

There's a workaround for the `--wp--style--block-gap` property, since it's not considered part of the `variables` set, so instead I'm setting it directly. If that changes for some reason, it'll need to be changed here too.

Fixes #66 

To test:

- Needs to be tested on a sandbox
- Visit a classic theme site, ex: [wordpress.org/showcase/](wordpress.org/showcase/)
- In this site, the `h3`s should have the correct (sans-serif) font
- View source, find the style tag `id=global-styles-for-classic-themes`, it should only have custom properties on `body`, no other styles


iandunn (approved):
I like this approach, kudos!

Assuming it doesn't cause side-effects on other sites, this LGTM.

It feels like caching may be premature (or at least a post-launch iteration), but that's just a hunch.

ryelle:
> Assuming it doesn't cause side-effects on other sites, this LGTM.

The news site won't cache this result since we're passing a type to `wp_get_global_stylesheet`, and clearing out the cached data should fix any local site side effects. and I didn't see anything in my testing.

> It feels like caching may be premature (or at least a post-launch iteration), but that's just a hunch.

Yeah I'm not sure, which is why I put this out here before spending that time. It's been a while since I've had to think about internals performance, but I thought using switch_to_blog was a big deal. I'm happy to merge this as-is, and see if we need it later.

iandunn:
👍🏻

Issue #63, opened 2022-01-10
Global Header: hover effects


beafialho:
The global header items currently have just an horizontal line separating them from the submenus. Both the hovered link as well as its submenu should have a darker background color `#1C2024`, as below:

<img width="965" alt="Captura de ecrã 2022-01-10, às 10 43 39" src="https://user-images.githubusercontent.com/65220155/148753386-7c327628-4607-4dfc-a33e-fac2cfe4e6ea.png">

ryelle:
@beafialho Is there a hover effect for the submenu items?

beafialho:
@ryelle yes, submenu links should be underlined.

ryelle:
@beafialho Should the main menu items also be underlined, or just the submenu items?

ryelle:
These should all be fixed now 👍🏻

beafialho:
@ryelle only the submenu items should be underlined. The main menu items should have a darker background, like this:

<img width="1206" alt="Captura de ecrã 2022-01-17, às 17 38 15" src="https://user-images.githubusercontent.com/65220155/149815892-aa5fb4e9-91b8-468e-8db2-55c449ac64e4.png">

PR #90, opened 2022-01-14
Header menu: Fixes for minor style issues

ryelle:
I noticed a few issues— the chevron was not changing color with the current menu item, and it was not bold on the `make.w.org` homepage. Also on the page, the logo had too much space next to it. In other themes, there's a `box-sizing` reset applied to fix this, so I've added it to the children in header & footer.

Before
before

After
after

This PR also adds the underline to submenu items, see https://github.com/WordPress/wporg-mu-plugins/issues/63#issuecomment-1013379510

Screen Shot 2022-01-14 at 3 35 17 PM


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -837,4 +838,14 @@ function( $val, $key ) use ( &$urls ) {
...
 	return home_url('/');
-}
\ No newline at end of file
+}
+
+/**
+ * Replace the current submenu down-arrow with a custom icon.
+ *
+ * @param string $block_content The block content about to be appended.
+ * @return string IDK
```
iandunn: What does `IDK` mean?

iandunn (approved):

mu-plugins/blocks/global-header-footer/blocks.php
ryelle: 🙃  I forgot to update it

Issue #12, opened 2021-10-14
Customize search `aria-label`


iandunn:
The search "menu" needs it's `aria-label`s customized (e.g., "Open search" rather than "Open menu"), but that can't be done until https://github.com/WordPress/gutenberg/issues/24369 is resolved.

xref https://github.com/WordPress/wporg-mu-plugins/pull/9/commits/6779dd78a7f38ba93df74bb6a4f79ffc5547921e#r728528115

ryelle:
This was done via JS in https://github.com/WordPress/wporg-mu-plugins/pull/119, so if we eventually switch over to a core method, that JS can be removed too.

Or we can consider this done and close the issue.

PR #95, opened 2022-01-17
Header search: Fix extra link in search dropdown

### LINKED ISSUE #94: Search: focus outline visible around invisible element on small screens

ryelle:
The search dropdown uses a navigation menu for functionality, which requires adding an extra link "Search" in the menu. When this is converted to the mobile menu, the link (`a` tag) still exists and takes up space, but the contents are invisible. On some sites, where a focus outline has been set, this creates a "focus state on nothing".

This PR hides the whole link tag in mobile-mode, since it's not needed as a label or for functionality. On larger screens, I've switched the label to be visually hidden, not `display: none`, so that the 🔎  icon has a label for screen reader users (this is not a fix for #12, that still requires GB, but it is a decent workaround).

Lastly, I wrapped the `navigation-link`'s `label` in a translation wrapper, since I don't think this will be picked up otherwise.

Fixes #94.

| Before | After |
|-----|----|
| menu-closed | pr-menu-closed |
| menu-open | pr-menu-open |
| mobile-closed | pr-mobile-closed |
| mobile-open | pr-mobile-open |

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/search.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/search.pcss
@@ -25,7 +25,18 @@
 	}
 
 	& .wp-block-navigation-item__label {
-		display: none;
+
+		/* See @mixin hide-accessibly in wporg-news-2021 */
+		border: 0;
+		clip: rect(1px, 1px, 1px, 1px);
+		clip-path: inset(50%);
+		height: 1px;
+		margin: -1px;
+		overflow: hidden;
+		padding: 0;
+		position: absolute;
+		width: 1px;
+		word-wrap: normal !important;
 	}
 
 	& .wp-block-navigation-item__content {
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/global-header-footer/postcss/header/search.pcss
```diff
@@ -25,11 +25,27 @@
 	}
 
 	& .wp-block-navigation-item__label {
-		display: none;
+
+		/* See @mixin hide-accessibly in wporg-news-2021 */
```
iandunn, line 29: It might be helpful to make it a mixin here too, so we can reuse it, and it's obvious which rules are part of it.

Not a big deal, though.

iandunn (approved):
LGTM 👍🏻

mu-plugins/blocks/global-header-footer/postcss/header/search.pcss
ryelle: Do we have a mixin plugin for postcss?

iandunn: I don't think so. There's https://github.com/postcss/postcss-mixins, but it's not a big deal if you don't think it's worth it

ryelle: Yeah, I don't think we need to add one. In most cases we're able to pass the core `screen-reader-text` as a className to the block, that just wouldn't work in this case.

PR #96, opened 2022-01-17
Header: Stick header to top of page while scrolling

### LINKED ISSUE #67: Make header stick to top on all sites
right now it's only being done by the news theme, but the global header should have it even if there isn't a local header, or a container around the global header. 

maybe duplicating the markup should be moved from the theme templates to this repo, to make sure it's done consistently, and to make the templates less cluttered. cc @shaunandrews , since it sounds like you may have an alternative to double-markup hack

ryelle:
Fixes #67. This uses `position: sticky` to attach the header to the top of the viewport (below the admin bar), so it stays in place while scrolling down the page.

No changes are needed on the news theme— the parent container is sticky there so that both the global and local nav are sticky.

I tested this on my sandbox on a few sites. Mostly this change was smooth, no visual changes (except now the header moves with you). I only had to fix two issues:

- A `z-index` issue with the showcase's `leftsidebar`
- A container height issue on translate.w.org, where `html { height: 100%; }` was set, which caused the header to scroll offscreen at that arbitrary point

I'd appreciate at least one other look-over from a dev with a sandbox- I've checked a handful of wporg sites, but i'm sure there are other edge cases.

**Screenshots**

https://user-images.githubusercontent.com/541093/149838664-91557f3b-c6b2-4fba-8253-5be990ec3304.mov

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/common.pcss
@@ -14,6 +14,11 @@
 @custom-media --desktop (min-width: 1152px);
 @custom-media --desktop-wide (min-width: 1440px);
 
+/* Offset from the top of the page which is covered by "stuck" items (admin bar & header). */
+html {
+	scroll-padding-top: calc(105px + var(--wp-admin--admin-bar--height, 0px)); /* 105px = height of header. */
+}
+
 /*
  * Shared styles
  */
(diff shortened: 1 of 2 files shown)


dd32:
> I'd appreciate at least one other look-over from a dev with a sandbox- I've checked a handful of wporg sites, but i'm sure there are other edge cases.

I haven't found any issues so far. I've additionally tested it with Trac, Planet, and Codex which were all fine. Even the Openverse iframe seems to still work.

I'll report back at the end of my day if I run into anything broken.

ryelle:
Thanks for checking it out!

PR #97, opened 2022-01-18
Menu: Close other modals when when the "open" button is clicked

### LINKED ISSUE #57: Mobile menu and search: inconsistent behaviour and visual refinement

ryelle:
This is a little bit of a workaround, since there isn't a way to do this via Gutenberg, and the library used does not support it either. Whenever an open button is clicked, we quickly check if there are any visible close buttons that don't belong to the just-opened modal, and trigger a click on them.

Fixes #57 

https://user-images.githubusercontent.com/541093/149850733-619e67ce-9d2f-4849-aecb-d147691d76e5.mov

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -1,3 +1,4 @@
+/* global MicroModal */
 /**
  * File wporg-global-header-script.js.
  *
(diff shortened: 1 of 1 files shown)


beafialho:
Thanks @ryelle, according to your screen recording it works as it should!

Issue #110, opened 2022-01-24
Menu Close X not shown in Safari


dd32:
As noted by @aurooba on Slack https://wordpress.slack.com/archives/C02QB8GMM/p1642992375174000 

https://user-images.githubusercontent.com/767313/150723404-e7b8b3be-ff4a-4eda-ad80-de20b8d785fb.MP4

It looks like safari isn't happy with rendering the close button outside of the menu when the menu is `overflow: auto`. Putting together a PR.

ryelle:
Fixed with #111.

PR #111, opened 2022-01-24
Fix the close button not showing up in Safari

dd32:
Fix the close button not showing up in Safari, due to the close button being rendered outside of the elements padding box.

This appears to be styles from upstream in Gutenberg, I'm still creating a PR here for testing, as I'm not 100% sure how to verify it against the Gutenberg trunk blocks.

See #110

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
@@ -167,13 +167,11 @@
 	}
 }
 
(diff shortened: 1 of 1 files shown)


dd32:
This change causes some padding to hide the menu items when scrolled, but the somewhat unrelated #112 fixes that.
Screen Shot 2022-01-24 at 4 19 25 pm

ryelle (approved):
👍🏻  Fixes the issue for me.

PR #112, opened 2022-01-24
Mobile Menu: Reduce the top padding

dd32:
Before:
Screen Shot 2022-01-24 at 3 45 20 pm

After:
Screen Shot 2022-01-24 at 3 44 56 pm

This brings it closer to the figma design, although maybe not pixel perfect.

Screen Shot 2022-01-24 at 3 46 46 pm


ryelle (approved):
Looks good, and fixes the issue here https://github.com/WordPress/wporg-mu-plugins/pull/111#issuecomment-1019758747 when both PRs are applied.

Issue #13, opened 2021-10-14
Replace search fork with Search block


iandunn:
we had to fork the search block in order to customize the url
xref https://github.com/WordPress/wporg-mu-plugins/pull/9#discussion_r718698946

once https://github.com/WordPress/gutenberg/issues/35572 is fixed, the fork should be replaced with the actual block, along with the new attribute/filter

PR #113, opened 2022-01-24
Attempt to remove our custom search html block with the code search block

dd32:
This is an attempt at migrating from using a custom html block to using the core search block.

This renders correctly, but I've not tested with FSE.

The HTML block is still needed, as there's no other way to wrap it in a `<li/>`..

The `global-header__search-form` class was removed, as I couldn't see any way to define it / didn't seem required with the CSS selector change - There's only so many search fields in the header.

Using `esc_attr_*()` inside JSON doesn't feel right, which is the reason for cb113fd instead.

See #13

**Draft as this is not intended on being used straight away, just looking at what is needed**


ryelle:
Closing since I've merged most of this code in with #155 (thanks for laying the groundwork 💯 )

PR #119, opened 2022-01-24
Search: Fix accessibility issues in search dropdown

### LINKED ISSUE #118: Search: Accessibility issues with screen reader
> Missing labels and no way to tell if Search link did anything. Not meeting standards.

https://wordpress.slack.com/archives/C02QB8GMM/p1642993201177300?thread_ts=1642992375.174000&cid=C02QB8GMM

The way we're hacking the nav menu/submenu behavior works visually, but is confusing for a screen reader user - there's a search link, but it goes nowhere. The input itself has a label attached correctly (id/for), but it's hidden using `display: none` so it's ignored.

ryelle:
Fixes #118 — this switches the search navigation menu to use `"overlayMenu":"always"`, so that the toggle button has the correct semantics for a show/hide button.

Currently the search icon in the header is a link that shows a submenu, which works visually, but if you're navigating via screen reader you just hear "Search, link", and clicking it appears to do nothing. By switching to the mobile menu interface always, the search link becomes a button with a popup, and clicking it shows the search menu.

⚠️  This does remove the hover effect (where the search appears on hover), but IMO fixing the accessibility issue is more important than the hover effect. cc @beafialho 

Props to @alexstine for the original bug report.

**Screenshots**

Aside from the hover issue, there should be no visual changes between trunk & this PR.
|   | Before | After |
|---|--------|-------|
| Default |  |  |
| Menu open |  |  |
| Search open |  |  |
| Default |  |  |
| Menu open | |  |
| Search open |  |  |
| Default |  |  |
| Search open |  |  |

**To test**

1. Using a screen reader, navigate through the header
2. You _should not_ hear a "Search, link" item
3. You _should_ hear something like "Open search, menu popup, button"
4. Activate the button, you'll be moved into the search field, which is labelled now
5. Tabbing through keeps you in the search popup - hit tab twice to get back to the close button
6. Hit close or escape to get out
7. Continue on to other content

Also test without using a screen reader, and/or at various sizes. On small screens, there's now a label difference between the navigation menu label ("Open menu") and search menu label ("Open search").

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -38,6 +38,14 @@ function register_block_types() {
 		filemtime( __DIR__ . '/js/wporg-global-header-script.js' ),
 		true
 	);
+	wp_localize_script(
+		'wporg-global-header-script',
+		'wporgGlobalHeaderI18n',
+		array(
+			'openSearchLabel' => __( 'Open Search', 'wporg' ),
+			'closeSearchLabel' => __( 'Close Search', 'wporg' ),
+		)
+	);
 
 	register_block_type(
 		'wporg/global-header',
--- a/mu-plugins/blocks/global-header-footer/header.php
+++ b/mu-plugins/blocks/global-header-footer/header.php
@@ -94,50 +94,46 @@ if ( ! empty( $locale_title ) ) {
 		calls for. It also provides a consistent experience with the primary navigation menu, with respect to
 		keyboard navigation, ARIA states, etc. It also saves having to write custom code for all the interactions.
 	-->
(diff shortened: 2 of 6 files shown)


tellyworth:
I haven't tested with a screen reader, but I did a visual check at various sizes and couldn't find any regressions or obvious problems.

beafialho:
> ⚠️ This does remove the hover effect (where the search appears on hover), but IMO fixing the accessibility issue is more important than the hover effect. cc @beafialho

To make sure I understand @ryelle the only visual change is that instead of the search appearing when you hover it, it appears when you click on it?

The only screenshots with changes I see are these, is that correct?

Captura de ecrã 2022-01-25, às 12 22 48

ryelle:
@beafialho That's correct. It's using the mobile toggle all the time now, so when the search is open, you can click the X to close it.

beafialho:
I see, LGTM 👍

Issue #120, opened 2022-01-25
Search: Tweak UI / interactions


kellychoffman:
A couple tweaks to make on the search field in the header: 

- [x] Mobile: Input should go full width. (See screenshot below)
- [x] When you open the search on small screen, then make the screen bigger, the search stays open and covers nearly all the screen. Instead, it should either snap to the large screen view or the search overlay should close. 

Design reference:

Search input width:
<img width="464" alt="Screen Shot 2022-01-24 at 6 28 35 PM" src="https://user-images.githubusercontent.com/5835847/150899754-f79567da-4e7e-494f-aa8d-16c54ce6e520.png">

dd32:
>  When you open the search on small screen, then make the screen bigger, the search stays open and covers nearly all the screen. Instead, it should either snap to the large screen view or the search overlay should close.

This is similar to #56 but now affects the search too.

dd32:
Looks like this is resolved via #121 & #122

Issue #115, opened 2022-01-24
Header has different height than the mockup


beafialho:
According to the design, the Global Nav bar should have a height of `90px`. In the site, it has `103px`.

<img width="1030" alt="Captura de ecrã 2022-01-24, às 15 14 06" src="https://user-images.githubusercontent.com/65220155/150810119-26a8c35c-7980-4fc2-a8a2-e81232a1d63e.png">

<img width="1531" alt="Captura de ecrã 2022-01-24, às 15 12 54" src="https://user-images.githubusercontent.com/65220155/150810240-3040ab7d-bcc4-4bfe-ba05-904362d0b090.png">

dd32:
Just noting that according to https://github.com/WordPress/wporg-news-2021/issues/181 the designed size was 104px, Did I misread that comment, or was this intended to be 104px and this is a request to reduce it to 90px?

beafialho:
@dd32 I was referring to the mobile header height in that comment. In [the design](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=900%3A27) on desktop it's `90px`.

dd32:
This was fixed via #121.

Issue #16, opened 2021-10-22
Increase header menu target size


dd32:
Currently the target areas for the menu items in the header are much smaller than the size of the menu, this leads to a lot of dead space around menu items that is unclickable.

Additionally, there's no feedback to the user (other than the mouse pointer) on if they are/are not over a menu item.

Can we increase the target area for menu's to be the full "area" of the item?

For example, I've added a background colour to the existing hover/clickable area in this video:

https://user-images.githubusercontent.com/767313/138382182-295d5d1a-cdd9-4ec9-b779-ab624e2e9a71.mov

At the one-second mark you'll also see a glitch where it looks like the menu is shown before the parent target is reached, looks like there might be a 1px window where the menu is hidden in the background.

dd32:
> At the one-second mark you'll also see a glitch where it looks like the menu is shown before the parent target is reached, looks like there might be a 1px window where the menu is hidden in the background.

Turns out that was because the mouse crossed the lower section of the support tab on it's way to News, as something like HoverIntent isn't used here to verify that the user was actually heading to that item. Similarly, the menu vanishes instantly without any form of grace period (even if the link for the menu is still in a hover state, but the mouse has left)

kellychoffman:
Two more things that can be done in the same PR: 

- [ ] Remove black background on hover. (not seeing that in any of the Figma files, @beafialho please confirm.)
- [ ] Remove excess top padding on mobile view

dd32:
>  Remove black background on hover. (not seeing that in any of the Figma files

This was requested via #63

beafialho:
> Currently the target areas for the menu items in the header are much smaller than the size of the menu, this leads to a lot of dead space around menu items that is unclickable.

I agree they're too small and think the entire area around the menu items could be clickable.

<img width="774" alt="Captura de ecrã 2022-01-26, às 16 19 30" src="https://user-images.githubusercontent.com/65220155/151202733-5b9320c9-dae6-47f3-b7cb-f817003103f2.png">

> Two more things that can be done in the same PR: Remove black background on hover. (not seeing that in any of the Figma files, @beafialho please confirm.) Remove excess top padding on mobile view

There should be a [dark background](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=11289%3A94606) on hover. @kellychoffman is this the top padding you're referring to?

[...]

PR #121, opened 2022-01-25
Header padding & sizing adjustments

dd32:
This PR is focused on several issues:

 - Desktop Header height was implemented at 105px rather than 90px, #115
 - Desktop Menu targets should have been larger, #16
 - Mobile Menu padding doesn't need to be as large, #16 
 - Desktop header is too large consistently on scroll for small viewports, common community complaint, as suggested by a few, reducing it on scroll make sense IMO. This might not be as required due to the 15px removed in #115, but another 30px is not insignificant on some screen resolutions.

60px for mobile was chosen as at 2/3's of the 90px expected height, and any smaller felt cramped. It's also just over twice the logo sizes.

A CSS variable is introduced for those different header sizes, as it's needed in other themes for determining placement.

The 'slimline' variant introduced on scroll here was an experiment, and easily removable, but did not make sense as a separate PR due to the other size changes required here (and for the functionality). 063b3487a0ecd5a29c31e73606bff6e6453c5203 & 9a8708da7723d4f57edb4f3bfcd17975c30947c6 can be removed if required and pushed to a different PR dependant upon this one.
It might make sense to only use the slim header on smaller window resolutions, or not implement it at all, and instead simply disable the fixed positioning in small windows.

Full header:
Screen Shot 2022-01-25 at 5 17 58 pm

Slim header:
Screen Shot 2022-01-25 at 5 23 04 pm

Video:
https://user-images.githubusercontent.com/767313/150933331-36fb7f15-b0b5-432e-b2ae-065b7bfae18e.mov

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -200,35 +200,4 @@
 			}
 		}
 	} );
-
-	/**
-	 * Scroll throttle flag
-	 *
-	 * @param {bool} throttled or not.
-	 */
-	let scrollThrottled = false;
-
-	/* eslint-disable @wordpress/no-global-event-listener */
-	window.addEventListener( 'scroll', function () {
-		if ( scrollThrottled ) {
-			return;
-		}
-
-		let isSlimline = document.documentElement.classList.contains('slimline'),
-			shouldBeSlimline = Math.max( document.body.scrollTop, document.documentElement.scrollTop ) > 250;
-
-		if ( ! isSlimline && shouldBeSlimline ) {
-			document.documentElement.classList.add('slimline');
-
-			scrollThrottled = true;
-		} else if ( isSlimline && ! shouldBeSlimline ) {
-			document.documentElement.classList.remove('slimline');
-
-			scrollThrottled = true;
-		}
-
-		if ( scrollThrottled ) {
-			setTimeout(() => scrollThrottled = false, 400 );
-		}
-	} );
 } )();
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -200,4 +200,35 @@
 			}
 		}
 	} );
+
+	/**
+	 * Scroll throttle flag
+	 *
+	 * @param {bool} throttled or not.
+	 */
+	let scrollThrottled = false;
+
+	/* eslint-disable @wordpress/no-global-event-listener */
+	window.addEventListener( 'scroll', function () {
+		if ( scrollThrottled ) {
+			return;
+		}
+
+		let isSlimline = document.documentElement.classList.contains('slimline'),
+			shouldBeSlimline = Math.max( document.body.scrollTop, document.documentElement.scrollTop ) > 250;
+
+		if ( ! isSlimline && shouldBeSlimline ) {
+			document.documentElement.classList.add('slimline');
+
+			scrollThrottled = true;
+		} else if ( isSlimline && ! shouldBeSlimline ) {
+			document.documentElement.classList.remove('slimline');
+
+			scrollThrottled = true;
+		}
+
+		if ( scrollThrottled ) {
+			setTimeout(() => scrollThrottled = false, 400 );
+		}
+	} );
 } )();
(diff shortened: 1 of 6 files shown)


dd32:
> Desktop header is too large consistently on scroll for small viewports, common community complaint, as suggested by a few, reducing it on scroll make sense IMO.

One such comment suggesting that is here: https://github.com/WordPress/wporg-news-2021/issues/181#issuecomment-1014781537 but that was also with the notion that the header was staying at 104px in size.

beafialho:
> Desktop header is too large consistently on scroll for small viewports, common community complaint, as suggested by a few, reducing it on scroll make sense IMO.

Thanks @dd32. Reducing the header on scroll was my initial idea for the header redesign if we were going to make it sticky, so I agree with the proposed solution. I had suggested the same behavior for mobile [here](https://github.com/WordPress/wporg-news-2021/issues/181), though I think we can adopt this **only for desktop** at this point. `60px` for the "slim header" looks reasonable to me.

ryelle:
>  It might make sense to only use the slim header on smaller window resolutions

+1 to using the slim header on shorter screens, not scroll. Changing it on scroll is distracting, borderline an accessibility issue (yes it's common behavior, but that doesn't mean it's good). Also, and I can't replicate it, but I did notice an issue where the slimline class was toggled causing the header to bounce. That _would_ be an a11y issue, so it would be safer to not.

I've rebased this after #119, and fixed a few small bugs I noticed.

beafialho:
I'm not able to test the header locally, could you post a screenshot with the changes if any have been made?

ryelle:
Screenshots with current changes:

Mobile is 60px tall
480-header

Mobile menu had some alignment issues that I fixed, so it looks like this:
480-header-menu

Desktop-sized header is 90px tall 
1200-header

I did not change the scroll behavior (but still think we should), so it flips to 60px tall when you're somewhat down the page:
1200-header-slim

zackkrida:
One thing I've observed, prior to any changes in this PR, is that on mobile there's some padding on the bottom of the menu. Depending on the height of the mobile device, this can obscure the fact that there's additional menu items:

CleanShot 2022-01-26 at 14 17 24@2x

CleanShot 2022-01-26 at 14 16 42@2x

Do y'all think this is better-reported elsewhere?

dd32:
> +1 to using the slim header on shorter screens, not scroll. Changing it on scroll is distracting, borderline an accessibility issue 

Agreed, the saving here is not exactly significant as a percentage of screen size either, if we can decide on a vertical resolution to apply the slimmer header to, I think I agree that this would be best.

> Also, and I can't replicate it, but I did notice an issue where the slimline class was toggled causing the header to bounce.

Hah. Glad you saw that :) There's a specific resolution and certain scroll position, and specific animation time, where that would happen.. I thought I had got rid of it, but evidently not!

beafialho:
> Mobile is 60px tall

> Mobile menu had some alignment issues that I fixed, so it looks like this:

> Desktop-sized header is 90px tall

Thank you, all looks good to me 👍 

> +1 to using the slim header on shorter screens, not scroll. Changing it on scroll is distracting, borderline an accessibility issue

Just to make sure I understand correctly, your suggestion is to keep it fixed to the top with `90px` in larger screens and also fixed but with `60px` in shorter screens? If that sounds like a better overall plan, I agree we stick to that.

ryelle:
> your suggestion is to keep it fixed to the top with 90px in larger screens and also fixed but with 60px in shorter screens? 

yeah, exactly.

> if we can decide on a vertical resolution to apply the slimmer header to

How about 800px? If it's shorter than 800px, show the 60px tall header, and larger should show 90px header. At 800px, the 90px header is about 11% of the screen height. (I think 900px could also work, but I don't think most people use browsers that tall - even fullscreen on my 13" macbook, the viewport is only 815px tall)

beafialho:
> How about 800px? If it's shorter than 800px, show the 60px tall header, and larger should show 90px header. At 800px, the 90px header is about 11% of the screen height.

800px seems ok 👍

ryelle:
Okay, I've removed the scroll toggle, and fixed the extra padding on the mobile menu mentioned [here](https://github.com/WordPress/wporg-mu-plugins/pull/121#issuecomment-1022525765), _and_ fixed the [search width](https://github.com/WordPress/wporg-mu-plugins/issues/120#issue-1113352934) issue from #120 so I think this will close #120 now, too.

Mobile:

| default | menu open | search open |
|---|---|---|
| mobile-closed | mobile-menu | mobile-search |

700px tall, short menu:
short

Short menu on a rosetta site:
rosetta

900px tall, regular menu:
full

Issue #56, opened 2022-01-05
Mobile menu not working correctly when browser window is resized


beafialho:
This is how the header currently behaves when I resize the browser window:

https://user-images.githubusercontent.com/65220155/148217309-8c8ffc76-f0fa-4c13-863b-6c4f5f80eea7.mp4

It should adopt its original responsive behaviour, with menu links showing or hiding as the window is resized, as it does when the menu button is not active:

https://user-images.githubusercontent.com/65220155/148218052-20959605-9073-4a69-837e-ed091b2bfad7.mp4

iandunn:
That's Gutenberg's standard behavior. Do you think anyone will do that in practice?

beafialho:
> Do you think anyone will do that in practice?

Yes, I think it can happen.

PR #122, opened 2022-01-25
Close all mobile menus on resize.

dd32:
See  #120 & #56

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -183,7 +183,7 @@
 	window.addEventListener( 'resize', () => {
 		// Hide any open mobile menus if we're no longer in a mobile view.
 		if ( ! document.querySelector('.global-header__navigation .wp-block-navigation__responsive-container-open').offsetWidth ) {
-			let closeMenuButton = document.querySelector('.wp-block-navigation__responsive-container.is-menu-open [data-micromodal-close]');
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
```diff
@@ -188,4 +178,15 @@
 			} );
 		} );
 	} );
+
+	/* eslint-disable @wordpress/no-global-event-listener */
+	window.addEventListener( 'resize', () => {
+		// Hide any open mobile menus if we're no longer in a mobile view.
+		if ( ! document.querySelector('.global-header__navigation .wp-block-navigation__responsive-container-open').offsetWidth ) {
+			let closeMenuButton = document.querySelector('.wp-block-navigation__responsive-container.is-menu-open [data-micromodal-close]');
```
ryelle:
```suggestion
			const closeMenuButton = document.querySelector( '.wp-block-navigation__responsive-container.is-menu-open button[data-micromodal-close]' );
```

For some reason this library adds that `data-micromodal-close` attr to the modal _and_ the button, so this just ensures you're grabbing the button. Both work, but specifying the button would probably be more stable.

ryelle (changes requested):
This works, but I have one comment about the selector.

mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
dd32: @ryelle Agreed, I committed your suggestion as-is, I had originally targeted the button, but swapped to the attribute and failed to notice that it added it on both elements when it still worked.

ryelle (approved):

Issue #128, opened 2022-01-26
Admin bar not consistent across sites


iandunn:
@justintadlock pointed out that [some sites in the network show the admin bar, and some don't](https://wptavern.com/wordpress-org-gets-new-global-header-and-footer-design).

`wporg-no-adminbar-if-no-blogs.php` intentionally disables it on sites that the user doesn't have a role on. The benefits I see are that it avoid confusions, preserves screen real estate, and avoids frustrating people who dislike sticky headers. The drawback is that it's not consistent.

Turning it on doesn't really add any value, other than fixing the inconsistency problem. It may also introduce links etc that don't make sense in a "i don't have a role on this blog" context. We could remove those, though. There is a `logged-out-admin-bar.php` plugin that's active on some sites.

For non-WP sites like Trac, we might be able to use the new API endpoint to include it, although we'll want to think about any security implications. It definitely should bypass Trac's cache, so we'd need to load it via JS, which might introduce some FOUC if it's not done early enough.

@beafialho, do you have any thoughts from the design perspective?

beafialho:
> @beafialho, do you have any thoughts from the design perspective?

I agree it's not consistent but I think addressing this would be a part of the future effort of rethinking the header architecture (menu items included), and possibly integrating the admin bar to the global header.

@kellychoffman may have more useful input.

kellychoffman:
I always wondered why it did that! Ian - Your reasoning makes sense for why it was implemented still makes sense and I say we keep it as is.

iandunn:
👍🏻  I'll close for now, and we can reopen in the future if we decide to combine the two headers.

Issue #34, opened 2021-12-14
Header: Refactor information architecture


iandunn:
The menu now has so many links that we need to use a "overflow" script (see #21). Folks have discussed an IA refactor, including the following rough ideas

### Header

- [ ] add blocks. they're in the plugins section already, but that's not obvious
- [x] add openverse
- [ ] maybe rename "openverse" to "images". maybe have both, one as `add-ons > images` and the other as `tools > openverse`
- [x] group plugins/themes/blocks/patterns/images under 1 top menu
- [ ] maybe rename "download and extend" to "add-ons" / "resources" / "extensions" ?
- [x] Moving Showcase, Mobile, and Hosting pages out of the header, or move into submenus. Showcase could go under About; Mobile and Hosting could go under Resources. Maybe convert the Get WP button to a regular menu, which has "hosting", "download", and "mobile" as subitems
- [ ] Direct access to internal pages of Learn and Support.
- [x] add WordPress.TV
- [ ] add Browse Happy? bbPress and/or BuddyPress?
- [x] add Developer tools like WP-CLI?
- [x] something for plugin/theme devs, since that’s such a fundamental part of the ecosystem. maybe a submenu item for https://developer.wordpress.org/, 
- [ ] maybe make Developers a top-level section?
- [x] Maybe add an “Community” menu, w/ WordCamps, meetups, Learn discussion groups etc?

### Footer

- [ ] add labels to each column so it's clear how they're organized
- [ ] some items are in the header and not footer, and vice-versa. evaluate those and decide if each makes sense

iandunn:
Closing since #251 is a duplicate with newer discussion

iandunn:
reopening since this is more of a long term IA refactor, while #251 ended up being a short term iteration. many of the items here were done by `251` though, so I've checked those off.

richtabor:
Marking as closed by https://github.com/WordPress/wporg-mu-plugins/pull/739

PR #130, opened 2022-01-28
Added Meetings and Team Updates submenus.

MuhammadFaizanHaidar:
Added Meetings and Team Updates submenus to Get Involved parent menu.
Track ticket: https://meta.trac.wordpress.org/ticket/5065

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -503,6 +503,16 @@ function get_global_menu_items() {
 
 			'submenu' => array(
 				array(
+					'title' => esc_html_x( 'Meetings', 'Menu item title', 'wporg' ),
+					'url'   => 'https://make.wordpress.org/meetings/',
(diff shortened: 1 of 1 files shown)


iandunn:
@beafialho do you have thoughts on this idea and the Trac ticket linked above?

Related: #34

beafialho:
I do have thoughts, but like in #128 I think this would also be a part of rethinking the header architecture, which is its own future effort. It'll take more than adding or removing links to reorganize it IMO. I also wonder if it's worth applying changes to menus too often.

coffee2code:
Thanks for the PR! It has been some time since this request was made and the global nav menu has changed in the meantime. The "Get Involved" menu option has been replaced with a "Community" option that has six submenu links (one of which is "Make WordPress"). Adding "Meetings" and "Team Updates" verges on presenting too many options now, but most of all, they are too niche to be included in the global nav. Their inclusion in the in-page nav menu on https://make.wordpress.org is more appropriate (and may have been added after this ticket?).

PR #131, opened 2022-01-28
Change header and footer font sizes to 14px

### LINKED ISSUE #127: Menu font is small
There's been some [feedback about the small font size in the menu](https://wptavern.com/wordpress-org-gets-new-global-header-and-footer-design#comment-413392). It feels "readable but uncomfortable" to me too.

Now that we have the priority navigation (the `...`), do we still need the small font size?

mikachan:
This changes the header and footer font sizes from `13px` to `14px`, as per the discussion in #127.

I've used the preset variable, `var(--wp--preset--font-size--small)`, as this is currently set to `14px`, but I don't know if these sizes should be independent of the variables. If so we could just use `14px`.

Closes #127.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/common.pcss
@@ -66,7 +66,7 @@ html {
 		font-size: 21px;
 
 		@media (--tablet) {
-			font-size: 13px;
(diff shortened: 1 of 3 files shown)


iandunn (approved):
LGTM, and using that var sounds good.

PR #132, opened 2022-01-31
Prevent header FOUC by forcing the navigation not to wrap, but rather be clipped

### LINKED ISSUE #41: Header FOUC before priority nav script loads
1. Size browser to `1025px`
2. Load a page w/ the header
2. Browser will render the header w/out the priority navigation, so menu items will wrap
3. Browser will load the priority nav script and re-render the header, this time w/ the extra items hidden behind `...`

We might need to load the script sooner, or something like that. Or maybe `overflow-x: none`

dd32:
Fixes #41.

This forces the navigational menu to overflow horizontally, and be clipped. 

The simplest way to test this is to use a browser with JS disabled, you should be able to access the first few menu items.

Then with JS enabled, you should be able to access the overflow menu, and everything should be visually correct.

https://user-images.githubusercontent.com/767313/151741568-c358a21f-6c30-4fae-895c-95e0f0c8557c.mov

A `.has-overflow-menu` class was added to allow disabling the Overflow-X, as otherwise you'd have this:
(Note the background of the menu is clipped)
Screen Shot 2022-01-31 at 3 05 52 pm

The removal of the `left: 0` is to place the submenu correctly, I'm not sure why It's needed.. but I can't find any regressions and you otherwise end up with this, with 5ftf overlaying on Get Involved:
Screen Shot 2022-01-31 at 3 08 19 pm

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
+++ b/mu-plugins/blocks/global-header-footer/js/wporg-global-header-script.js
@@ -29,6 +29,8 @@
 			this.removeSubMenu();
 			this.hasHiddenItems = false;
 			this.listItems = this.getListItems();
+
+			this.wrapper.classList.remove( 'has-overflow-menu' );
 		};
 
 		/**
@@ -93,7 +95,9 @@
 
 				let itemsContainer = this.wrapper.querySelector( '.wp-block-navigation__container' );
 
-				//create the ... menu list item
+				this.wrapper.classList.add( 'has-overflow-menu' );
+
+				// Create the ... menu list item.
 				let newItem = document.createElement( 'li' );
 				newItem.classList.add(
 					'wp-block-navigation-item',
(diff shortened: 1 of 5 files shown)


ryelle:
The `flex-wrap: no-wrap` was causing this for me:

Screen Shot 2022-02-01 at 3 14 54 PM

Instead, I switched the `align-items` on the whole header, so that the items would wrap under, rather than above the menu line. This made some alignment issues more obvious, so I've also fixed those.

dd32:
> Instead, I switched the `align-items` on the whole header, so that the items would wrap under, rather than above the menu line. This made some alignment issues more obvious, so I've also fixed those.

Thanks, looks good to me!

I originally abandoned that approach as it was causing the items to wrap onto the next line (it still does) which would then be displayed below the header, and then quickly vanish. It looks like a later change caused the overflowed items to not be visible, resolving that approach.

ryelle:
>  It looks like a later change caused the overflowed items to not be visible, resolving that approach.

Yeah, I also had trouble with that - the `overflow: clip` + set height on the menu fixed that, though 👍🏻

PR #140, opened 2022-02-02
Menu: Correctly select the correct support item.

dd32:
The previous regular expression only contained a limited selection of Support URIs, missing /support/plugin/ and /support/theme/.

This flips it around to matching primarily for Documentation pages, defaulting to forums otherwise.

Also switches to using PHP8 str_starts_with() for consistency.

Reported here: https://wordpress.slack.com/archives/C02QB8GMM/p1643707505001439
> Hi, I'm not sure if this is the best place to ask, but I find it weird that at a plugin's main support page (https://wordpress.org/support/plugin/[plugin-name]) The Support > Documentation page is highlighted instead of Support > > Forums. But if you go to any topic, then the "Forums" item is higlighted correctly.
> See:
> https://wordpress.org/support/plugin/classic-editor/
> Screenshot: https://imgur.com/ngOtp9D

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -983,12 +983,15 @@ function set_current_item_class( $menu_items ) {
  * @return string
  */
 function get_menu_url_for_current_page( $menu_items ) {
+	$host    = strtolower( $_SERVER['HTTP_HOST'] );
+	$uri     = strtolower( $_SERVER['REQUEST_URI'] );
+	$compare = "https://{$host}{$uri}";
 
-	if ( 'translate.wordpress.org' === $_SERVER['HTTP_HOST'] ) {
+	if ( 'translate.wordpress.org' === $host ) {
 		return 'https://make.wordpress.org/';
 	}
 
-	if ( 'developer.wordpress.prg' === $_SERVER['HTTP_HOST'] ) {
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -983,12 +983,15 @@ function set_current_item_class( $menu_items ) {
...
+	$uri     = strtolower( $_SERVER['REQUEST_URI'] );
+	$compare = "https://{$host}{$uri}";
 
-	if ( 'translate.wordpress.org' === $_SERVER['HTTP_HOST'] ) {
+	if ( 'translate.wordpress.org' === $host ) {
 		return 'https://make.wordpress.org/';
 	}
 
-	if ( 'developer.wordpress.prg' === $_SERVER['HTTP_HOST'] ) {
+	if ( 'developer.wordpress.prg' === $host ) {
```
ryelle: I think this is a typo?

```suggestion
	if ( 'developer.wordpress.org' === $host ) {
```

ryelle (approved):
Looks good (aside from the minor typo). I tried a bunch of support subpages and they seemed to highlight the correct sub-item 👍🏻

dd32:
> aside from the minor typo

oops! That was in there from the first PR!

>  I tried a bunch of support subpages and they seemed to highlight the correct sub-item 👍🏻

Thank you!

PR #143, opened 2022-02-02
Fix --short-screen @media queries for non-FF browsers

dd32:
The Small-screen header minimizing is only working in Firefox, Not in Chrome/Safari.

This appears to be caused by these browsers not accepting `@media ( (a) and (b) )`, but they're happy with `@media (a) and (b)`.
Example:
 - working: https://jsfiddle.net/7386khqe/
 - broken: https://jsfiddle.net/vdh6pq4n/ (but working in Firefox)

I think this is because the CSS @media queries are not designed for nested sets, but rather only `(property:value)`.

It seems like this syntax is the expected format, as post-css-custom-media supports it:
https://github.com/postcss/postcss-custom-media/blob/master/test/export-media.css#L8

See #121


ryelle (approved):
Thanks for catching this!

PR #149, opened 2022-02-07
Use the environment type, rather than the WPORG_SANDBOX constant

### LINKED ISSUE #138: Global styles are not loaded on local dev sites
Again, I'm trying to use this on the Pattern Directory repo (#136), and I've run into another issue… this is a classic theme, and I don't have the News site locally to generate the custom properties, so the menu loads but is mostly unstyled:

Screen Shot 2022-02-01 at 5 19 04 PM

It seems like the `fetch_global_styles` function should handle this but it isn't, it won't connect to load the styles (I _think_ because I'm sandboxing wordpress.org while working, but it should still work).

My hacky workaround so I can move on is to enqueue the styles endpoint directly in a `sandbox.php` mu-plugin:

```php
add_action(
	'wp_enqueue_scripts',
	function() {
		wp_enqueue_style(
			'global-header-footer',
			'https://wordpress.org/wp-json/global-header-footer/v1/styles',
			array(),
			'1'
		);
	}
);
```

dd32:
When the header is used outside of the WordPress.org network, it uses the binary `WPORG_SANDBOX` flag as to whether it should use localhost or wordpress.org to fetch the global styles from.

Unfortunately, some local environments, such as the Pattern Directory, declare themselves to be sandboxes, as they had no other way to differentiate between Production and not production.

WordPress.org sandboxes now declare themselves to be a staging environment, which gives some more granularity here:
 - production: Should request against wordpress.org
 - sandbox: Should request against localhost
 - local/development: Should request against wordpress.org (Unless they're developing the header, in which case, this branch probably won't be hit).

Fixes #138

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -915,7 +915,7 @@ function fetch_global_styles() {
 		$request_args = array();
 
 		// Route request to sandbox when testing.
-		if ( defined( 'WPORG_SANDBOXED' ) && WPORG_SANDBOXED ) {
+		if ( 'staging' === wp_get_environment_type() ) {
 			$hostname                        = '127.0.0.1';
 			$request_args['headers']['host'] = 'wordpress.org';
 


ryelle (approved):
Looks good, and wp-env automatically sets this to local on our dev envs 👍🏻

PR #152, opened 2022-02-08
Header: Switch from `sticky` to `fixed` positioning

### LINKED ISSUE #147: Navigation: Tabbing through menu causes upward scroll on page
Reported in slack by @joedolson

> I'm not sure where to report this issue, as I can't really tell where it's coming from, but navigating the new header menu while in trac seems to cause some strange positional shifts in the page content. Gif coming to show it. I've only been able to reproduce this on trac, so far.
>
> keyboard-shift
> The gif shows me tabbing through the main menu, and the trac content shifting with each keypress.

I was able to reproduce the same issue on make.wordpress.org, in firefox & safari. I imagine any long page will show the issue.

To reproduce:

1. Scroll partway down a page
2. Tab through the header & nav menu
3. Watch as the page content scrolls up slightly with each tab

ryelle:
Fixes #147 by switching from `sticky` positioning to `fixed`. I'm pretty sure the reasoning here is that `fixed` takes the element out of the document context, so the browser doesn't think it needs to scroll up to get the focused content on-screen.

This uses a `margin-top` to cover the space of the menu, to offset the whole page under the menu; and `scroll-padding-top` to make sure in-page anchor links work correctly.

To test:

- Sync this to a sandbox
- Try a few different pages, with & without admin bars
- Try at different browser sizes — at `890px` the header stops being sticky, so try at least one size above & one below
- The header should stay stuck to the top of the page
- When scrolled to the top of the page, all content should be visible, no content hidden by the header
- Try using some relative links (for example, [a make/core handbook page](https://make.wordpress.org/core/handbook/#contribute-with-code)), the header should not overlap the link target.

Note: The news site needs an extra fix, will PR that next.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -7,6 +7,7 @@ use Rosetta_Sites, WP_Post, WP_REST_Server, WP_Theme_JSON_Resolver;
 defined( 'WPINC' ) || die();
 
 add_action( 'init', __NAMESPACE__ . '\register_block_types' );
+add_action( 'init', __NAMESPACE__ . '\remove_admin_bar_callback' );
 add_action( 'enqueue_block_assets', __NAMESPACE__ . '\register_block_types_js' );
 add_action( 'rest_api_init', __NAMESPACE__ . '\register_routes' );
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_compat_wp4_styles', 5 ); // Before any theme CSS.
@@ -67,6 +68,15 @@ function register_block_types() {
 			'editor_style'    => 'wporg-global-header-footer',
 		)
 	);
+}
+
+/**
+ * Remove the default margin-top added when the admin bar is used.
+ *
+ * The core handling uses `!important`, which overrides the sticky header offset in `common.pcss`.
+ */
+function remove_admin_bar_callback() {
+	add_theme_support( 'admin-bar', array( 'callback' => '__return_false' ) );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


coreymckrill (commented):
Mostly works as described when testing on my sandbox. Except:

> Try using some relative links (for example, [a make/core handbook page](https://make.wordpress.org/core/handbook/#contribute-with-code)), the header should not overlap the link target.

While the header does not overlap the link target, it does overlap the Welcome Box at the top, covering up the "Hide welcome box" button in the upper right corner. (This is on Firefox.)

handbook-overlap

ryelle:
@coreymckrill Good catch, looks like that comes from a reset in `p2-breathe`. I added a rule for slightly more specificity (https://github.com/WordPress/wporg-mu-plugins/pull/152/commits/ad2e7d872449304f4061a190a458182cf03a7614 ), though I don't really _love_ this solution… I would rather a way to make sure this CSS is loaded last. While writing this comment I decided to try it, and now have https://github.com/WordPress/wporg-mu-plugins/pull/152/commits/1fd46bdd462af0e313f3d3cb8ba139159348d16c, which filters the styles before output to shift the block styles to the end of the list.

I don't think there are any drawbacks to the style filter approach, I'm pretty sure `html` is the only element selector used, so there shouldn't be any unexpected cascade issues. I'll need to update https://github.com/WordPress/wporg-news-2021/pull/297 either way. Do you have a preference for either solution?

coreymckrill:
🤔 I guess either solution is fine. I don't anticipate any real side effects either. The one thing I would say about the style filter approach is it's rather unusual/clever, in that normally you would try to set the order of stylesheets with different priorities on `add_action( 'wp_enqueue_styles' )`. But maybe that doesn't work here? If someone later is trying to figure out why these particular styles are loading last, it may not be immediately obvious.

On the other hand, I get that having to use high specificity in stylesheets is usually not great. So again, it seems like either solution is fine 🤷

ryelle:
> in that normally you would try to set the order of stylesheets with different priorities on `add_action( 'wp_enqueue_styles' )`

Good call, definitely trying to be too clever. I thought the styles had to be in `init` because they're used in the block registration, but that's not the case. Last change 🤞🏻 , it's now using `wp_enqueue_scripts` with a priority of 200.

coreymckrill (approved):

PR #155, opened 2022-02-11
Search: Use search block with icon button

### LINKED ISSUE #13: Replace search fork with Search block

### LINKED ISSUE #99: CORS issue with Search field
When the Header is used outside of WordPress, by serving a cached copy of the REST API endpoints, browsers will encounter a CORS issue with the search field button svg.

```
Access to image at 'https://wordpress.org/wp-content/mu-plugins/pub-sync/blocks/global-header-footer/images/search.svg' from origin 'https://codex.wordpress.org' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

This is because `button.wp-block-search__button` has a `mask-image: url( ...svg )` style applied. When images are used as a mask-image, appropriate CORS headers must be present on the image to allow the browser to load the bitmap of the URL into memory, rather than simply rendering it as a background-image or inline image.

Options:
 - Ensure the image files are served with CORS headers
 - Use `background-image` with `-webkit-filter: invert(100%)`
 - Embed the SVG in the CSS; `-webkit-mask-image: url("data:image/svg+xml;base64,...`
 - (Not an option) Always serve the image from the same domain as the page is rendered on.

Ideally, we'd also serve all of the static images from s.w.org, which also doesn't set CORS headers for SVGs at present. So a solution here may be to ensure s.w.org sets appropriate CORS headers and then move all statics to the CDN.

This is low-priority as it only affects Trac, Codex, and Planet at present. The lack of a visual search button there isn't critical as enter processes the search regardless.

### LINKED ISSUE #148: No keyboard focus state on search submit button
Pretty self-explanatory, I think. The submit button for the search form does not have any keyboard focus state.

image

ryelle:
Fixes #148, fixes #99, fixes #13. This takes some of the code from #113 to use the search block itself, rather than hard-coding it with the custom HTML. Now the search form is generated by the search block, with the "use button with icon" and "button inside" settings.

The button uses the core icon, which looks the same to me, and fixes the CORS issue. Using the "button inside" format of the search lets us remove some of the button positioning code so that the button is visible again, and can get a proper focus style.

Note that the icon button doesn't currently have an accessible label - there's a PR in progress for that here: https://github.com/WordPress/gutenberg/pull/38136

| Before | After: News | After: Make p2 |
|--------|-------------|----------------|
| 480-trunk | 480-branch-news | 480-branch-make |
| 1200-trunk-news | 1200-branch-news | 1200-branch-make |
| 1200-trunk-news-focus | 1200-branch-news-focus | 1200-branch-make-focus |

_The spacing difference on the small make screenshot is because of a gutenberg version difference._

### DIFF
--- a/mu-plugins/blocks/global-header-footer/header.php
+++ b/mu-plugins/blocks/global-header-footer/header.php
@@ -21,7 +21,6 @@ if ( ! empty( $locale_title ) ) {
 $search_args = array(
 	'label' => _x( 'Search', 'button label', 'wporg' ),
 	'placeholder' => _x( 'Search WP.org...', 'input field placeholder', 'wporg' ),
-	'buttonText' => _x( 'Submit search', 'button label', 'wporg' ),
 	'buttonPosition' => 'button-inside',
 	'buttonUseIcon' => true,
 	'formAction' => 'https://wordpress.org/search/do-search.php',
--- a/mu-plugins/blocks/global-header-footer/header.php
+++ b/mu-plugins/blocks/global-header-footer/header.php
@@ -18,6 +18,15 @@ if ( ! empty( $locale_title ) ) {
 	$container_class .= ' global-header__has-locale-title';
 }
 
+$search_args = array(
+	'label' => _x( 'Search', 'button label', 'wporg' ),
+	'placeholder' => _x( 'Search WP.org...', 'input field placeholder', 'wporg' ),
+	'buttonText' => _x( 'Submit search', 'button label', 'wporg' ),
+	'buttonPosition' => 'button-inside',
+	'buttonUseIcon' => true,
+	'formAction' => 'https://wordpress.org/search/do-search.php',
+);
+
 ?>
 
 <!-- wp:group {"tagName":"header","align":"full","className":"<?php echo esc_attr( $container_class ); ?>"} -->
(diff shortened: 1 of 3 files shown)


coreymckrill (approved):
👍 Kudos, this looks like a good, seamless solution. One non-blocking issue I noticed is that because the Search dropdown is treated like a modal, in that you can't scroll down the page while it's open, the scroll bar goes away, which causes a layout jump:

https://user-images.githubusercontent.com/916023/153688768-bc9fbfd2-68f5-4a04-8c5a-a66c8a41dda7.mp4

ryelle:
hm, the layout jump isn't great, but it's intentional that you can't scroll, since you can't actually interact with the rest of the page until the modal is closed (because it's functionally a dialog). Is that new to this branch though? the dialog behavior is the same on trunk.

coreymckrill:
> Is that new to this branch though? the dialog behavior is the same on trunk.

Hmm, not new, I guess. I'm seeing it on the production test site as well.

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -89,51 +97,8 @@
 	<!-- /wp:navigation -->
 
 	<?php if ( $show_search ) : ?>
-	<!--
-		The search block is inside a navigation submenu, because that provides the exact functionality the design
-		calls for. It also provides a consistent experience with the primary navigation menu, with respect to
-		keyboard navigation, ARIA states, etc. It also saves having to write custom code for all the interactions.
-	-->
```
iandunn, line 96: Was removing the above comment intentional? It's not related to how the search block was forked, and still seems like helpful documentation to me.

ryelle: I thought it was outdated since the submenu wrapper was removed in #119, but I guess it could be referring to the nav wrapper itself? Feel free to put it back, if so.

iandunn: Yeah, I think it's there to address the confusion over why a search box would be inside a menu.

iandunn: restored in 1951c1643d1bcfa999fdd927c6fb6e299c12287e

PR #161, opened 2022-02-15
New plugin: Global Fonts for loading Inter and EB Garamond.

ryelle:
See https://github.com/WordPress/wporg-news-2021/issues/234 — This PR adds a new mu-plugin that can be used across the network to load self-hosted versions of EB Garamond and Inter. Both fonts are used in the global header & footer (EB Garamond only on rosetta sites), and both fonts are used on News. There's a corresponding PR in News to remove the google font calls there.

To test:

- Make sure you're loading the new mu-plugin, either by requiring `loader.php` or specifically adding `mu-plugins/global-fonts/index.php` to your test site
- Everything should look the same, but there should be no request to Google Fonts (except maybe for Open Sans)

⚠️  ~Right now I'm getting a 403 on the `.ttf` for EB Garamond, assuming that filetype is not on an allowed list. I'm not sure if that's a sandbox issue or everything issue.~ Looks like this was a sandbox issue, but I converted the files to `woff2` to check, and the new files are only 70% smaller, so I'll stick with those.


coreymckrill (approved):
👍 Looking good locally. I'm still occasionally seeing calls to gstatic.font, but I think that must be coming from the theme rather than the global header.

ryelle:
>  I'm still occasionally seeing calls to gstatic.font, but I think that must be coming from the theme rather than the global header.

Yeah, the classic themes use Open Sans from google fonts, so it's still requested there.

PR #166, opened 2022-02-17
Unregister the `global-styles` from classic themes

### LINKED ISSUE #86: Classic themes: Double global styles
It looks like for classic themes, we now have two copies of the global styles being output

```
$ curl -s https://test.wordpress.org | grep -B1 -E 'body{--wp--preset--color--black'

<style id="global-styles-for-classic-themes">
	body{--wp--preset--color--black: #000000; ....
</style>
--
<style id='global-styles-inline-css' type='text/css'>
body{--wp--preset--color--black: #000000; .... </style>
```

### LINKED ISSUE #87: Search menu item is invalid HTML
The `wp:search` replacement markup appears to be missing an `<li>` tag, which causes invalid HTML.

```
$ curl https://test.wordpress.org/ -s | grep 'wp:search' -B2 -A37

<ul class="wp-block-navigation__submenu-container">
			......
			<form
				role="search"
				method="get"
				action="https://wordpress.org/search/do-search.php"
				class="wp-block-search__button-outside wp-block-search__text-button global-header__search-form wp-block-search"
			>
....
			</form>
			</ul></li></ul>
```
(output trimmed to relevant lines)

It appears that `wp:navigation-link` is failing to output the `<li>`s with the inner `wp:html`?

https://github.com/WordPress/wporg-mu-plugins/blob/4a227690dd194a08e88f8eeaba5faebd8e164b19/mu-plugins/blocks/global-header-footer/header.php#L81-L124

ryelle:
Fixes #86  — By default, Gutenberg outputs some basic presets (font sizes, colors, etc). The News theme overwrites a few of them (by virtue of using the same slugs), like font-sizes. In the header & footer, the font size is set to `--wp--preset--font-size--small`, which should be 14px (set by News), but the default `global-styles` code resets it to 13px on classic themes.

The news styles output in `global-styles-for-classic-themes` includes the gutenberg default presets, so we can safely remove the second set of custom property definitions from classic themes.

To test:

- Visit a classic theme (ex, learn) on desktop
- The header & footer font sizes should be 14px
- Visit News
- The header & footer font sizes should be 14px
- There should be no other visible changes on either News or classic themes

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -13,6 +13,7 @@ add_action( 'enqueue_block_assets', __NAMESPACE__ . '\register_block_types_js' )
 add_filter( 'wp_enqueue_scripts', __NAMESPACE__ . '\register_block_assets', 200 ); // Always last.
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_compat_wp4_styles', 5 ); // Before any theme CSS.
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_fonts' );
+add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\unregister_classic_global_styles', 20 );
 add_action( 'wp_head', __NAMESPACE__ . '\preload_google_fonts' );
 add_filter( 'style_loader_src', __NAMESPACE__ . '\update_google_fonts_url', 10, 2 );
 add_filter( 'render_block_core/navigation-link', __NAMESPACE__ . '\swap_submenu_arrow_svg' );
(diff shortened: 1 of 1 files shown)


coreymckrill (approved):
👍 Worked in testing on sandbox

ryelle:
This has the side-effect of removing all the `has-*` classes, but https://github.com/WordPress/wporg-mu-plugins/pull/166/commits/666e2ef7edce38042740614fda9ec8e4211ff782 brings them back.

Tested by going to a page with the default styles, ex the "January" button on this page [Keeping up with Gutenberg: Index 2022 ](https://make.wordpress.org/core/handbook/references/keeping-up-with-gutenberg-index/).

| Before | After |
|--------|------|
|  |  |

PR #172, opened 2022-02-24
Add an mu-plugin for tweaking the Stream plugin to meet our needs

coreymckrill:
For now this replaces Stream's posts connector with our own modified version so that it will log post changes correctly when made in the block editor. Other tweaks may follow.

This will help https://github.com/WordPress/pattern-directory/issues/57 move forward.


dd32:
Can the custom posts connector extend the Stream one and just overwrite the minimal things needed?

I'm concerned that any upstream fixes to the connector won't get applied to this class when needed.

> so that it will log post changes correctly when made in the block editor

I thought this was fixed via https://github.com/xwp/stream/pull/1264 ?

coreymckrill:
> I thought this was fixed via https://github.com/xwp/stream/pull/1264 ?

Is was, but that change hasn't made it into a version release yet 😕 

> Can the custom posts connector extend the Stream one and just overwrite the minimal things needed?

Yeah, I guess that might be a better path in the short term, just to get the logging working. The posts connector is... lacking in other ways, so I figure we might replace it entirely, or at least big chunks of it, eventually. But that's probably a separate, bigger conversation. Let me go back and change it to a child class of the Stream one for now...

dd32:
> Is was, but that change hasn't made it into a version release yet 😕

Oh :/ It was merged moments after the 3.8.2 release I guess.. @kasparsd Any chance of getting Stream master merged to plugins.svn trunk?  or releasing a new version?

coreymckrill:
^ Ok, simplified it to a child class that only overrides the one property.

ryelle (approved):
I've tested this with Stream 3.8.2 locally, and it captures patterns created from the frontend, and changes made in the wp-admin block editor. I think we should merge this, and if/when Stream updates we can remove it.

kasparsd:
I've tagged Stream 3.9.0 a few moments ago which now includes the fix that was merged a while ago. There was an issue with unit tests which took me a bit longer to resolve.

coreymckrill:
Thank you @kasparsd !

PR #173, opened 2022-03-01
Add a CDNise mu-plugin

dd32:
This adds a CDNisation plugin that loads assets from s.w.org instead of from WordPress.org in production situations.

It's draft while pending https://make.wordpress.org/systems/2022/02/22/cors-headers-for-s-w-org/ to allow the removal of the blocked list of script/style handles that cannot currently load due to CORS issues.


mu-plugins/cdn/cdn-assets.php
```diff
@@ -0,0 +1,123 @@
...
+
+	// These scripts/styles break if loaded from the CDN without CORS headers.
+	// See https://make.wordpress.org/systems/2022/02/22/cors-headers-for-s-w-org/
+	$blocked_handles = [
+		'wporg-news-style', // Mask images
+		'wporg-global-fonts-css', 'jetpack-icons', 'wp-stream-icons', // Fonts
+	];
+	if ( in_array( $handle, $blocked_handles, true ) ) {
+		$use_cdn = false;
+	}
```
dd32:
```suggestion
```

Remove the block list, as https://make.wordpress.org/systems/2022/02/22/cors-headers-for-s-w-org/ results in the mask images and fonts serving correctly.

Issue #182, opened 2022-03-11
RFC: Add a REST endpoint for accessing public user data across the whole network


coreymckrill:
As described in https://github.com/WordPress/pattern-directory/issues/431, there are some situations on wporg sites where we have a user ID and we want to retrieve additional info about the user from the global user table, but the current `wp/v2/users` endpoint won't work, because the user isn't actually a member of that particular site.

There's some [prior art](https://core.trac.wordpress.org/ticket/39544) exploring a similar issue from a Core perspective, but I think for our purposes, we don't need that level of complexity. All we need is a read-only endpoint that will give back basically the same data as the normal `users` endpoint, but for any row of the global users table. The `link` field would substitute the user's URL on profiles.wordpress.org for the normal author archive link.

`mu-plugins` seems like a good place for this because then the endpoint would be available on all sites on the network, rather than having to make "external" requests to one particular site.

Questions:

* Are there any privacy considerations? This wouldn't expose anything that the normal `users` endpoint doesn't, but maybe it would make scraping user data easier?
* One alternative to adding a new REST endpoint might be to modify the existing endpoint so that it can access users who aren't members of the site. But would that have unintended side effects if, e.g. WP has some code that makes the assumption that only site members will be returned?

iandunn:
This seems like a reasonable idea to me. It'd be less fragile than modifying `wp/v2/users`. 

The Core ticket suggests restricting it to users with `manage_network_capabilities`, but I'm guessing that we need it to work for any user, and maybe even unauthenticated ones?

I'm also guessing this would only add `users/{ID}` and not `users/`, right? If so, then our rate limiting would hinder scraping, but not prevent it. It's probably already possible to get most of the users today anyway, so making it a little easier doesn't seem like a big deal, since there's no tangible exposure.

coreymckrill:
> The Core ticket suggests restricting it to users with manage_network_capabilities, but I'm guessing that we need it to work for any user, and maybe even unauthenticated ones?

Yes. There are at least some use cases, like the Pattern Creator, where we'd want it to work for users who have no role on the site. I'm not actually sure about unauthenticated requests though.

> I'm also guessing this would only add users/{ID} and not users/, right?

Yes. Although recently I was contemplating adding a limited `users/` endpoint that would only work with the `include` parameter, so you could get more than one user record at a time, but you'd still have to specify user IDs.

ryelle:
[...]

PR #184, opened 2022-03-11
Add a REST API endpoint for wporg users

coreymckrill:
This is a proof-of-concept for #182. It simply clones the Core users controller, but then strips out all the endpoints except the read access ones. The single user endpoint takes a slug instead of a numerical ID. The multiple users endpoint requires adding the `slug` query parameter and listing the specific user slugs you want to retrieve.

Additionally this controller limits the request context to either `view` or `embed` (omitting `edit`) so that only data fields that are already public are available in the response.

Another thing we could do here is add in some fields specifically from profiles.wordpress.org. Things that are viewable at a user's public profile page, like their bio, whether they are sponsored, maybe their badges?

## To test

1. Best to do this on a wporg sandbox.
2. Upload the files and then while sandboxed, go to `https://wordpress.org/wp-json/wporg/v1/users/{your wporg user slug}` in your browser.
3. Try adding the context parameter `?context=embed`. That should show identical info to the default response. Changing `embed` to `edit` or anything else should give you an error message.
4. Now try going to `https://wordpress.org/wp-json/wporg/v1/users?slug={your wporg user slug}`
5. It should be the same as the first endpoint, except returns an array with one item.
6. Try adding other user slugs, separated by commas.
7. Compare the data shown to something like `https://wordpress.org/wp-json/wp/v2/users`. It should be all the same fields.
8. Trying to retrieve user data via fields other than slug (e.g. id) shouldn't work.

### DIFF
--- a/mu-plugins/loader.php
+++ b/mu-plugins/loader.php
@@ -5,5 +5,6 @@
 
 require_once __DIR__ . '/blocks/global-header-footer/blocks.php';
 require_once __DIR__ . '/global-fonts/index.php';
+require_once __DIR__ . '/rest-api/index.php';
 require_once __DIR__ . '/skip-to/skip-to.php';
 require_once __DIR__ . '/stream-tweaks/index.php';
--- a/mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace WordPressdotorg\MU_Plugins\REST_API;
+
+/**
+ * Users_Controller
+ */
+class Users_Controller extends \WP_REST_Users_Controller {
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->namespace = 'wporg/v1';
+	}
+
+	/**
+	 * Registers the routes for users.
+	 *
+	 * At this time, this endpoint is exclusively read-only. Other routes from the parent class have been omitted.
+	 *
+	 * @see register_rest_route()
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<id>[\d]+)',
+			array(
+				'args'   => array(
+					'id' => array(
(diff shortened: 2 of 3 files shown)


ryelle (approved):
> Changing embed to edit or anything else should give you an error message.

I tried view first, and `view` is allowed (that seems intentional though) — edit and other arbitrary strings do error. I noticed `view` (or no context) gives me an extra property on each user, `meta`. It's an empty array for the few users I tested. I assume it's because we don't have any registered meta on users.

Also tried querying on other properties with a slug, and it worked as expected— I couldn't `include` or `exclude` users, or query by anything but the slug. Order by name worked. I was able to restrict the fields with `_fields=id,name`, but couldn't add extra fields.

TIL there are users with the names `"1"` and `"2"`. Otherwise trying to pass random IDs correctly errors 🙂 

Overall, I like this as a first iteration - we might want to add onto it later, but it's set up so we can. 💯

coreymckrill:
> TIL there are users with the names "1" and "2".

🤦

iandunn:
How valuable is it to extend `WP_REST_Users_Controller`? I'm wondering if some future changes there could have unexpected implications, since we're using it in a context it's not designed for.

If you want to err on the side of caution you could remove that and make this only return the specific fields that we currently have use cases for.

I haven't seen any tangible problems as a result of extending, though.

coreymckrill:
> How valuable is it to extend WP_REST_Users_Controller? I'm wondering if some future changes there could have unexpected implications, since we're using it in a context it's not designed for.

Good question. The use cases we have for this endpoint right now (pattern directory, internal notes) expect responses to use the normal users endpoint schema, and have (some of) the same collection parameters. It would be simplest to ensure any future use cases also expect the same schema, so that if this endpoint ever gets supplanted by [something in Core](https://core.trac.wordpress.org/ticket/39544), everything would probably still work seamlessly. So, with that in mind, I think it makes sense to extend `WP_REST_Users_Controller` because it already defines the schema and collection parameters, and has some helper methods related to those.

iandunn:
Yeah, that probably outweighs the risks 👍🏻

iandunn (approved):
I couldn't find any way to get private data 👍🏻

PR #190, opened 2022-03-21
Skip adding current item to downloads button.

### LINKED ISSUE #187: Mobile Get WP button not styled correctly on Downloads page
It looks fine on most pages, but on the Downloads page it looks off because of `.current-menu-item`:

Untitled

StevenDufresne:
Fixes #187 

Skip adding the page menu indicator `current-menu-item` to the download button when on the `/downloads` page.

Props @vijayhardaha

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -734,7 +734,8 @@ function get_download_url() {
 	}
 
 	if ( ! $url ) {
-		$url = 'https://wordpress.org/download/';
+		$host = strtolower( $_SERVER['HTTP_HOST'] );
+		$url  = "https://$host/download/";
 	}
 
 	return $url;
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -734,8 +734,7 @@ function get_download_url() {
 	}
 
 	if ( ! $url ) {
-		$host = strtolower( $_SERVER['HTTP_HOST'] );
-		$url  = "https://$host/download/";
+		$url = 'https://wordpress.org/download/';
 	}
 
 	return $url;
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -975,12 +975,6 @@ function set_current_item_class( $menu_items ) {
 
 	foreach ( $menu_items as & $item ) {
 		$sub = false;
-
-		// The download menu item is a button, we don't highlight it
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -975,6 +975,12 @@ function set_current_item_class( $menu_items ) {
 
 	foreach ( $menu_items as & $item ) {
```
dd32, line 981: The check could probably be done outside of the loop, to reduce the calls to `get_download_url()`:

`if ( get_download_url() === $current_url ) return $menu_items;`

The only times when this inner item would be needed is if it's going to highlight the downloads url, which means it must be the `$current_url`, right?

vijayhardaha:
Needs one correction. store `get_download_url()` outside the loop in variable. No need to call on each loop.
```php
$download_url = get_download_url();
```

mu-plugins/blocks/global-header-footer/blocks.php
StevenDufresne: Yep that sounds good.

In order to get the `$current_url` working locally, I updated 

```
if ( ! $url ) {
	$url = 'https://wordpress.org/download/';
}
```
**to**

```
if ( ! $url ) {
	$host = strtolower( $_SERVER['HTTP_HOST'] );
	$url  = "https://$host/download/";
}
```

On paper, this seems like an innocuous change but can i be missing something? I don't mind backing it out if it was intentional.

dd32: Should be right, but given the rest of the file has hard-coded URLs for menus, I'm not sure it makes much difference.

The change would mean that on Rosetta sites, if it can't find a downloads page for some reason, it'll load a 404 `/downloads` instead of reverting to using the primary page.

iandunn: Do local envs have a `/download` page? I didn't think they do

StevenDufresne: No, they don't. I'll just remove it.

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -973,6 +973,9 @@ function fetch_global_styles() {
 function set_current_item_class( $menu_items ) {
 	$current_url = get_menu_url_for_current_page( $menu_items );
 
+	// The download menu item is a button, we don't highlight it
+	if ( get_download_url() === $current_url ) return $menu_items;
```
dd32: Coding standards would be to not have one-liners, despite it being exactly what I suggested :D 

```suggestion
	if ( get_download_url() === $current_url ) {
		return $menu_items;
	}
```

PR #192, opened 2022-03-30
Add a "plugin tweaks" module, and add a tweak for Internal Notes

coreymckrill:
The main mu-plugins directory on wporg is cluttered with files that are for one or more "tweaks" to a particular plugin. The goal with this PR is to not proliferate the same pattern in our pub-sync mu-plugins by stashing all such tweaks in a "plugin-tweaks" subdirectory. Eventually we could also clean up the main mu-plugins directory by migrating open-sourcable tweaks to this new place.

The first tweak for this new system is for the Internal Notes plugin. It started out as [a PR](https://github.com/WordPress/pattern-directory/pull/463) for the pattern directory, but after some discussion it seems better to put it in mu-plugins since it will benefit any use of the Internal Notes plugin anywhere on wporg.

Note that with this PR I haven't moved the existing "stream-tweaks" in this repo into "plugin tweaks". This is because that particular tweak is no longer needed, and will be removed in a separate PR.

For reviewing this PR, I'm more interested in feedback about the implementation of `plugin-tweaks` than in the particular Internal Notes tweak, which I believe was already tested in the other PR. E.g. would it be better to have each tweaked plugin have its own file even if there's only one function?

### DIFF
--- a/mu-plugins/loader.php
+++ b/mu-plugins/loader.php
@@ -5,6 +5,7 @@
 
 require_once __DIR__ . '/blocks/global-header-footer/blocks.php';
 require_once __DIR__ . '/global-fonts/index.php';
+require_once __DIR__ . '/plugin-tweaks/index.php';
 require_once __DIR__ . '/rest-api/index.php';
 require_once __DIR__ . '/skip-to/skip-to.php';
 require_once __DIR__ . '/stream-tweaks/index.php';
--- a/mu-plugins/plugin-tweaks/index.php
+++ b/mu-plugins/plugin-tweaks/index.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * Tweaks for specific plugins on WordPress.org.
+ *
+ * If there's just one tweak for a plugin, feel free to add it here on the index file. If there's more than
+ * one, or multiple functions/files are needed, consider adding a separate file or directory for that plugin
+ * and loading those files from here.
+ *
+ * When adding a function here, please prefix it with the plugin's name.
+ */
+
+namespace WordPressdotorg\MU_Plugins\Plugin_Tweaks;
+
+defined( 'WPINC' ) || die();
+
+/**
+ * Actions and filters.
+ */
+add_filter( 'wporg_internal_notes_rest_prepare_response', __NAMESPACE__ . '\wporg_internal_notes_replace_rest_author_link' );
+
+/**
+ * Replace the Internal Notes default embeddable author link with one from the wporg endpoint.
+ *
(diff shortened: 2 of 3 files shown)


mu-plugins/loader.php
```diff
@@ -5,6 +5,7 @@
 
 require_once __DIR__ . '/blocks/global-header-footer/blocks.php';
 require_once __DIR__ . '/global-fonts/index.php';
+require_once __DIR__ . '/plugin-tweaks/index.php';
```
iandunn, line 8: > It started out as https://github.com/WordPress/pattern-directory/pull/463 for the pattern directory, but after some discussion it seems better to put it in mu-plugins since it will benefit any use of the Internal Notes plugin anywhere on wporg.

I'm leery of trying too hard to support 3rd party usage after our experience w/ CampTix. If we put this directly in the Notes plugin, then 3rd parties could disable it w/ `remove_filter()`.

IMO it'd be better to not have tweaks for 1st party plugins, b/c I worry I'll forget they're there and only look in the main plugin itself.

I don't feel strongly, though, whatever you decide WFM.

mu-plugins/plugin-tweaks/index.php
```diff
@@ -0,0 +1,49 @@
+<?php
```
iandunn, line 1: >  would it be better to have each tweaked plugin have its own file even if there's only one function?

I'd lean towards that. Over time the file will get big if there's only one, and become unwieldy. We could break off a few separate files for groups of tweaks at that point, but then `index.php` wouldn't be named descriptively. 

Having a separate file also makes it show up in IDE searches, which can help devs learn about it if they didn't already know (or forgot) that it exists. 

So something like:

```
mu-plugins/plugin-tweaks/jetpack.php
mu-plugins/plugin-tweaks/gutenberg.php
mu-plugins/plugin-tweaks/camptix.php
```

Not a big deal either way though.

iandunn (approved):
>  not proliferate the same pattern ... stashing all such tweaks in a "plugin-tweaks" subdirectory

💯

mu-plugins/plugin-tweaks/index.php
coreymckrill: Good points. I think this is a better approach.

mu-plugins/loader.php
coreymckrill: > I'm leery of trying too hard to support 3rd party usage after our experience w/ CampTix.

Me too. Definitely not thinking we should release this as a standalone into the plugin directory. That said, it feels cleaner and less brittle to me if we avoid tight coupling between wporg and plugins when we can, and to put integration code in an mu-plugin or some other separate plugin.

PR #200, opened 2022-04-07
Remove the correct actions to fix the Header/Footer with Trunk/Gutenberg 12.9

dd32:
Is what ultimately caused https://meta.trac.wordpress.org/ticket/6261

PR for merge/testing purposes.


dd32:
Props @StevenDufresne for review :)

Issue #196, opened 2022-04-01
In support page in responsive width around 890px the hamburger menu navbar has background color issue.


Smit2808:
when the hamburger menu starts displaying on a width less than 890px, the menu has different bg color on top. And it's looking bad. I am attaching an image for better understanding.
Steps:
1. Open https://make.wordpress.org/support/ in responsive view width less than 890px.
2. Then open the hamburger menu and you can clearly see the background color issue on top of the menu, its background color is different from the menu color.

Image:
<img width="1440" alt="Screenshot 2022-04-01 at 12 35 47 PM" src="https://user-images.githubusercontent.com/64188895/161212784-16dd27e8-ac1f-4cbf-bdf0-5f867a16e8a3.png">

iandunn:
Good catch. Looks like `wp-block-navigation__responsive-container-content` needs `padding-top: 0` or something similar.

Smit2808:
Hi @iandunn sir,

As per your suggestion, I have created this pull request: https://github.com/WordPress/wporg-mu-plugins/pull/198/commits/63508474de6d5eef0a83130e6ac1a2cc96c01243 . I removed the padding top from the menu.

so this will work. But what I suggest is to change the background color and keep the padding-top as it is. If we change the background color like menu then it will look good. The color issue will be gone and padding will also be there as it is. So please if you like this suggestion then tell me. I will make PR for this.

iandunn:
🤔 That's a good point, removing the background from `wp-block-navigation__container` also works, and we'll want some padding there to match [the mockup](https://www.figma.com/file/O9I8Jl7kaUMTX39LlRGbGO/News-Page?node-id=7413%3A31182) (which isn't totally up to date).

It looks like we didn't use to have that style, but it was added in #178 for browser support. @StevenDufresne , do you have any thoughts on removing it, or changing it to `var(--wp--preset--color--darker-grey)` ?


_( Note: This can be reproduced in the local News repo too, not just on the Make sites. )_

StevenDufresne:
> It looks like we didn't use to have that style, but it was added in #178 for browser support. @StevenDufresne , do you have any thoughts on removing it, or changing it to `var(--wp--preset--color--darker-grey)` ?

Ah, I see, definitely a regression. We need the background color because of inheritance problems in older browsers. We can probably just use a media query and turn on that background color for our desktop view.  Removing the padding brings the clickable logo and the plugin link closer together than we probably want.

[...]

PR #201, opened 2022-04-07
added media query for bg color only for desktop view

Smit2808:


mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
```diff
@@ -232,9 +232,10 @@
 
 	See: https://github.com/WordPress/wporg-mu-plugins/issues/159
 	*/
-
-	& .wp-block-navigation__container {
-		background: var(--wp--preset--color--dark-grey);
+	@media (--tablet) {
+		& .wp-block-navigation__container {
+			background: var(--wp--preset--color--dark-grey);
+		}
```
iandunn, line 239: I like to put the media query inside the element. That way default mobile styles can go in there too without having to create a duplicate block. eg.,

```postcss
& .wp-block-navigation__container {
	/* other stuff can go here in the future */

	@media (--tablet) {
		background: var(--wp--preset--color--dark-grey);
	}
}
```

iandunn:
fixes #196

Smit2808:
Hi @iandunn

Thank you for reviewing my PR. As you have added the "fixes" in the latest comment, So now do I need to change anything as you said in the first comment to add media query inside the element?

iandunn:
Ah, no, that was just to [link the PR to the issue](https://github.blog/2013-01-22-closing-issues-via-commit-messages/).

I think what's remaining here is just to move the media query inside the element, and then hopefully @StevenDufresne can test in Falkon since he's already got a VM setup for that.

StevenDufresne:
I have preemptively tested the changes that @iandunn is requesting. Works as expected 👍. If you can update the PR, we can merge it in!

PR #204, opened 2022-05-12
Add PHPUnit tests for mu-plugins

dd32:
This PR adds unit tests to #173

I've separated it due to the amount of work I had to do to make it work.

`wp_get_environment_type()` is not exactly designed to be used during unit testing, and while it allows for it to be bypassed, `wp-env` forces phpunit to be a `local` environment.

Draft while I figure out if it's viable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,11 +17,9 @@
 			"url": "git@github.com:WordPress/wporg-repo-tools.git"
 		}
 	],
-	"require": {
-	},
 	"require-dev": {
 		"composer/installers": "~1.0",
-		"phpunit/phpunit": "^9.5"
+		"phpunit/phpunit": "^9.5",
 		"wporg/wporg-repo-tools": "dev-trunk"
 	},
 	"config": {
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,10 @@
 			"url": "git@github.com:WordPress/wporg-repo-tools.git"
 		}
 	],
+	"require": {
+		"composer/installers": "~1.0"
+	},
 	"require-dev": {
-		"composer/installers": "~1.0",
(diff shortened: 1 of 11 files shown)


dd32:
Aside from the hacky `sed "/WP_ENVIRONMENT_TYPE/d" -i ....` within the action setup, I think this is working well enough that it's worth committing and expanding upon over time for other plugins.

An upstream issue to allow removing the hacky hack above is here: https://github.com/WordPress/gutenberg/issues/41083

composer.json
```diff
@@ -21,7 +21,10 @@
 		"composer/installers": "~1.0"
 	},
 	"require-dev": {
-		"wporg/wporg-repo-tools": "dev-trunk"
+		"phpunit/phpunit": "^9.5",
```
dd32, line 24: I'm unsure if this is actually needed here, I suspect it's not, as it would only be needed if you're running `phpunit` outside of `wp-env`.

The below wp-phpunit & yoast deps are however required.

phpunit/test-cdn-assets.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+class Test_CDN_Assets extends WP_UnitTestCase {
+
```
dd32, line 4: This needs a setUp/tearDown method to check/reset the environmental WP_ENVIRONMENT_TYPE value.

dd32:
> @[dd32](https://github.com/dd32) dd32 deleted the branch WordPress:trunk [6 minutes ago](https://github.com/WordPress/wporg-mu-plugins/pull/204#event-10152447227)

Ha. No I didn't. I changed the base target of this PR, the _old base branch for the PR was deleted_.

PR #205, opened 2022-05-16
Users: Add a filter to alter the author link on post-type responses to use our custom endpoint

dd32:
Currently post endpoints link to a non-accessible endpoint for posts, eg:

```
$ curl -s https://wordpress.org/photos/wp-json/wp/v2/photos/9781 | jq ._links.author[0].href
"https://wordpress.org/photos/wp-json/wp/v2/users/14322318"

$curl -s https://wordpress.org/photos/wp-json/wp/v2/users/14322318 | jq .message
"Invalid user ID."

$ curl -s 'https://wordpress.org/photos/wp-json/wp/v2/photos/9781?_embed=true' -s | jq ._embedded.author[0].message
"Invalid user ID."
```

After this change:
```
$ curl -s https://wordpress.org/photos/wp-json/wp/v2/photos/9781 | jq ._links.author[0].href
"https://wordpress.org/photos/wp-json/wporg/v1/users/femkreations"

$ curl -s https://wordpress.org/photos/wp-json/wporg/v1/users/femkreations | jq .name
"Femy Praseeth"

$ curl -s 'https://wordpress.org/photos/wp-json/wp/v2/photos/9781?_embed=true' -s | jq ._embedded.author[0].name
"Femy Praseeth"
```

See https://meta.trac.wordpress.org/ticket/6313

### DIFF
--- a/mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
@@ -58,6 +58,59 @@ class Users_Controller extends \WP_REST_Users_Controller {
 				'schema' => array( $this, 'get_public_item_schema' ),
 			)
 		);
+
+		// Filter the post endpoints to alter the public URL for non-authenticated users.
+		$post_types = get_post_types(
+			array(
+				'public'       => true,
+				'show_in_rest' => true,
+			)
+		);
+
+		foreach ( $post_types as $post_type ) {
+			add_filter( "rest_prepare_{$post_type}", array( $this, 'alter_post_author_link' ) );
+		}
+	}
+
+	/**
+	 * Change the author link for posts to this endpoint.
+	 * 
+	 * @param \WP_REST_Response $response Response object for the request.
+	 * 
+	 * @return \WP_REST_Response Response object for the request.
+	 */
+	public function alter_post_author_link( $response ) {
+		$links = $response->get_links();
+		if (
+			empty( $links['author'][0]['href'] ) ||
+			! preg_match( '!wp/v2/users/(?P<id>\d+)!i', $links['author'][0]['href'], $m )
+		) {
+			return $response;
+		}
+
+		$user_id = absint( $m['id'] );
+
+		// If the core users endpoint will work for this user, no need to change it.
+		if ( is_multisite() && is_user_member_of_blog( $user_id ) ) {
+			return $response;
+		}
+
+		// Get the user
+		$user = get_user_by( 'id', $user_id );
+
+		// Remove the existing author link
+		$response->remove_link( 'author', $links['author'][0]['href'] );
+
+		// Add a link to our endpoint instead
+		$response->add_link(
+			'author',
+			rest_url( $this->namespace . '/' . $this->rest_base . '/' . urlencode( $user->user_nicename ) ),
+			array(
+				'embeddable' => true
+			)
+		);
+
+		return $response;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
```diff
@@ -58,6 +58,59 @@ public function register_routes() {
...
+	public function alter_post_author_link( $response ) {
+		$links = $response->get_links();
+		if (
+			empty( $links['author'][0]['href'] ) ||
+			! preg_match( '!wp/v2/users/(?P<id>\d+)!i', $links['author'][0]['href'], $m )
+		) {
+			return $response;
+		}
+
+		$user_id = absint( $m['id'] );
```
dd32: In retrospect, this may be cleaner than using regular expressions to extract the author ID..

```suggestion
			add_filter( "rest_prepare_{$post_type}", array( $this, 'alter_post_author_link' ), 10, 2 );
		}
	}

	/**
	 * Change the author link for posts to this endpoint.
	 * 
	 * @param \WP_REST_Response $response Response object for the request.
	 * @param \WP_Post $post The post the $response represents.
	 * 
	 * @return \WP_REST_Response Response object for the request.
	 */
	public function alter_post_author_link( $response, $post ) {
		$links = $response->get_links();
		if (
			empty( $links['author'][0]['href'] ) ||
			! str_contains( $links['author'][0]['href'], 'wp/v2/users/' ) ||
			! $post->post_author
		) {
			return $response;
		}

		$user_id = $post->post_author;
```

coreymckrill: It might be even simpler to just check if the response has an `author` property. Something like:

```php
$data = $response->get_data();
$user = get_user_by( 'id', $data['author'] ?? 0 );
if ( ! $user->exists() ) {
	return $response;
}
```

mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
```diff
@@ -58,6 +58,59 @@ public function register_routes() {
...
+		// If the core users endpoint will work for this user, no need to change it.
+		if ( is_multisite() && is_user_member_of_blog( $user_id ) ) {
+			return $response;
+		}
+
+		// Get the user
+		$user = get_user_by( 'id', $user_id );
+
+		// Remove the existing author link
+		$response->remove_link( 'author', $links['author'][0]['href'] );
```
coreymckrill: Could this be simplified to just `$response->remove_link( 'author' );`?

coreymckrill (approved):
I did something similar, but only for the Internal Notes plugin, [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/plugin-tweaks/wporg-internal-notes.php#L12-L40).

That's a special case, since it doesn't use the normal posts rest controller, but I think this is the right approach for post types that do use it.

mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
dd32: Looks like it, I didn't realise you could skip passing the 2nd parameter! 

```suggestion
		$response->remove_link( 'author' );
```

mu-plugins/rest-api/endpoints/class-wporg-rest-users-controller.php
```diff
@@ -58,6 +58,59 @@ public function register_routes() {
...
+		$user_id = absint( $m['id'] );
+
+		// If the core users endpoint will work for this user, no need to change it.
+		if ( is_multisite() && is_user_member_of_blog( $user_id ) ) {
+			return $response;
+		}
+
+		// Get the user
+		$user = get_user_by( 'id', $user_id );
+
```
dd32: [From corey](https://github.com/WordPress/wporg-mu-plugins/pull/205/files#r874056178)

```suggestion
		$data = $response->get_data();
		$user = get_user_by( 'id', $data['author'] ?? 0 );
		if ( ! $user->exists() ) {
			return $response;
		}

		// If the core users endpoint will work for this user, no need to change it.
		if ( is_multisite() && is_user_member_of_blog( $user->ID ) ) {
			return $response;
		}

```

PR #210, opened 2022-06-16
Header: Switch to submenu blocks for navigation items with submenus

### LINKED ISSUE #207: Navigation sub menus not totally accessible via the keyboard
While viewing a site such as https://make.wordpress.org, the main navigation contains sub menus. An example of such sub menu is Support. It has Documentation and Forums links contained within.

Here are the issues that need to be fixed to make the sub menus accessible.
1. [ ] Sub menus should not open on focus. This can become some super confusing UX to hear you have been thrown in to a new list of links.
2. [ ] If Support has a sub menu, there should be a button directly after the Support link that says Open Support Sub Menu.
3. [ ] The above button should receive `aria-expanded="false"` as the default and toggle to true when the sub menu is visible.
4. [ ] If inside a sub menu, pressing Escape should close the sub menu and move focus to the trigger that opened it.

Thanks to @courtneyr-dev for letting me know about the sub menus. I otherwise would have missed this unless I somehow found it by mistake.

ryelle:
Fixes #207 - Menu items with submenus should use the `navigation-submenu` block, this provides the expected markup for accessibility:

- Submenus don't open on focus.
- Items with a submenu (Support & Get Involved) have buttons following that say "Support submenu"/"Get Involved submenu".
- The button has `aria-expanded="false"` by default, and `true` when open

The remaining item on the list in #207 is this:

> If inside a sub menu, pressing Escape should close the sub menu and move focus to the trigger that opened it.

This menu doesn't seem to do that, but this doesn't seem to be core behavior either. I'd recommend opening an issue in gutenberg for the navigation block if this is required.

Props @alexstine.

**Screenshots**

In both of the videos below, I tab through the menu. In "before" you can see I tab down into the submenus with no other action. After, I need to activate the 🔽  button. There's no visual change, and no change in behavior with a mouse.

https://user-images.githubusercontent.com/541093/174152708-901a580b-468d-4401-93a0-a441ca8d3d74.mp4

https://user-images.githubusercontent.com/541093/174152715-28aa70fb-62ac-4ddb-ad4d-f7dc1aefccce.mp4

| Before | After |
|--------|------|
|  |  |
|  |  |

### DIFF
--- a/mu-plugins/blocks/global-header-footer/header.php
+++ b/mu-plugins/blocks/global-header-footer/header.php
@@ -28,7 +28,7 @@ $search_args = array(
 );
 
 /**
- * Ouput menu items (`navigation-link`) & submenus (`navigation-submenu`). If a submenu, recursively iterate
+ * Output menu items (`navigation-link`) & submenus (`navigation-submenu`). If a submenu, recursively iterate
  * through submenu items to output links.
  *
  * @param array   $menu_item An item from the array in `get_global_menu_items` or `get_rosetta_menu_items`.
@@ -49,7 +49,7 @@ function recursive_menu( $menu_item, $top_level = true ) {
 		);
 	}
 
-	$ouput = sprintf(
+	$output = sprintf(
 		'<!-- wp:navigation-submenu {"label":"%1$s","url":"%2$s","kind":"%3$s","className":"%4$s"} -->',
 		$menu_item['title'],
 		$menu_item['url'],
@@ -58,12 +58,12 @@ function recursive_menu( $menu_item, $top_level = true ) {
 	);
 
 	foreach ( $menu_item['submenu'] as $submenu_item ) {
-		$ouput .= recursive_menu( $submenu_item, false );
+		$output .= recursive_menu( $submenu_item, false );
 	}
 
-	$ouput .= '<!-- /wp:navigation-submenu -->';
+	$output .= '<!-- /wp:navigation-submenu -->';
 
-	return $ouput;
+	return $output;
 }
 
 ?>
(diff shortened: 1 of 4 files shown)


StevenDufresne:
When the menu is truncated, the submenu is not reachable:

https://user-images.githubusercontent.com/1657336/174197467-5b3d5b14-fe4e-49ca-a049-5c1cf9ba1e95.mp4

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -26,6 +27,45 @@
...
+			'<!-- wp:navigation-link {"label":"%1$s","url":"%2$s","kind":"%3$s","isTopLevelLink":%4$s,"className":"%5$s"} /-->',
+			$menu_item['title'],
+			$menu_item['url'],
+			$menu_item['type'],
+			$top_level ? 'true' : 'false',
+			$menu_item['classes'] ?? '',
+		);
+	}
+
+	$ouput = sprintf(
```
StevenDufresne: I think this is a typo: `$output` maybe?

alexstine:
Thanks @ryelle ! I opened a PR upstream in Gutenberg which should fix up Escape key functionality.

mu-plugins/blocks/global-header-footer/header.php
ryelle: 🤦🏻

ryelle:
> When the menu is truncated, the submenu is not reachable

Ah, I bet when the menu item is cloned, the toggle looses the event listener. I've updated the style on submenus in the overflow menu so they'll always display to work around that. Also because 3rd-level flyouts are awkward :)



I also updated the `...` link when the menu is truncated to be a button and use the same toggle-on-hover/click pattern. This fixes an issue where screen readers would just announce the link as "dot" (at least, VoiceOver did), now it's "More menu" with the expanded context.

StevenDufresne (approved):
Great, looks good.

PR #211, opened 2022-06-29
Update build process for multiple blocks

ryelle:
The current build process only works for the one block, and only for CSS. This PR adds two new scripts, `bin/build.js` and `bin/watch.js`, to build the JS and CSS for as many projects as we add.

I've made some assumptions in the build script, but I think they'll bear out in our blocks.

1. This only builds files in `blocks` — right now I don't think we'll need building in other directories, but if we do, it wouldn't be hard to update that.
2. CSS in blocks should not use Sass, they're processed with PostCSS instead — this keeps things consistent from `global-header-footer`.
3. The CSS builds each top-level file in `postcss`, unless it starts with a `_` (which is a practice I've pulled from Sass).
4. The JS is build with `webpack` using the webpack config from `@wordpress/scripts` — so it builds from the same `src/index.js` entrypoint
5. New blocks should use the format generated by [the repo-tools mu-block template](https://github.com/WordPress/wporg-repo-tools/pull/21)

I've also tweaked the `postcss.config.js`. I removed the `postcss-custom-media` & `postcss-nesting` plugins, since they're part of `postcss-preset-env` and just needed to be enabled. `postcss-preset-env` now supports [all stage 2+](https://preset-env.cssdb.org/features/#stage-2) features, plus the [nesting feature](https://preset-env.cssdb.org/features/#nesting-rules) (technically a stage 1 feature). `cssnano` is now only added if you're building for production.

The new commands provided here are:

- `npm start`: Start up a watch script, if there are any changes in a block, it rebuilds that block. Files are not minified.
- `npm run build`: Build all blocks. Files are minified.
- `npm run build global-header-footer`: Build just the one block. Files are minified.
- `./bin/build.js`: Build all blocks. Files are not minified.
- `./bin/build.js global-header-footer`: Build just the one blocks. Files are not minified.

Building output:

```bash
$  npm run build        
JavaScript built for new-block
CSS built for new-block
CSS built for global-header-footer
```

Watch output:

```bash
$ npm start

> wporg-mu-plugins@0.0.1 prestart wporg-mu-plugins
> ./bin/build.js

JavaScript built for new-block
CSS built for new-block
CSS built for global-header-footer

> wporg-mu-plugins@0.0.1 start wporg-mu-plugins
> NODE_ENV=development ./bin/watch.js

-> Watching for changes...
-> update: new-block
JavaScript built for new-block
CSS built for new-block
```

If there are 🛑 errors or ⚠️  warnings in the CSS or JS, it should tell you about them. Errors will prevent building for that project, but others will still build, and fixing the error should rebuild automatically when watching.

**To Test**

1. `npm install` to get the new dependencies
2. Delete the `blocks/global-header-footer/build` folder
3. Run any of the above commands, and you should see the build folder come back

Test in combination with https://github.com/WordPress/wporg-repo-tools/pull/21

1. Move into `mu-plugins/blocks`
2. Run `npx @wordpress/create-block -t "/[SOME_PATH]/wporg-repo-tools/templates/mu-block"` to create a new block
3. Grab the `require` from the new block's readme and paste it into `loader.php`
4. Run any of the build commands
5. Open a block editor, you should find your new block
6. Run the watch command
7. Make changes in some files, it should rebuild the correct project


StevenDufresne:
I'm ready & willing to work with this. Go for it!

StevenDufresne:
I'm running into this now. Where will these be built? Are they built here or in the sandbox pre-svn update? If so, I imagine we need to update the sync script.

dd32:
> I'm running into this now. Where will these be built?

My assumption is that it'll be the GitHub action, which merges it to the [build branch](https://github.com/WordPress/wporg-mu-plugins/tree/build). The sync script pulls from that (So you actually have to wait for the build to complete on GitHub before you can sync to dotorg)

dd32:
> I'm running into this now. Where will these be built? Are they built here or in the sandbox pre-svn update? If so, I imagine we need to update the sync script.

@StevenDufresne Figured out what you've been missing :)

https://github.com/WordPress/wporg-mu-plugins/blob/trunk/.github/workflows/build.yml#L25-L31

The `gitignore` rule there needs to be updated to something like `git add mu-plugins/blocks/*/build/* --force`, but `npm run build` also doesn't appear to create the `screenshot-preview` build files yet too :( ...I think if you rename the `js` folder to `src` this new tooling will pick it up?

ryelle:
> I think if you rename the js folder to src this new tooling will pick it up?

Yeah, sorry I didn't spell that out anywhere explicitly, I'd assumed anyone testing this would use the block template. Since the `global-header-footer` JS script doesn't require building, I designed the script to only do a build step if the relevant folder is found:

For each subfolder in `mu-plugins/blocks`…
1. If there is a `src` folder, run the JS build.
2. If there is a `postcss` folder, run the CSS build. It'll build any top-level CSS files (unless they start with `_`).

So your JS should be in `src`, and it uses the `@wordpress/scripts` logic to figure out the entry points (it's [surprisingly complicated](https://github.com/WordPress/gutenberg/blob/2280462a203a2c9458c6af4d2ff132fe570f5b40/packages/scripts/utils/config.js#L181), but if you have a block.json it should catch everything). `src/index.js` is the default.

Any CSS that needs PostCSS processing needs to be in a `postcss` folder.

PR #217, opened 2022-07-07
Fix positioning of global header search box when open

### LINKED ISSUE #216: Fix global header search positioning
The update to Gutenberg 13.6.0 has broken the positioning of the global header search modal.

Screen Capture on 2022-07-07 at 13-09-44

adamwoodnz:
See #216 

Override margin set in navigation block when search box is open


StevenDufresne:
I took this change and added one more in: https://github.com/WordPress/wporg-mu-plugins/commit/abc02533432fb1987ea9b96d5a7fef28c0c51d73.

Thanks for the head start!

PR #222, opened 2022-07-21
CSS: Update class name after gutenberg 13.7 change

ryelle:
The `wp-container-*` class is no longer added to all blocks, instead we should use the `wp-block-*` class. In `_common`, the selector is necessary for specificity, so use `wp-block-*` instead. In `logos`, it's not needed for specificity, so it's been removed.

| Before | After |
|--------|------|
| before-header | after-header |
| before-footer | after-footer |

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -89,12 +89,12 @@ html {
  * @link https://github.com/WordPress/wporg-news-2021/pull/30
  */
 
-[class*="wp-container-"].global-header,
-[class*="wp-container-"].global-footer,
-[class*="wp-container-"].global-header > * + *,
-[class*="wp-container-"].global-footer > * + *,
-.global-header [class*="wp-container-"] > * + *,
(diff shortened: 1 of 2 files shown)


adamwoodnz (approved):
:shipit: LGTM

andrewserong:
Just copy+pasting a similar comment from https://github.com/WordPress/wporg-news-2021/pull/387 for visibility:

---

This PR was required due to the change introduced in https://github.com/WordPress/gutenberg/pull/40875 that refactored how the Layout block support works. Part of the effort in that PR is to reduce the usage of container ID-based classes in favour of semantic classnames, to help address https://github.com/WordPress/gutenberg/issues/41434 and an overall direction in https://github.com/WordPress/gutenberg/issues/38719.

Targeting the `wp-block-` classname prefix looks safe to me, as that particular classname structure is very unlikely to change. For further changes to how the Layout block support works, we're currently tracking work in https://github.com/WordPress/gutenberg/issues/39336. One of the goals is to settle on stable classnames for layout that can be relied upon in themes, and then document these reliable classnames in the handbook. For now, it's great to see that the styling rules here can safely rely upon the `wp-block-` prefix instead!

PR #223, opened 2022-07-22
Change folder: js -> src to match build script.

StevenDufresne:
This PR renames the `js` folder to `src` to match the build script. We could update the build script but `src` makes sense.


dd32:
This plus https://github.com/WordPress/wporg-mu-plugins/commit/9d54f55d437a42bc56349d3dd8f115f11366eb40 works for me on my fork, so I'm going to merge this to get everything finalised before the weekend.

dd32:
I merged too early, failing to note the changes to the global-header-footer blocks.

The JS in that block doesn't require building, and is used as-is.
I moved it back in https://github.com/WordPress/wporg-mu-plugins/commit/5b7941a7af77a45f7a5b91680906d13480e738db which restores the previous fileset for that block.

PR #224, opened 2022-07-25
Improve accessibility for screenshot block.

StevenDufresne:
This PR does 3 things:
- Adds `caption` as screen reader text
- Adds focus state to card
- Updates block to use `postcss` for build

## Adds `caption` as screen reader text
- Imports `screen-reader-text` styles from [wporg]( https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/css/objects/_accessibility.scss). I've done this in case the style is absent. May be overkill.

## Adds focus state to card
 | w/ Light `` | w/ Dark `` |
 | --- | --- |
|  |  | 

## Updates block to use `postcss` for build
- Sets up `rlt` styles even though there currently aren't any differences.


mu-plugins/blocks/screenshot-preview/src/block.js
```diff
@@ -46,7 +46,7 @@ function Block( { link, previewLink, version } ) {
 				height: frameHeight,
 			} }
 			href={ link }
-		>
+		>   { caption && <span className="screen-reader-text">{ caption }</span> }
```
ryelle: Linter fix:

```suggestion
		>
			{ caption && <span className="screen-reader-text">{ caption }</span> }
```

ryelle (approved):
One minor linter issue, but otherwise looks good 👍🏻 

VoiceOver + FF (i can't test sandbox with Safari) correctly read the screen-reader-text as the link label. Tabbing through works correctly, and focus stays in the expected place when opening the pattern, and when loading all patterns.

> Imports screen-reader-text styles from wporg. I've done this in case the style is absent. May be overkill.

It's added by both wporg and the gutenberg block-library CSS. I think it's safe to assume the theme or WordPress will provide it, even if we end up using this block elsewhere. 

Another solution would be to use [VisuallyHidden](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/visually-hidden), which will add the relevant CSS itself.

Issue #195, opened 2022-04-01
Mobile Get WP button not styled correctly on focus


dd32:
Potentially similar to #187 

Originally reported at https://meta.trac.wordpress.org/ticket/6237
![outline-issue-btn](https://user-images.githubusercontent.com/767313/161180378-f0c2d1ee-16cb-4274-85b2-5172767900b0.png)

The `:active` or `:focus` styles from the menu are being applied to the Download button.

aezazs-multidots:
Hello @dd32 

For the solution to this ticket, I have created a PR.
[https://github.com/WordPress/wporg-mu-plugins/pull/199](https://github.com/WordPress/wporg-mu-plugins/pull/199)

Can you please review it?

Thank you.

aezazs-multidots:
Hello, @iandunn and @dd32  . For the solution to this ticket, I have created a PR. https://github.com/WordPress/wporg-mu-plugins/pull/199. Please review it and merge it.

Thank you.

dd32:
@aezazs-multidots Please be patient, Bumping tickets/issues is not welcome. We're aware of the PR, but I've not verified if it's the intended design.

ryelle:
This issue was fixed as part of some CSS cleanup in #225.

PR #225, opened 2022-07-25
Add variations for header & footer styles

### LINKED ISSUE #209: Header: Active indicator should have same width as the menu link background
| Currently | Intended |
| ----- | ----- |
| Captura de ecrã 2022-06-13, às 14 52 02 | Captura de ecrã 2022-06-13, às 14 52 02 |

### LINKED ISSUE #221: Customize the colors in global header & footer
In the latest mockups for the w.org redesign, the header & footer are not always white links on a black background.

The color could probably be controlled by an attribute on the header & footer blocks.

ryelle:
Fixes #221 — Added block variations and a `style` attribute to support the different color combinations. The default block shouldn't change, but now there are options for a "Black on white" and "White on blue." When selected, the style changes, updating the classes on the rendered group block, which controls how the custom properties are defined. I've also added filters so we can control this in PHP on classic themes if needed, `wporg_header_color_scheme` & `wporg_footer_color_scheme`.

Screen Shot 2022-07-25 at 1 48 50 PM

This also introduces a local layer of custom properties for the background, link color, and some related colors.

Since variations are registered in JS, I created two new folders `src/header` and `src/footer`, each with a `block.json`, and updated the PHP registration code to read from the `block.json`s.

While I was in the code, I also fixed #209, so the border is on the container `li` to span the width of the link text and dropdown button.

**Screenshots**

| Default | White on Blue | Black on White |
|---------|------------------|-----------------|
| default-default | default-blue | default-white |

Active states— I combined a few screenshots in one. The Support submenu is hovered open. The Showcase link is focused. The search box is open. The Get WordPress button is hovered. **These (especially the white one) will need design input, but can be done in a followup PR.**

| Default | White on Blue | Black on White |
|---------|------------------|-----------------|
| active-states-default | active-states-blue | active-states-white |

Small screen:

| Default | White on Blue | Black on White |
|---------|------------------|-----------------|
| menu-mobile-default | menu-mobile-blue | menu-mobile-white |
| search-mobile-default | search-mobile-blue | search-mobile-white |

**To Test**

1. If you can, open the block in the site editor (maybe use the parent theme, since News doesn't work in the site editor)
2. Note, if you're trying to select the header, it's got 0 height because of the positioning used, so select it via the List view
3. Select a different variation
4. It should live-update in the editor
5. Save your changes & view on the frontend
6. The header and/or footer should match the style you've picked

Or, update the news template to manually add the attribute:

```html
<!-- wp:wporg/global-header {"style":"black-on-white"} /-->

<!-- wp:wporg/global-footer {"style":"black-on-white"} /-->
```

Or, use the filter:

```php
add_filter( 'wporg_header_color_scheme', __NAMESPACE__ . '\update_color_scheme' );
add_filter( 'wporg_footer_color_scheme', __NAMESPACE__ . '\update_color_scheme' );

function update_color_scheme( $style ) {
	if ( ! $style ) {
		return 'white-on-blue';
	}
	return $style;
}
```


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -776,9 +730,11 @@ function rest_render_global_footer( $request ) {
 /**
  * Render the global footer in a block context.
  *
+ * @param array $attributes The block attributes.
+ *
  * @return string
  */
-function render_global_footer() {
+function render_global_footer( $attributes ) {
```
StevenDufresne, line 737: This doesn't appear to be used, is this a best practice?

StevenDufresne:
I tested this a few ways:

**Connecting this PR to news env.**
✅  When I added any variation of `{"style":"{colorScheme}"}` to the **global-header** element in the template, it worked as expected.

⚠️  When I added any variation of `{"style":"{colorScheme}"}`, to the **global-footer** element in the `index.html` template, nothing happened. Am I doing something wrong maybe?


**Adding to new environment, using block editor**
✅  This worked as expected, the transforms were applied properly.

⚠️  When I add the global header in the block editor, the header styles warp the block editor. The footer does not create the same side effect.

Screen Shot 2022-07-26 at 1 22 59 PM

ryelle:
> to the global-footer element in the index.html template, nothing happened.

Which page were you previewing? The `index.html` template is a fallback, it's only used on non-category archive pages (like dates - `/2021/`). The front page template is `front-page.html`.

> When I add the global header in the block editor, the header styles warp the block editor. 

Hm, we should probably disable adding it in pages. The iframe method in the site editor prevents that issue.

mu-plugins/blocks/global-header-footer/blocks.php
ryelle: It's passed through to `footer.php` where it is used, setting it here sets up the context. It's not super clear, but the same happens in the header function.

ryelle:
I'm pretty sure this also fixes https://github.com/WordPress/wporg-mu-plugins/issues/195

beafialho:
I haven't tested yet but just wanted to let you know we've added these variations specifications to the Library [here](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=194%3A1557). Right next to it are link states.

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
```diff
@@ -315,8 +321,10 @@ Source: https://github.com/WordPress/gutenberg/blob/3b2eccc289cfc90bd99252b12fc4
 		}
 
 		/* Submenu indentation when there's no background. */
-		left: -1px; /* Border width. */
-		top: 100%;
+
+		/* Changed from source — left has no border, top needs 4px for current item border. */
+		left: 0;
+		top: calc(100% + 4px);
```
StevenDufresne: Should the border-width be parameterized?

mu-plugins/blocks/global-header-footer/src/header/index.js
```diff
@@ -0,0 +1,40 @@
...
+import { registerBlockType } from '@wordpress/blocks';
+import ServerSideRender from '@wordpress/server-side-render';
+import { useBlockProps } from '@wordpress/block-editor';
+
+/**
+ * Internal dependencies
+ */
+import metadata from './block.json';
+
+const variations = [
```
StevenDufresne, line 15: It would be nice if these came from the `block.json` attributes. Just noting that, I recognize how this is limited by the translations.

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -9,16 +9,32 @@
...
  * Defined in `render_global_header()`.
  *
+ * @var array  $attributes
  * @var array  $menu_items
  * @var string $locale_title
  * @var string $show_search
  */
 
-$container_class = 'global-header';
+$container_class = 'global-header has-text-color has-background';
```
StevenDufresne, line 18: Is it worth denoting which styles are Gutenberg for future developers? I think `has-text-color` and `has-background` are correct?

StevenDufresne:
I added a few more minor comments. I _think_ this is working properly, I wasn't able to get the header and footer to work on a blank template, but it's most likely how I am trying to use the site editor within a different block-based theme. As long as this gets some proper sandbox testing before it's deployed, I think this is good.

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
ryelle: Oh, It already is, I can use `--active-menu-item-border-height` - good catch.

mu-plugins/blocks/global-header-footer/src/header/index.js
ryelle: There's also the `isActive` callback, so the variations need to be set in JS, not JSON. We could move the variation definitions to a separate shared folder later, but I didn't think the added overhead was worth it yet.

mu-plugins/blocks/global-header-footer/header.php
ryelle: I'm not sure what you mean - I'm mirroring the way colors are applied to group blocks with `has-X-color` and `has-X-background-color`, so I've added these two for consistency. The only thing I didn't add was the `backgroundColor` & `textColor` attributes to the group comment, since I don't think they're necessary.

ryelle:
I've tested this in my sandbox & made a few fixes, and it seems to be working now. I'm going to deploy it so that we can set it up on the `main-test` site for better testing.

mu-plugins/blocks/global-header-footer/header.php
StevenDufresne: I don't see these styles in any associated CSS styles so my assumption is that these CSS classes are coming from Gutenberg. We may be assuming that future developers are familiar with Gutenberg styling conventions (and business logic).

PR #226, opened 2022-07-25
Add linters gitHub action.

StevenDufresne:


ryelle:
I've tweaked the workflow a bit for this repo, but it's still not working. I don't know why the linter can't read the config file, since it seems to exist where it's looking, and I _think_ it should have the right permissions.

StevenDufresne:
Ah Ok, Thanks. This was supposed to be a draft... I got distracted. :) I'll take a look at prettier, I feel we ran into this before.

StevenDufresne:
For some reason, `prettier`, a peer dependency, was not getting installed consistently. I was able to reproduce locally but I think using `yarn` is a bit more reliable.

So the linting is running now. 

**Question**:
The linter is complaining about missing dependencies in a specific block. The block doesn't have a `package.json` since the build is triggered from the root. I imagine we will just add those dependencies to the root `package.json` as `devDependencies`?  It could get a bit unwieldy since our block list will grow and may include dependencies unique to a specific block but my gut is telling me that we are probably ok with that for now. Thoughts?

StevenDufresne:
I wonder if we should skip php linting for now:

https://github.com/WordPress/wporg-mu-plugins/runs/7532391324?check_suite_focus=true

StevenDufresne:
> **Question**: The linter is complaining about missing dependencies in a specific block. The block doesn't have a `package.json` since the build is triggered from the root. I imagine we will just add those dependencies to the root `package.json` as `devDependencies`? It could get a bit unwieldy since our block list will grow and may include dependencies unique to a specific block but my gut is telling me that we are probably ok with that for now. Thoughts?

Nvm, I see that the template adds these config files. I'll update this block to match that.

ryelle:
> This was supposed to be a draft.

Ah, sorry about that. I saw the PR & had set up the others before, so I just took over :|

> I was able to reproduce locally but I think using yarn is a bit more reliable.

Okay, if you switch the project to using yarn, could you also remove the `package-lock.json`? 

> I wonder if we should skip php linting for now

Yeah, after I got that to run locally I also saw all the issues. I think we should fix them, but php linting can happen in a separate PR.

ryelle:
> Nvm, I see that the template adds these config files.

Hm, the blocks shouldn't have their own dependencies/`package.json`s, there should only be the root `package.json`. That sounds like something to fix in the template… actually most of the config in `mu-block/plugin` shouldn't be there 🙃  I guess seeing it in action is different.

(Since this wasn't designed to have yarn workspaces, I was aiming for one single package.json & list of deps - it's not like these blocks will be used/built outside of the mu-plugins context)

StevenDufresne:
Yeah, that makes sense. 👍 .

.github/workflows/linters.yml
```diff
@@ -0,0 +1,32 @@
...
+                  node-version-file: '.nvmrc'
+                  cache: yarn
+
+            - name: Install dependencies & setup configs
+              run: |
+                  yarn setup:tools
+
+            - name: Lint JavaScript and Styles
+              run: |
+                  yarn lint:js
```
ryelle, line 32:
```suggestion
                  yarn lint:css
                  yarn lint:js
```

I deleted the CSS task at some point during my debugging

mu-plugins/blocks/screenshot-preview/src/screenshot.js
```diff
@@ -69,6 +69,7 @@ function ScreenShotImg( { src, isReady = false } ) {
 	const convertResponseToBase64 = async ( res ) => {
 		const blob = await res.blob();
 
+		/* eslint-disable no-undef */
```
ryelle: Is `FileReader` a global? If you add it to the top of the file, ex `/* global FileReader */`, eslint should understand that. Disabling `no-undef` like this also disables it for the rest of the file.

ryelle (approved):
Two comments, but good to merge once those are resolved.

mu-plugins/blocks/screenshot-preview/src/screenshot.js
StevenDufresne: Ah, good catch. I wanted `// eslint-disable-next-line`. 

Adding it as a global to the top of the file works too.

PR #228, opened 2022-07-26
Create a block for language suggestions.

### LINKED ISSUE #11: Block: Language detection banner
If your browser indicates that you use languages other than English, you get a little banner telling you that WordPress can be downloaded in those languages too.

Screen Shot 2022-07-20 at 5 15 43 PM

And in the mockup

Screen Shot 2022-07-20 at 5 16 34 PM

This could probably continue to render an empty div with `id=lang-guess-wrap`, which would keep working with the code in `mu-plugins/main-network/lang-guess.php`, or it could become a block in `wporg-mu-plugins` to eventually be used on wordpress.org/support too.

StevenDufresne:
Fixes https://github.com/WordPress/wporg-main-2022/issues/11

This PR adds a block that suggests a "guessed" locale. This feature already exists, this ports it to a block for future use.

## Questions
- I've hardcoded default styles, should I be using parent theme CSS variables instead?

## Screenshot
Screen Shot 2022-07-29 at 12 54 22 PM

## Roll out
I think we can use this block and leave the old lang code working but exclude the home & downloaded pages for now.


mu-plugins/blocks/language-suggest/src/block.json
```diff
@@ -0,0 +1,24 @@
...
+	"title": "Language Suggest",
+	"category": "widgets",
+	"icon": "admin-site-alt3",
+	"description": "A block for use across the whole wp.org network.",
+	"textdomain": "language-suggest",
+	"editorScript": "file:./index.js",
+	"viewScript": "file:./front.js",
+	"style": "file:./style.css",
+	"supports": {
+		"align": true,
```
StevenDufresne, line 15: @ryelle Do you know why this isn't loading the full width alignment in `wporg-main-2022`? Based on [this documentation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#align), it should work?

Screen Shot 2022-07-29 at 11 38 12 AM

ryelle: Huh, I got it to be full-width when testing on `main-test` with my sandbox:

Screen Shot 2022-08-01 at 5 34 23 PM

Maybe you've got it in a container that doesn't support wide/full alignment?

mu-plugins/blocks/language-suggest/src/front.js
```diff
@@ -0,0 +1,23 @@
+const init = () => {
+	const container = document.querySelector( '.wp-block-wporg-language-suggest' );
```
ryelle, line 2: Do you think there might be a case where more than one of these will be on a page?

mu-plugins/blocks/language-suggest/src/edit.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * The edit function describes the structure of your block in the context of the
+ * editor. This represents what the editor will render when the block is used.
+ *
+ * @return {WPElement} Element to render.
+ */
+export default function Edit() {
+	return (
+		<p { ...useBlockProps() }>
+			This is a suggestion, which usually <a href="http://wordpress.org">links</a> to something.
```
ryelle: This message might be confusing to someone skimming the page in the editor. How about…
```suggestion
			WordPress is also available in <a href="http://wordpress.org">language</a> (if applicable).
```

mu-plugins/blocks/language-suggest/postcss/style.pcss
```diff
@@ -0,0 +1,18 @@
...
+.wp-block-wporg-language-suggest {
+	background-color: #eff2ff;
+	color: #1e1e1e;
+	font-size: 12px;
+	text-align: center;
+}
+
+/* The api returns a div element but that isn't guaranteed. */
+.wp-block-wporg-language-suggest > * {
+	padding: 2px 4px;
```
ryelle: There's a little more padding in the mockup, and a little less line-height

```suggestion
	line-height: 1;
	text-align: center;
}

/* The api returns a div element but that isn't guaranteed. */
.wp-block-wporg-language-suggest > * {
	padding: 10px 4px;
```

mu-plugins/blocks/language-suggest/language-suggest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * Block Name: Language Suggest
+ * Description: A block for use across the whole wp.org network.
+ *
+ * @package wporg
+ */
+
+namespace WordPressdotorg\MU_Plugins\wporg;
```
ryelle: This was maybe a mistake from the generator script, but this should probably be `Language_Suggest`

```suggestion
namespace WordPressdotorg\MU_Plugins\Language_Suggest;
```

mu-plugins/blocks/language-suggest/src/block.json
```diff
@@ -0,0 +1,24 @@
...
+		"color": {
+			"background": true,
+			"text": true
+		},
```
ryelle: Can you add the link color too? Or remove this support for now, otherwise it's possible to create an unreadable link:

Screen Shot 2022-08-01 at 6 33 27 PM

ryelle (commented):
> I've hardcoded default styles, should I be using parent theme CSS variables instead?

Sure - this block probably won't be used anywhere the custom properties aren't defined, so that would be a good way to stay consistent with the design. You could add fallbacks just in case.

mu-plugins/blocks/language-suggest/language-suggest.php
StevenDufresne: Yeah, I think it is. I got passed me in another review as well :)

mu-plugins/blocks/language-suggest/src/front.js
StevenDufresne: I don't think so.

mu-plugins/blocks/language-suggest/src/block.json
StevenDufresne: Added.

mu-plugins/blocks/language-suggest/src/front.js
```diff
@@ -0,0 +1,23 @@
...
+
+	if ( ! container ) {
+		return;
+	}
+
+	// Construct the URL without protocol
+	const path = window.location.host + window.location.pathname;
+
+	fetch( `/lang-guess/lang-guess-ajax.php?uri=${ encodeURIComponent( path ) }` )
```
StevenDufresne, line 11: This was repurposed from the original code but it would probably be better if it wasn't relative.

mu-plugins/blocks/language-suggest/src/block.json
```diff
@@ -0,0 +1,25 @@
...
+	"supports": {
+		"align": true,
+		"color": {
+			"background": true,
+			"text": true,
+			"link": true
+		},
+		"typography": {
+			"fontSize": true
+		}
```
ryelle: If we don't expect to have more than one, and the JS in `front.js` doesn't work with more than one, we can enforce that with a block supports setting

```suggestion
		},
		multiple: false
```

ryelle (approved):
Looks good 👍🏻 

Feel free to merge with or without the multiple suggestion.

PR #230, opened 2022-07-29
Remove `theme.json` dependency

ryelle:
In https://github.com/WordPress/wporg-parent-2021/issues/17, the colors in the parent theme's `theme.json` will be updated. That means some color custom properties will not be defined on sites using the parent theme. If we switch the custom properties in the header-footer CSS to those used in the new theme.json color set, then they won't be defined on News.

Rather than try to jump through hoops to load the "right" custom properties, this PR just defines them locally. While this does technically mean some duplicate colors, it also means the remote loading of global styles isn't needed anymore.

These custom properties are scoped to `.global-header` & `.global-footer` so that they don't override any theme-specific values.

**To test**

1. Spin this branch up on a sandbox
2. Visit block theme sites & classic sites, there should be no visual changes in the header or footer
3. Spin up the parent theme or a site that uses it (make sure it's the latest), and use this branch of mu-plugins
4. The header & footer should be visible


StevenDufresne:
I pulled this into my sandbox... looks good. 

I think @iandunn would understand the implications of this PR better than myself though.

ryelle:
We can probably also remove the `unregister_classic_global_styles` workaround, since we don't need to worry about the custom properties being overwritten now.

mu-plugins/blocks/global-header-footer/postcss/_common.pcss
```diff
@@ -31,6 +31,30 @@ html {
 	}
 }
 
+/**
+ * Set up custom properties.
```
iandunn, line 35: I think it'd be good to document that these are duplicates (both here and at the original source), so that folks remember to update them in both places when they change in the future.

iandunn (approved):

PR #231, opened 2022-08-01
WIP: Block to pull news posts

StevenDufresne:
See https://github.com/WordPress/wporg-main-2022/issues/31.

This PR adds a block: `wporg/multisite-latest-posts`. 

The block makes a server-side (dynamic block) request via `wp_remote_get` to retrieve posts from a WordPress REST API. 

Since we plan to use this within our network, it was suggested by @dd32 that the block first try switching blogs. This is definitely a good idea but I think we can consider that behavior a next iteration (Feel free to disagree!).

This block is also fetching categories in a separate HTTP request because the news blog's API is not embedding the `podcast` categories properly. If fixes are merged, we should update this code to use `?_embed=true`.  

# Note-Worthy
- This block tries to inherit CSS variables that exist in the wporg-parent-2021 [ theme.json](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/theme.json).
- Doesn't expose styling controls on `html` elements. Ie: Can't change font size via editor.
- The current URL defaults to the news API endpoint :)
- The designs have the post titles and the content below by `15px` but we don't have that preset so I used `10px`.

# Default Frontend View 
Screen Shot 2022-08-02 at 3 34 51 PM

# Features
- User can input a WordPress REST API for requests.
- User can change the number of items displayed.

## Editor
Screen Shot 2022-08-02 at 12 38 33 PM

## Parent Theme Properties
We will need to add these to the `custom` section of the theme.json for the parent theme.

```
"multisite-latests-post": {
	"link": {
		"color": "var(--wp--preset--color--charcoal-1)",
		"spacing": "var(--wp--preset--spacing--10)",
		"details": {
			"fontSize": "var(--wp--preset--font-size--small)"
		}
	},
	"spacing":  "var(--wp--preset--spacing--40)",
	"title": {
		"fontFamily": "var(--wp--preset--font-family--eb-garamond)",
		"fontSize": "var(--wp--preset--font-size--heading-5)",
		"lineHeight": "var(--wp--custom--heading--level-3--typography--line-height)"
	}
}
```

# Testing

1. Add this block to your environment
2. Add block into content
3. Expect to see News posts
4. Change URL to `wordpress.org/showcase/wp-json/wp/v2`
5. Change "Items to Show", expect to see component update in editor (and on page)


mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
```diff
@@ -0,0 +1,165 @@
...
+namespace WordPressdotorg\MU_Plugins\wporg;
+
+/**
+ * Helper function to find item in array by id.
+ *
+ * @param WP_Term[] $arr List of categories
+ * @param number    $id Category ID
+ * @return WP_Term|string
+ */
+function get_by_id( $arr, $id ) {
```
StevenDufresne, line 26: I feel there's a better PHP way of doing this :).

dd32: https://developer.wordpress.org/reference/functions/wp_list_filter/ maybe

mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
```diff
@@ -0,0 +1,165 @@
+<?php
+/**
+ * Plugin Name: Multisite Latest Posts
+ * Description: A block for use across the whole wp.org network.
+ *
+ * @package wporg
+ */
+
+namespace WordPressdotorg\MU_Plugins\wporg;
```
dd32:
```suggestion
namespace WordPressdotorg\MU_Plugins\Multisite_Latest_Posts;
```

mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Returns a list of posts.
+ *
+ * @param string  $endpoint URL
+ * @param string  $post_type WP_Post_Type label
+ * @param integer $limit Numbers of posts to return
+ * @return WP_Post[]|WP_Error
+ */
+function get_posts_via_api( $endpoint, $post_type = 'posts', $limit = 10 ) {
```
dd32: I note that none of these api methods have any form of caching - less important for `switch_to_blog()` but highly important for HTTP calls.

I'd probably suggest that `get_posts_via_api()` use a transient to cache the data.
`
$data = get_transient( __NAMESPACE__ );
if ( $data ) {
   return $data;
}
`
and
`set_transient( __NAMESPACE__, $data, HOUR_IN_SECONDS );` before returning at the end.

Ideally, `get_category()` would also be called in here prior to the data being cached, so it's just a singular transient that needs clearing.

StevenDufresne: I always forget the cache. Thanks. 

It would be nice if we had middleware to do the caching.

mu-plugins/blocks/multisite-latest-posts/postcss/style.pcss
```diff
@@ -0,0 +1,45 @@
+
+body {
+	--wporg-link-title-color: var(--wp--preset--color--charcoal-1);
+	--wporg-link-title-font-family: var(--wp--preset--font-family--eb-garamond);
+	--wporg-link-title-font-size: var(--wp--preset--font-size--heading-5);
+	--wporg-link-title-font-size-line-height: var(--wp--custom--heading--level-3--typography--line-height);
+	--wporg-link-spacing: var(--wp--preset--spacing--20, 20px);
+	--wporg-post-spacing: var(--wp--preset--spacing--30, 30px);
+	--wporg-link-details-font-size: var(--wp--preset--font-size--small);
+}
```
StevenDufresne: I'm torn on assigning this CSS up here... but it feels weird having those uber-specific variables littered in the code base. 

Doing it this way also seems like it could pollute the CSS variable space too though...

mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
StevenDufresne: That function probably does more than necessary. I think I'll just stick to this helper.

dd32: FWIW I agree. although `return wp_list_filter( $arr, [ 'id' => $id ] )[0] ?? '';` doesn't have _that much_ overhead, but at the same time, doesn't make it any more readable.

StevenDufresne: Yeah, just needing to create a [wp_list_util](https://developer.wordpress.org/reference/classes/wp_list_util/) object just to filter an array seems a bit much.

StevenDufresne: Might as well just use `array_filter`...?

mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
StevenDufresne: Do you think we'll land a fix for the news podcast embed issue? If so I'll leave out category caching for now because I will remove it.

dd32: > Do you think we'll land a fix for the news podcast embed issue?

Tomorrow hopefully

StevenDufresne:
Noting: After running some speed tests, I am not satisfied with how long this block is taking and delaying page rendering....

I'm going to try blog switching, if that doesn't speed it up significantly, maybe fork this and do it all on the front.

mu-plugins/blocks/multisite-latest-posts/postcss/style.pcss
ryelle: Maybe style it roughly so it doesn't look bad on its own, and add the theme-specific styles into the parent theme?

mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
dd32: @StevenDufresne FYI; The podcast category embed issue should now be resolved, in a way that won't require reverts again in future..

StevenDufresne:
Thanks for all the help here although I will be closing this and going with #236.

PR #235, opened 2022-08-03
Remove underline from global header and footer navigation links in block themes

### LINKED ISSUE #232: Header & footer links have underlines on block themes
Seen on https://wordpress.org/main-test/

Screen Shot 2022-08-01 at 4 51 25 PM

Either add CSS to remove this (should only appear on hover), or try the text decoration attribute on the nav menu block.

Note - not seen on News because the news theme uses [the higher-specificity selector](https://github.com/WordPress/wporg-news-2021/blob/031d9db053102bd34c9be0987768d0642b8eafb2/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss#L2) (matching GB) to remove it.

adamwoodnz:
See #232

Using `theme.json` to set `text-decoration` wasn't possible (see details in #232) so this PR provides a CSS based solution, matching [what the news theme does](https://github.com/WordPress/wporg-news-2021/blob/trunk/source/wp-content/themes/wporg-news-2021/sass/base/_elements.scss#L2).

Kapture 2022-08-03 at 16 10 17

Testing:
1. Build the CSS and replace the corresponding stylesheet in a locally running instance of the [wporg-main-2022 repo](https://github.com/WordPress/wporg-main-2022)
2. Ensure the navigation links in the header and footer have no underline unless hovered or focused

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -132,6 +132,7 @@ html {
 		&:hover,
 		&:focus {
 			text-decoration-line: underline;
+			color: inherit;
 		}
 	}
 }
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -123,6 +123,17 @@ html {
 	& * {
 		box-sizing: border-box;
 	}
+
+	& .wp-block-navigation-item__content {
+		text-decoration: none;
+		text-decoration-thickness: 1px;
+		text-underline-offset: 0.15em;
+
+		&:hover,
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/global-header-footer/postcss/_common.pcss
```diff
@@ -123,6 +123,17 @@ html {
 	& * {
 		box-sizing: border-box;
 	}
+
+	& .wp-block-navigation-item__content {
```
ryelle: Could this just be `a`? the double class parent selector should be enough specificity, and using just `a` would work for any link added in the future.

ryelle (commented):
🤦🏻  I think the GB issue was fixed in 13.8 — but I'm happy to add this here anyway to hopefully prevent any future issues.

PR #236, opened 2022-08-03
Redesign: Add news post block

### LINKED ISSUE #31: Block: Create a News block to show latest posts
The homepage has a section with 3 latest posts from News. We can't control the RSS feed block to get this exact content, so let's spin up a quick block here.

It should show a configurable number of posts (1-10, default to 3), with title, category, and date. In the future, we might want to configure specific categories to pull from, for now we can just pull the latest.

The "View All" link (in the mockup) will be a link in a paragraph below the block, not part of the block.

List

StevenDufresne:
Fixes https://github.com/WordPress/wporg-main-2022/issues/31. Alternative to #231.

In making performance adjustments to #231 to help page load times, I started to think that a "dynamic" Gutenberg block was not the best approach. 

I chose that approach initially because it's similar to how the core latest-posts block works and it allows us to serverside render the block in the editor. Since we are leaning towards a richer WYSIWYG building experience, I thought the dynamic block was the obvious winner. With that being said, just running `wp_remote_get` is slow, ok when cached but slow. Improving that meant implementing a `switch_to_blog` preference within the block. It works and is fairly performant.

However, it feels over-engineered. So I spun up a front-end-only block. We lose the serverside rendering but we have much less PHP & JS code and performance is best case in terms of page load since it request in the view.

## Parent Theme Properties
We will need to add CSS variables to the parent theme: https://github.com/WordPress/wporg-parent-2021/pull/29

## Without Parent Styles
Default theme styles applied.

Screen Shot 2022-08-04 at 10 06 00 AM

## With Parent Styles
Screen Shot 2022-08-04 at 10 04 48 AM


mu-plugins/blocks/multisite-latest-posts/src/block.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const Block = ( { endpoint, itemsToShow } ) => {
+	const [ posts, setPosts ] = useState( [] );
+	const [ error, setError ] = useState();
+	const [ loading, setLoading ] = useState( false );
+
+	const getPosts = async () => {
+		setLoading( true );
+		const response = await fetch( `${ endpoint }/posts?per_page=${ itemsToShow }&_embed=true` );
```
ryelle: Would `apiFetch` be useful here?

mu-plugins/blocks/multisite-latest-posts/src/edit.js
```diff
@@ -0,0 +1,48 @@
...
+						onChange={ handleEndpoint }
+					/>
+					<NumberControl
+						label={ __( 'Items To Show', 'wporg' ) }
+						onChange={ handleItems }
+						value={ itemsToShow }
+					/>
+				</PanelBody>
+			</InspectorControls>
+			<div>A list of posts will render on the front end.</div>
```
ryelle: Could you use the `Block` component here too? That would bring back the editor view, which sounds like the only downside of this approach currently.

mu-plugins/blocks/multisite-latest-posts/src/front.js
```diff
@@ -0,0 +1,25 @@
...
+	const blockElements = document.querySelectorAll( '.wporg-multisite-latest-posts-js' );
+
+	if ( ! blockElements ) {
+		return;
+	}
+
+	for ( let i = 0; i < blockElements.length; i++ ) {
+		const blockEl = blockElements[ i ];
+
+		render( createElement( Block, blockEl.dataset ), blockEl );
```
ryelle: Is there a reason you're using `createElement` rather than JSX here?

mu-plugins/blocks/multisite-latest-posts/multisite-latest-posts.php
```diff
@@ -0,0 +1,83 @@
...
+	$block_info = require $deps_path;
+
+	if ( ! is_admin() ) {
+		wp_enqueue_script(
+			'wporg-multisite-latest-post',
+			plugin_dir_url( __FILE__ ) . 'build/front.js',
+			$block_info['dependencies'],
+			$block_info['version'],
+			true
+		);
```
ryelle: Technically this has already been registered by the `block.json` registration, but [since we have a `render_callback` ](https://github.com/WordPress/wordpress-develop/blob/e6f752026650b26152c3b420643c6caade624033/src/wp-includes/class-wp-block.php#L266)it's not enqueued.

If you update the render callback to also accept the rest of the args, you can enqueue the registered script like this:

```php
function render_block( $attributes, $content, $block ) {
	wp_enqueue_script( $block->block_type->view_script );
```

It would be better to do that anyway, since there are more deps for `index` that would get loaded on the frontend for no reason.

mu-plugins/blocks/multisite-latest-posts/src/edit.js
StevenDufresne: Oh weird, you are right. For some reason, I never thought that was possible. 🤯

mu-plugins/blocks/multisite-latest-posts/src/block.js
StevenDufresne: I don't think so. Is there anything it provides that would help?

mu-plugins/blocks/multisite-latest-posts/src/front.js
StevenDufresne: The habit goes back to the [meeting calendar.](https://github.com/WordPress/meeting-calendar/blob/a16b720dc76a7a4cb27b21a00707af4b301ea645/src/frontend/index.js#L22). I don't mind jsx.

mu-plugins/blocks/multisite-latest-posts/src/block.js
ryelle: Maybe browser compat? I'm not totally sure for this use case (it has some nice handling of the WP API, but I'm not sure how well that will translate to requesting a different WP site's API).

StevenDufresne: I'll take a look.

StevenDufresne: It looks like it does have some better error handling that could be useful.  

I couldn't get it to work using:
`apiFetch.use( apiFetch.createRootURLMiddleware( rootURL ) );`

Stepping through the code, it was updating the rootURL but doing so **after** triggering the ajax request, even when I created the middleware outside of the component itself. Not sure what was going on, but I think I'm comfortable not making use of it anyways.

I've added some better error handling which I think is all we need.

ryelle:
I made a few updates — most notably switching the attribute `itemsToShow` to `perPage` for consistency with the API, and building the div on save rather than in PHP, since there's no reason for it to be dynamic. I also noticed that the editor preview didn't update when the attributes changed, so I've fixed that in https://github.com/WordPress/wporg-mu-plugins/pull/236/commits/6ad0df2b78ccc499249ee8abb8c292188d98d5ec.

If all that looks good to you, this works for me 👍🏻

StevenDufresne:
> and building the div on save rather than in PHP, since there's no reason for it to be dynamic.

Ha, I thought about this last night... Awesome 👍 .

PR #239, opened 2022-08-04
ScreenShot Block: Only enqueue for themes.

### LINKED ISSUE #237: Screenshot block assets are loading on pages without it.
The resource loading does not check to make sure the block exists on the page before adding. 

We should probably do it the same way we did it for the meeting calendar:

https://github.com/WordPress/meeting-calendar/blob/a16b720dc76a7a4cb27b21a00707af4b301ea645/plugin.php#L118

StevenDufresne:
Fixes #237.

The screenshot block assets are being loaded on every page. Ideally, we only want to load on pages that contain the block. Currently, this block does not register a block for the block editor seeing that it was built only to work on the themes page where we have a Backbone.js managed single-page web app. As a short-term measure, this PR only loads the assets if we are on the theme's blog. In the future, we can fully implement the block editor view and use a better way to enqueue/dequeue assets.

### DIFF
--- a/mu-plugins/blocks/screenshot-preview/block.php
+++ b/mu-plugins/blocks/screenshot-preview/block.php
@@ -51,7 +51,7 @@ function register_assets() {
 
 	$block_info = require $deps_path;
 
-	if ( ! is_admin() ) {
+	if ( ! is_admin() && get_current_blog_id() === WPORG_THEME_DIRECTORY_BLOGID ) {
 		wp_enqueue_script(
 			'wporg-screenshot-preview',
 			plugin_dir_url( __FILE__ ) . 'build/index.js',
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
:shipit: LGTM

Tested in my sandbox

dd32:
I'm wondering if it would be better to use something like `class_exists( 'Theme_Directory' )`, so that the plugin can work outside of Multisite Environments (which wouldn't define `get_current_blog_id()`)

StevenDufresne:
> I'm wondering if it would be better to use something like `class_exists( 'Theme_Directory' )`, so that the plugin can work outside of Multisite Environments (which wouldn't define `get_current_blog_id()`)

I will absolutely use whatever you suggest here.

StevenDufresne:
That class doesn't appear to exist (returns `false`), I also can't find an example in `wporg-themes` looking for context 🤔

dd32:
> That class doesn't appear to exist 

Ugh I always forget we don't have the Theme Directory in a class or namespace'd. I guess `function_exists( 'wporg_themes_init' )` would be the replacement in that case.

I don't really mind which way it's done.

PR #240, opened 2022-08-04
ScreenShot Block: Remove spinner and prefer text for loading.

StevenDufresne:
The block currently imports 1 component from `@wordpress/components`: `<Spinner />`.

By including that component, we inherit more dependencies that are queued by the `wordpress/components` library. Since that component is not really nedded, this PR removes it and replaces it with text. We save roughly `1.2mbs` by removing it.

# Screnshot
| Before | After |
| --- | --- |
| Screen Shot 2022-08-04 at 3 22 52 PM | Screen Shot 2022-08-04 at 3 16 39 PM|

### DIFF
--- a/mu-plugins/blocks/screenshot-preview/postcss/style.pcss
+++ b/mu-plugins/blocks/screenshot-preview/postcss/style.pcss
@@ -1,6 +1,6 @@
 
 :root {
-	--wp-screenshot-theme-color: #007cba;
+	__wp-screenshot-theme-color: #007cba;
 }
 
 .wporg-screenshot-card {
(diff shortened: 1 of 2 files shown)


iandunn:
IIRC https://github.com/WordPress/gutenberg/issues/17079 made Spinner directly importable.

If not, though, then maybe https://github.com/WordPress/gutenberg/issues/28797 or a new issue should be (re)opened?

In the meantime, an alternative would be to directly load the spinner gif

ryelle (approved):
Works for me, the `components` file no longer loads on the theme page.

mu-plugins/blocks/screenshot-preview/postcss/style.pcss
```diff
@@ -26,9 +26,14 @@
...
-.wporg-screenshot--has-error:hover {
+.wporg-screenshot__has-error,
+.wporg-screenshot__has-error:hover {
 	font-size: 10px;
 	color: #aaa;
 	text-decoration: none;
 }
+
+.wporg-screenshot__loading {
+	color: #aaa;
```
ryelle: This contrast is technically too low, though it's more of a decoration… maybe bump it to `#767676` just to be safe.

PR #244, opened 2022-08-09
Latest Posts: Decode HTML entities in titles.

### LINKED ISSUE #242: Latest post block has html entites in the title
News item titles have encoded characters

Screen Shot 2022-08-05 at 9 58 32 AM

iandunn:
Fixes #242

### DIFF
--- a/mu-plugins/blocks/multisite-latest-posts/src/block.js
+++ b/mu-plugins/blocks/multisite-latest-posts/src/block.js
@@ -5,6 +5,8 @@ import { Fragment, useEffect, useState } from '@wordpress/element';
 import { gmdate, gmdateI18n } from '@wordpress/date';
 import { __ } from '@wordpress/i18n';
 
+/* global DOMParser */
+
 const getCategory = ( embed, id ) => {
 	if ( ! embed[ 'wp:term' ] ) {
 		return;
@@ -53,13 +55,27 @@ const Block = ( { endpoint, perPage } ) => {
 		return <div>{ error }</div>;
 	}
 
+	let domParser;
+	if ( 'function' === typeof DOMParser ) {
+		domParser = new DOMParser();
+	}
+
 	return (
 		<ul className="wporg-multisite-latest-posts">
 			{ posts.map( ( i ) => {
+				let title;
 				const category = getCategory( i._embedded, i.categories[ 0 ] );
+
+				// Decode HTML entities that appear in titles in a way that is safe against XSS.
+				if ( 'object' === typeof domParser ) {
+					title = domParser.parseFromString( i.title.rendered, 'text/html' ).documentElement.textContent;
+				} else {
+					title = i.title.rendered;
+				}
+
 				return (
 					<li key={ i.id }>
-						<a href={ i.link }>{ i.title.rendered }</a>
+						<a href={ i.link }>{ title }</a>
 						<div className="wporg-multisite-latest-posts-details">
 							{ category && (
 								<Fragment>


iandunn:
I went ahead and started this even though #242 is assigned to @tellyworth , due to the deadline and time zone differences. This may still need a fallback for older browsers. I'm also working on an opaque linting warning.

iandunn:
An alternative approach might be to pass an arg to the API that would cause it to return `title.raw`. I haven't checked yet if that'd require authentication or not, though.

ryelle:
> I'm also working on an opaque linting warning.

It's asking you to put in a line break, since the line is longer than 115 characters

```
				const decodedTitle = dom.parseFromString( i.title.rendered, 'text/html' ).documentElement
					.textContent;
```

iandunn:
Ah, thanks, the unprintable characters and lack of search results was throwing me off 👍🏻 

When I just format it w/ newlines it complains with:

> Replace `⏎↹↹↹↹↹i.title.rendered,·'text/html'⏎↹↹↹↹).documentElement` with `·i.title.rendered,·'text/html'·).documentElement⏎↹↹↹↹↹`  prettier/prettier

`--fix` doesn't work, nor does running prettier directly. 

Looks like refactoring to split it into multiple statements will, though.

iandunn:
Looks like `title.raw` needs the `edit` context, so that's probably not as good as decoding it on the client side.

iandunn:
I added some code to gracefully handle situations where `DOMParser` isn't available. That might be a bit overkill, but without it the whole block will be blank on [~3% of browsers](https://caniuse.com/mdn-api_domparser_parsefromstring).

mu-plugins/blocks/multisite-latest-posts/src/block.js
```diff
@@ -53,13 +55,27 @@ const Block = ( { endpoint, perPage } ) => {
...
 
+	let domParser;
+	if ( 'function' === typeof DOMParser ) {
+		domParser = new DOMParser();
+	}
+
 	return (
 		<ul className="wporg-multisite-latest-posts">
 			{ posts.map( ( i ) => {
+				let title;
```
adamwoodnz, line 66: This would remove the need for the `else` below
```suggestion
				let title = i.title.rendered;
```

mu-plugins/blocks/multisite-latest-posts/src/block.js
```diff
@@ -53,13 +55,27 @@ const Block = ( { endpoint, perPage } ) => {
...
 				const category = getCategory( i._embedded, i.categories[ 0 ] );
+
+				// Decode HTML entities that appear in titles in a way that is safe against XSS.
+				if ( 'object' === typeof domParser ) {
+					title = domParser.parseFromString( i.title.rendered, 'text/html' ).documentElement.textContent;
+				} else {
+					title = i.title.rendered;
+				}
```
adamwoodnz, line 74: Another technique mentioned [here](https://www.delftstack.com/howto/javascript/javascript-decode-html-entities/#decode-html-entities-using-vanilla-javascript) works for me, could be a fallback.

```
function decodeEntity( inputStr ) {
	const textarea = document.createElement( 'textarea' );
	textarea.innerHTML = inputStr;

	return textarea.value;
}

title = decodeEntity( i.title.rendered );
```

I don't seem to be able to execute JS using it, maybe safe against XSS too?

adamwoodnz (approved):
👍 Works for me, couple of suggestions inline

mu-plugins/blocks/multisite-latest-posts/src/block.js
tellyworth: That's exactly the technique I was going to try. It seems hacky but comes widely recommended. If we used that one, is DOMParser beneficial?

tellyworth:
Re the API and `title.raw`: I think it's worth an upstream ticket about making raw content more available in a public REST API context. Title is an obvious one, I'm sure there are others.

tellyworth (approved):
WFM. Let's merge as-is, we can consider improvements later.

mu-plugins/blocks/multisite-latest-posts/src/block.js
ryelle: This textarea method is also what's used by [the `html-entities` GB package](https://github.com/WordPress/gutenberg/blob/trunk/packages/html-entities/src/index.js) — and actually, that has no dependencies so we could use it without adding a cascade of more stuff to the page.

mu-plugins/blocks/multisite-latest-posts/src/block.js
iandunn: 👍🏻

mu-plugins/blocks/multisite-latest-posts/src/block.js
iandunn: I didn't test, but that didn't seem safe in this context based on https://stackoverflow.com/a/34064434/450127

IIRC in the past I've used `innerText` rather than `innerHTML`, but it sounds like [that's not entirely safe either](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#usually-safe-methods).

iandunn:
There might be some security/privacy edge cases around making `raw` fields public. e.g.,

`content.raw = 'Hello world, here's some generated data: [sass_shortcode apikey=1234]'`

_( same concept applies to blocks, I just didn't wanna write 3x the markup to show an example )_

That's probably much less common in titles, though. That particular example wouldn't work without custom code running the title through `do_shortcode()`. That would kind of "void your warranty", but it also wouldn't always be an obvious vuln to the dev doing it.

Doesn't hurt to open a ticket for a more detailed discussion, though.

PR #245, opened 2022-08-09
Block: Multisite Latest Posts: Move view script to footer

ryelle:
See https://github.com/WordPress/wporg-main-2022/issues/45

This updates the view script registered by the block code to be displayed in the footer of the page instead of the head.

**To test**

1. View a page with this block
2. View source
3. The script should be at the end of the page (search for `wporg-multisite-latest-posts-view-script-js`)


adamwoodnz (approved):
This is pretty effective; the scripts for dependencies (lodash, moment, wp-date) have all moved from `head` to the end of `body` too 🎉 

Lighthouse still lists them as render blocking resources but I'm confident this will improve performance so I think it's worth merging.

adamwoodnz:
✅ Latest posts still load and match design

PR #247, opened 2022-08-17
Add a way to preload fonts

dd32:
This PR attempts to add the ability to preload fonts when required, for example:
```php
/* wp_style_add_data( 'wporg-global-fonts', 'preload', [ 'Inter', 'Inter italic' ] ); */
global_fonts_preload( [ 'Inter', 'Inter italic' ] );
```

See https://github.com/WordPress/wporg-main-2022/issues/92

The way I see this working, would be that a theme would include the CSS as a dependency of their stylesheet, and then, conditionally specify preload when required.

Downsides of this approach:
 - There are multiple places the font is hard-coded
 - ~Subsequent calls to `wp_style_add_data()` specifying additional fonts for preload will overwrite previous ones specified~
 - It's not automatically detected as to whether the font is used and automatically preloaded, the theme/plugin/etc has to do that for us.
 - ~The font face list is case-sensitive, currently.~

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -336,6 +336,11 @@ function render_global_header( $attributes = array() ) {
 		$show_search  = true;
 	}
 
+	// Preload the menu font.
+	if ( is_callable( 'global_fonts_preload' ) ) {
+		global_fonts_preload( 'Inter' );
+	}
+
 	// The mobile Get WordPress button needs to be in both menus.
 	$menu_items[] = array(
 		'title'   => esc_html_x( 'Get WordPress', 'Menu item title', 'wporg' ),
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -338,7 +338,9 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		global_fonts_preload( 'Inter latin' );
+		if ( ! is_rosetta_site() ) {
+			global_fonts_preload( 'Inter latin' );
+		}
 	}
 
 	// The mobile Get WordPress button needs to be in both menus.
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -338,7 +338,7 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		global_fonts_preload( 'Inter' );
+		global_fonts_preload( 'Inter latin' );
 	}
 
 	// The mobile Get WordPress button needs to be in both menus.
--- a/mu-plugins/global-fonts/helper-functions.php
+++ b/mu-plugins/global-fonts/helper-functions.php
@@ -1,0 +1,11 @@
+<?php
+
+/**
+ * Specify a font to be preloaded.
+ *
+ * @param array|string $font_faces The font(s) to preload.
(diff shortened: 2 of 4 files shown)


ryelle:
> Subsequent calls to wp_style_add_data() specifying additional fonts for preload will overwrite previous ones specified

I think this might be a problem, since we probably want to set Inter to load in the global header block (+ EB Garamond if rosetta) for all of w.org, but then also load the rest of them in the parent theme for the new themes.

Maybe a helper function which can merge the font list and then apply the `wp_style_add_data`?

dd32:
> I think this might be a problem [...] Maybe a helper function which can merge the font list and then apply the wp_style_add_data?

I feared that might be the case :) In that case, a helper function seems like the best option forward - whether it uses style data or a global/static variable. I'll take a look at implementing this today.

dd32:
I've updated this with a global function to call, and tested it with the Global Header (included in this PR) + the homepage via https://github.com/WordPress/wporg-main-2022/pull/103

mu-plugins/global-fonts/helper-functions.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+/**
+ * Specify a font to be preloaded.
+ *
+ * @param array|string $font_faces The font(s) to preload.
+ * @return bool If the font will be preloaded.
+ */
+function global_fonts_preload( $font_faces ) {
```
ryelle, line 9: Should this get a `wporg_` prefix? It threw me for a second because I expected it to either have a prefix or be namespaced.

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -334,6 +334,11 @@ function render_global_header( $attributes = array() ) {
 		$show_search  = true;
 	}
 
+	// Preload the menu font.
+	if ( is_callable( 'global_fonts_preload' ) ) {
```
ryelle, line 340: Will we need to add this `is_callable` every time we use it, or is this just for while making the PR?

ryelle (approved):
Works great, I tested with & without https://github.com/WordPress/wporg-main-2022/pull/103 and it preloaded all the expected fonts.

Left a few questions, but nothing blocking.

mu-plugins/blocks/global-header-footer/blocks.php
dd32: It was mostly for while PR'ing, but also for safety afterwards.

I guess it's safe to assume that everything that depends on the mu-plugin, will have the mu-plugins enabled, and a fatal when it doesn't is okay.

I'll strip out the is_callable's.

mu-plugins/global-fonts/helper-functions.php
dd32: Probably, but, if dropping `is_callable()` might as well just drop these global functions, and force dependencies to call the namespaced function names, which can be `use function` imported for simpler code.

dd32:
This does need updating to support #255, we'd probably have to add a translation string on https://github.com/WordPress/wporg-main-2022/pull/103 to specify which font variants to pre-load.

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -338,7 +338,9 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		global_fonts_preload( 'Inter latin' );
+		if ( ! is_rosetta_site() ) {
```
outdoor2kode, line 341: Currently, the preloaded font is specified as `Inter Latin`, this would make the rosetta sites preload unnecessary font files. For example, https://ru.wordpress.org/ should only load the Cyrillic font subsets, but without `if ( ! is_rosetta_site() )`, it would preload `Latin` additionally. 

#293 was opened to fix it.

PR #250, opened 2022-08-25
Update footer CIP text for rosetta sites.

### LINKED ISSUE #85: Localised Code Is Poetry footer colours wrong
**Describe the bug**
As seen below, the code is poetry footer on test.wordpress.org is the wrong colour.

**To Reproduce**
1. Go to test.wordpress.org
2. Observe unreadable Code is Poetry footer.

**Expected behavior**
A clear and concise readable phrase

**Screenshots**

Screen Shot 2022-08-16 at 1 36 24 pm

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 104

StevenDufresne:
Fixes: https://github.com/WordPress/wporg-main-2022/issues/85

This PR updates the default `code_is_poetry` element so it renders better on backgrounds:

- Sets default color to `rgba(255,255,255,0.25)` since we have more dark-colored backgrounds
- Sets `is-dark` on element if background variation ends in `white`

### DIFF
--- a/mu-plugins/blocks/global-header-footer/footer.php
+++ b/mu-plugins/blocks/global-header-footer/footer.php
@@ -20,7 +20,6 @@ $container_class = 'global-footer ' . get_container_classes( $color_scheme );
 $code_is_poetry_src = str_contains( $container_class, 'has-charcoal-2-color' ) ?
 	plugins_url( '/images/code-is-poetry-for-light-bg.svg', __FILE__ ) :
 	'https://s.w.org/style/images/code-is-poetry-for-dark-bg.svg';
-
 ?>
 
 
@@ -115,7 +114,7 @@ $code_is_poetry_src = str_contains( $container_class, 'has-charcoal-2-color' ) ?
 
 		<?php else : ?>
 			<!-- Use text so it can be translated. -->
-			<span class="global-footer__code_is_poetry">
+			<span class="global-footer__code_is_poetry <?php echo str_ends_with( $color_scheme , 'white' ) ? 'is-dark' : '' ;?>">
 				<?php echo esc_html( get_cip_text() ); ?>
 			</span>
 
--- a/mu-plugins/blocks/global-header-footer/footer.php
+++ b/mu-plugins/blocks/global-header-footer/footer.php
@@ -115,7 +115,7 @@ $code_is_poetry_src = str_contains( $container_class, 'has-charcoal-2-color' ) ?
 
 		<?php else : ?>
 			<!-- Use text so it can be translated. -->
-			<span class="global-footer__code_is_poetry <?php echo str_ends_with( $color_scheme , 'white' ) ? 'is-dark' : '' ;?>">
+			<span class="global-footer__code_is_poetry">
 				<?php echo esc_html( get_cip_text() ); ?>
 			</span>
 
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
```diff
@@ -115,13 +115,17 @@
...
+		color: rgba(255, 255, 255, 0.25);
 		text-transform: uppercase;
 		letter-spacing: 0.1em;
 
 		@media (--tablet) {
 			margin-top: 0;
 			text-align: center;
 		}
+
+		.is-dark {
```
StevenDufresne: I can't remember if we prefer chaining short classes or extending `{component}--is-dark`. I couldn't find any examples and the block editor [coding guidelines](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/) don't appear to be clear.

mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
```diff
@@ -115,13 +115,17 @@
 		margin-bottom: 0;
 		font-family: serif; /* MrsEaves doesn't support non-latin characters well. */
 		font-size: 20px;
-		color: #4f5357;
+		color: rgba(255, 255, 255, 0.25);
```
StevenDufresne, line 118: This opacity value seems to match the images.

mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
```diff
@@ -115,13 +115,17 @@
...
 		text-transform: uppercase;
 		letter-spacing: 0.1em;
 
 		@media (--tablet) {
 			margin-top: 0;
 			text-align: center;
 		}
+
+		.is-dark {
+			color: #8f8f8f;
```
StevenDufresne: This color matches the image better than the original color.

dd32 (approved):
Testing this on test.wordpress.org looks good to me! Tested both white and dark backgrounds, and tested with image on wordpress.org + text.

mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
ryelle: Short for these `is-`/`has-` style classes 👍🏻 

But you don't need to create a new class, you could change the color based on the background color:

```css
.wp-block-group.global-footer {
	&.has-white-background-color span.global-footer__code_is_poetry {
		color: #8f8f8f;
	}
}
```

If you do want to stick with `is-dark`, this needs the `&` prefix, otherwise the styles don't build correctly.

Screen Shot 2022-08-25 at 12 46 36 PM

mu-plugins/blocks/global-header-footer/footer.php
```diff
@@ -115,7 +115,7 @@
 
 		<?php else : ?>
 			<!-- Use text so it can be translated. -->
-			<span class="global-footer__code_is_poetry">
+			<span class="global-footer__code_is_poetry <?php echo str_ends_with( $color_scheme , 'white' ) ? 'is-dark' : '' ;?>">
```
ryelle: Some lint issues with the spacing here:
```suggestion
			<span class="global-footer__code_is_poetry <?php echo str_ends_with( $color_scheme, 'white' ) ? 'is-dark' : ''; ?>">
```

mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
StevenDufresne: > But you don't need to create a new class, you could change the color based on the background color:

Right, I mistakenly thought these were applied by Gutenberg.  👍

mu-plugins/blocks/global-header-footer/footer.php
StevenDufresne: I'll run the linter and open a pr so we can add it to the workflow.

ryelle (approved):
Looks good now 👍🏻

PR #253, opened 2022-08-25
Add new Meta Block to control page description

ryelle:
I saw this comment in https://github.com/WordPress/wporg-main-2022/issues/79

> maybe the solution to this is to have a special block in the post/pattern that specifies the meta description.

and decided to give it a shot. I've got a dynamic block with a description attribute. When rendering on the frontend, it returns an empty string, but sets up a filter on `jetpack_open_graph_tags` to set the description to the one from its attributes. This works for setting the page descriptions, but unfortunately doesn't work as well for the embed excerpt, since the block isn't rendered there.

I tried getting around that by updating the excerpt directly when this attribute is changed, but that will only happen when someone edits the content in the editor, not via the pattern code.

Example of the block in a pattern:

```
<!-- wp:wporg/meta-block {"description":"Download WordPress today, and get started on creating your website with one of the most powerful, popular, and customizable platforms in the world."} /-->
```

When viewed in the editor, it's just a placeholder:
Screen Shot 2022-08-25 at 4 46 20 PM

The description itself is in the sidebar:
Screen Shot 2022-08-25 at 4 46 41 PM

And it does correctly inject the description into the page:
Screen Shot 2022-08-25 at 4 47 51 PM

We could move the text field into the main content of the editor, but I wanted to keep text inputs out of the WYSIWYG section.

This could be expanded if we need to set other meta information per page, like title, etc.

**To test**

- Build the branch
- Insert a Meta Block into your page, or copy the block code above into a template
- View the page source
- The `<meta name="description"` and other description meta should use the string from your block

Try visiting the embed page, `your-url/embed/` — if you set the description on the block in the editor, your description should appear in the embed.

### DIFF
--- a/mu-plugins/blocks/meta-block/index.php
+++ b/mu-plugins/blocks/meta-block/index.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * Block Name: Meta Block
+ * Description: A server-side dynamic block to set meta fields for SEO.
+ *
+ * @package wporg
+ */
+
+namespace WordPressdotorg\MU_Plugins\Meta_Block;
+
+/**
+ * Registers the block using the metadata loaded from the `block.json` file.
+ * Behind the scenes, it registers also all assets so they can be enqueued
+ * through the block editor in the corresponding context.
+ *
+ * @see https://developer.wordpress.org/reference/functions/register_block_type/
+ */
+function meta_block_init() {
+	register_block_type(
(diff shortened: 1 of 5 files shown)


ryelle:
Closing this, it would make more sense to use (and translate) the excerpt, or support the Jetpack SEO settings.

Screenshot 2024-12-16 at 12 43 12 PM

PR #254, opened 2022-08-26
Remove forked Gutenberg nav block styles for sub menus

adamwoodnz:
See #246, some Gutenberg styles were forked to support navigation block sub menus in older browsers which don't support the `:where` selector.

The usage of this has now been [removed in Gutenberg 13.9](https://github.com/WordPress/gutenberg/pull/42967) so the forked styles can be removed.

#### Screenshots

Safari 13.1:

Screen Shot 2022-08-26 at 12 03 31 PM

#### Testing

1. Build the `global-header-footer` styles to produce `mu-plugins/blocks/global-header-footer/build/style.css`
2. Load this stylesheet in a local dev environment, eg. https://github.com/WordPress/wporg-main-2022 or sandbox
3. Test the navigation sub menus across a range of browsers including those that don't support `:where`, eg. Safari 13.1, Falkon 3.1.0. Use mouse, keyboard, VoiceOver. Compare behaviour with wordpress.org


ryelle (approved):
Works in Chrome 87 👍🏻

PR #255, opened 2022-08-29
Reduce font sizes

### LINKED ISSUE #95: Reduce the fonts file size
Ref: https://wordpress.slack.com/archives/C02QB8GMM/p1660640323188989?thread_ts=1658956086.933959&cid=C02QB8GMM

The fonts files seem larger than usual, let's see if we can reduce their sizes.

image

outdoor2kode:
Fixes https://github.com/WordPress/wporg-main-2022/issues/95

The font subsets here are based on [Google Fonts](https://fonts.googleapis.com/css2?family=Inter&display=swap) and the [post](https://cloudfour.com/thinks/font-subsetting-strategies-content-based-vs-alphabetical/#footnote-2). And if we download one of the font files in [Google Fonts](https://fonts.googleapis.com/css2?family=Inter&display=swap), for example, [Latin](https://fonts.gstatic.com/s/inter/v12/UcCO3FwrK3iLTeHuS_fvQtMwCp50KnMw2boKoduKmMEVuLyfAZ9hiJ-Ek-_EeA.woff2), we can find that it's much smaller than ours (17kB vs. 78kB), that's because it only provides one font-weight while ours provide all weights that `Inter` has supported. 

We could also subset the fonts to a particular font weight as well for the home page (which has higher traffic and it's the entrance of the whole w.org) to pursue better performance.

A font-subset script was also committed in this PR, not sure if it's suitable to put it here, let me know if there's a better place for it.

#### Sizes are around one-third of the original ones.

| Before | After |
| ------- | ----- |
| image | image |

#### Italic, different font weights, and Greek/Cyrillic/Latin are still provided.
https://user-images.githubusercontent.com/18050944/187210593-6185e308-4d38-43b1-b1f5-5d21e3777f26.mp4

### How to test the changes in this Pull Request:

1. Build `wporg-mu-plugins` in `wporg-main-2022`.
2. Go to homepage and check if the fonts remain the same.
3. Press Edit Site, add/edit a Heading block, and open devTools to check if the sizes are really smaller than the old ones.
4. Change the font family to Inter/EB Garamond in the editor and see if it still provides _italic_, **different font weights**, and Greek, Latin, and Cyrillic text.


ryelle (approved):
Looks good! I tested on my sandbox with a handful of Rosetta sites and saw all the correct font subsets downloaded 👍🏻 

> A font-subset script was also committed in this PR, not sure if it's suitable to put it here, let me know if there's a better place for it.

Having it here makes sense to me, thanks for adding it.

> We could also subset the fonts to a particular font weight as well for the home page (which has higher traffic and it's the entrance of the whole w.org) to pursue better performance.

I don't think we need to do that - we'd loose the caching benefit on subsequent pages, for one thing, and probably not get _that_ much gain in the grand scheme of things.

outdoor2kode:
> I don't think we need to do that - we'd loose the caching benefit on subsequent pages, for one thing, and probably not get that much gain in the grand scheme of things.

Oh, you're right, cache does bring more benefits if taking everything into consideration, it's probably not worth spending time separating that many font-weight subsets that would be used in a short while on subsequent pages. Thanks!

PR #256, opened 2022-08-30
Header: Simplify the admin bar and style to match global header

### LINKED ISSUE #252: Simplifying the wp-admin bar that appears above the site navigation
There's a discussion happening on the [Make Meta blog](https://make.wordpress.org/meta/2022/08/18/navigation/) as the moment about the WordPress.org site navigation being a bit of an unorganized mess. Some of the issues have to do with the wp-admin bar that appears above the navigation:

- This bar duplicate the logo, showing it twice in close proximity
- The colors and styles feel different from the rest of the components on the site
- It has a totally separate dropdown menu that feels hidden/not integrated well with the main navigation of the site

I have a few suggestions to clean up and better integrate this UI.

### Homepage — logged out

home-logged-out

- It would be great to show the admin bar on the homepage even when logged out
- The admin bar and associated menus could be re-styled to match the typography, color, spacing of the rest of the site.
- For the logged out state, we could hide several items that current appear including the W logo and the search. This would leave two links on the right hand side to Register and Log in.

### Homepage — logged in

home-logged-in

- For the logged in state, we could incorporate the "My Site" links into the existing (re-styled) user profile menu that's accessible from the upper right.
- The "W" logo and search would be hidden in this view (same as the logged out state).

### Subpages

subpage-logged-in

- The logged out state for homepage could be the exact same as the homepage
- When logged in, we might want to show an action/menu for creating new items in the upper left. Any other post management actions can take place by navigating to that site from the user menu.

For future iterations, it would be cool to look at how to more fully integrate the wp-admin bar into the navigation. There are a lot of different things that could be explored, such making a state change that would display the user info more prominently when logged in:

| Possible future integration (logged out) | Possible future integration (logged in) |
| ------------- | ------------- |
| future-logged-out | future-logged-in  |

Would love to hear your thoughts! Thanks @javierarce and @jasmussen for collaborating on this proposal.

Figma: https://www.figma.com/file/SQwX90pGxS0XlNTEyNhDRE/WordPress.org-Navigation-Explore?node-id=2%3A1965

ryelle:
Fixes #252. Simplify the admin bar content and style to match the global header. This also enables the admin bar across all sites - or it will, once we remove [mu-plugins/pub/wporg-no-adminbar-if-no-blogs.php](https://github.com/WordPress/wordpress.org/blob/c54eb93f3f27d02bba495f6a418db2316e9a03c6/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-no-adminbar-if-no-blogs.php).

For logged-in users…

- Remove the W logo & about menu, site name, comments, stats, likes widget, and search.
- Move My Sites dropdown content into the user profile dropdown
- Create a new "Edit" dropdown for "Edit (current object)", "Customize", and "Edit Site" — grouping these to consolidate the menu, but leaving them for fast access for site editors. These only show if relevant to the current site/page/user.
- The "+ New" menu stays as-is, and only shows if you have access to create content
- Shift the "Edit Profile" and "Logout" links to the end of the menu in their own group

For logged-out users…

- Remove all menu items, except Register & Log In
- Re-add Register and Log In if they don't exist

Finally, this styles the admin bar to match the new design library colors (which is why this is bundled as part of `global-header-footer`, so we can easily use those custom properties).

To do in other places:

- [x] Remove [mu-plugins/pub/wporg-no-adminbar-if-no-blogs.php](https://github.com/WordPress/wordpress.org/blob/c54eb93f3f27d02bba495f6a418db2316e9a03c6/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-no-adminbar-if-no-blogs.php), since the admin bar should be always visible
- [x] Remove [mu-plugins/pub/wporg-admin-bar.php](https://github.com/WordPress/wordpress.org/blob/c54eb93f3f27d02bba495f6a418db2316e9a03c6/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-admin-bar.php) as those link customizations are no longer needed
- [x] Update `mu-plugins/admin-bar-sandbox.php` to add the sandbox flag as a menu item (?)
- Anything else?

### Screenshots

|  | Mobile | Desktop |
|--------|--------|-------|
| Logged out |  450-loggedout | 1320-logged-out |
| Average user | 450-avguser | 1320-avguser |
| Average user with Edit | 450-avguser-w-edit | 1020-avguser-w-edit |
| Super admin | 450-superadmin | 1320-superadmin |

Dropdown examples:

|     | Screenshot |
|---|------------|
| + New menu |  |
| Edit menu |  |
| Average user |  |
| Network admin user |  |
| Forums |  |

This applies to the rosetta sites, too (again with my super-admin user, plus rosetta adds a copy-post plugin).
1320-superadmin-ja

### How to test the changes in this Pull Request:

This ideally needs a sandbox to see the full extent of the changes, but you can test it locally.

If you test on a sandbox, you'll want to delete `/mu-plugins/pub/wporg-no-adminbar-if-no-blogs.php` first.

1. View any page
2. You should see the admin bar, whether logged in or not
3. Log in, you should see a simplified admin bar

I've checked on wordpress.org, rosetta, make.wordpress.org, and subsites; and quick checks on translate.w.org. The admin bar is not used on trac, planet, or codex.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/admin-bar.php
+++ b/mu-plugins/blocks/global-header-footer/admin-bar.php
@@ -17,7 +17,6 @@ add_filter( 'show_admin_bar', __NAMESPACE__ . '\should_show_admin_bar' );
  * @return bool
  */
 function should_show_admin_bar( $show_admin_bar ) {
-	// @todo Don't enable on the login page.
 	return true;
 }
 
--- a/mu-plugins/blocks/global-header-footer/admin-bar.php
+++ b/mu-plugins/blocks/global-header-footer/admin-bar.php
@@ -67,6 +67,8 @@ function filter_admin_bar_links( $wp_admin_bar ) {
 				$ab_item->parent = 'edit-actions';
 				$wp_admin_bar->remove_node( $ab_item->id );
 				$edit_items[] = $ab_item;
+			} else if ( preg_match( '/blog-\d+/', $ab_item->parent ) ) {
+				$wp_admin_bar->remove_node( $ab_item->id );
 			}
 		}
 
--- a/mu-plugins/blocks/global-header-footer/admin-bar.php
+++ b/mu-plugins/blocks/global-header-footer/admin-bar.php
@@ -1,0 +1,138 @@
+<?php
+
+namespace WordPressdotorg\MU_Plugins\Global_Header_Footer;
+
+use Rosetta_Sites, WP_Post, WP_REST_Server, WP_Theme_JSON_Resolver;
+
+defined( 'WPINC' ) || die();
+
+/* Actions & filters */
+add_action( 'admin_bar_menu', __NAMESPACE__ . '\filter_admin_bar_links', 500 ); // 500 to run after all items are added to the menu.
+add_filter( 'show_admin_bar', __NAMESPACE__ . '\should_show_admin_bar' );
+
+/**
+ * Always show the admin bar.
+ *
+ * @param bool $show_admin_bar Whether the admin bar should be shown.
+ * @return bool
+ */
+function should_show_admin_bar( $show_admin_bar ) {
+	// @todo Don't enable on the login page.
+	return true;
+}
+
+/**
+ * Filter the admin bar links to simplify the frontend view.
+ *
+ * @param \WP_Admin_Bar $wp_admin_bar The WP_Admin_Bar instance, passed by reference.
+ */
+function filter_admin_bar_links( $wp_admin_bar ) {
+	// Only change the admin bar on the frontend.
+	if ( is_admin() ) {
+		return;
+	}
+
+	if ( is_user_logged_in() ) {
+		// Remove all items in these submenus.
+		$parent_remove_list = [ 'wp-logo-external', 'wp-logo', 'site-name', 'appearance' ];
+
+		// Remove all these items (includes the top-level items from the parent list).
+		$remove_list = array_merge(
+			$parent_remove_list,
+			[ 'comments', 'stats', 'search', 'my-sites', 'admin-bar-likes-widget' ]
+		);
+
+		// Empty array so we can reverse it to keep the correct order.
+		$edit_items = [];
+
+		foreach ( $wp_admin_bar->get_nodes() as $ab_item ) {
+			if ( in_array( $ab_item->parent, $parent_remove_list ) ) {
+				$wp_admin_bar->remove_node( $ab_item->id );
+			} else if ( in_array( $ab_item->id, $remove_list ) ) {
+				$wp_admin_bar->remove_node( $ab_item->id );
+			} else if ( 'my-sites' === $ab_item->parent ) {
+				// Move items in "My Sites" into user's dropdown.
+				$ab_item->parent = 'my-account';
+				if ( empty( $ab_item->meta['class'] ) ) {
+					$ab_item->meta['class'] = 'ab-sub-secondary';
+				}
+				$wp_admin_bar->add_node( $ab_item );
+			} else if ( in_array( $ab_item->id, [ 'edit-profile', 'logout' ] ) ) {
+				// Move "Edit Profile" and "Logout" to a new group at the end of the dropdown.
+				$ab_item->parent = 'my-account-actions';
+				$wp_admin_bar->add_node( $ab_item );
+			} else if ( in_array( $ab_item->id, [ 'edit', 'site-editor', 'customize' ] ) ) {
+				// Move "Edit [object]", "Customize", and "Edit Site" (if exist) to list to be
+				// added to a new dropdown later.
+				$ab_item->parent = 'edit-actions';
+				$wp_admin_bar->remove_node( $ab_item->id );
+				$edit_items[] = $ab_item;
+			}
+		}
+
+		// Reverse the list to preserve order.
+		$edit_items = array_reverse( $edit_items );
+		if ( 1 === count( $edit_items ) ) {
+			// Only one item, let it be the top level item.
+			$edit_items[0]->parent = false;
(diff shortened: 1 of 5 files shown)


mu-plugins/blocks/global-header-footer/admin-bar.php
```diff
@@ -0,0 +1,138 @@
...
+add_filter( 'show_admin_bar', __NAMESPACE__ . '\should_show_admin_bar' );
+
+/**
+ * Always show the admin bar.
+ *
+ * @param bool $show_admin_bar Whether the admin bar should be shown.
+ * @return bool
+ */
+function should_show_admin_bar( $show_admin_bar ) {
+	// @todo Don't enable on the login page.
```
ryelle: Currently the admin bar shows up on login.w.org. I'm not sure how to target the login site to exclude it, but also it seems like [this code should be disabling it](https://github.com/WordPress/wordpress.org/blob/c54eb93f3f27d02bba495f6a418db2316e9a03c6/wordpress.org/public_html/wp-content/themes/pub/wporg-login/functions.php#L47). Suggestions welcome 🙂

mu-plugins/blocks/global-header-footer/admin-bar.php
```diff
@@ -0,0 +1,138 @@
...
+
+	if ( is_user_logged_in() ) {
+		// Remove all items in these submenus.
+		$parent_remove_list = [ 'wp-logo-external', 'wp-logo', 'site-name', 'appearance' ];
+
+		// Remove all these items (includes the top-level items from the parent list).
+		$remove_list = array_merge(
+			$parent_remove_list,
+			[ 'comments', 'stats', 'search', 'my-sites', 'admin-bar-likes-widget' ]
+		);
```
ryelle, line 42: Specifically removing elements like this ensures that items added by plugins (debug, query monitor, copy post, etc) are not removed.

critterverse:
Awesome! Haven't tested but leaving some notes based on the screenshots:

Do we definitely need the arrows + flyout menu for all the "my sites" options? I was imagining that we could remove the flyout menu altogether and just have those links go directly to the site's Dashboard. Also wondering if we need the site icon there — this list would look a lot cleaner without. What do others think?

arrows

Could we also add more spacing in the menus? I think each item could have a cell height of 32px:

menu

Also one small alignment thing, it looks like the display name is inset a little further than the user name:

alignment

ryelle:
> Do we definitely need the arrows + flyout menu for all the "my sites" options?

I opted not to do that since we were already moving things around, I didn't want to remove _more_ things people might use, but linking off to the dashboard should be enough 👍🏻 

> Also wondering if we need the site icon there — this list would look a lot cleaner without. What do others think?

Agreed, especially since it looks like most sites don't customize it 😆 

> Could we also add more spacing in the menus? I think each item could have a cell height of 32px:

😅  This caused some cascading issues on mobile when I tried it, but I can make another pass.

> Also one small alignment thing, it looks like the display name is inset a little further than the user name:

🤦🏻  totally caused by fixing something else.

ryelle:
Okay, I've updated the height of each menu item (and fixed the mobile issue i mentioned), removed the submenus & site icons from the site list, and fixed that name/username alignment issue.

I left the submenu in Network Admin since it's useful to have those shortcuts, and most people won't have that menu.

Screen Shot 2022-08-30 at 5 34 59 PM

dd32:
I've added this to the list:
> [ ] Remove [mu-plugins/pub/wporg-admin-bar.php](https://github.com/WordPress/wordpress.org/blob/c54eb93f3f27d02bba495f6a418db2316e9a03c6/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-admin-bar.php)

Testing this change, it seems good to me.

The only issue I can find is, that if you have a long list of blogs, like me, which causes the menu to be longer than the screen height, it's not possible to scroll it / see the bottom actions
_edit: I just realised the CSS from this PR didn't apply properly, if this was already accounted for, just ignore this...._

This is a slightly-hacky suggestion, but one quick-solution to that without needing to have the entire menu scroll (which seems to be hard to do, and actually results in a worse UX anyway IMHO) is this rule:
```
#wp-admin-bar-my-sites-list {
    max-height: 50vh;
    overflow: scroll;
}
```
which results in this: 
https://user-images.githubusercontent.com/767313/187582543-1bf53874-a13f-431c-a543-29348d2cc362.mov

mu-plugins/blocks/global-header-footer/admin-bar.php
dd32: This is because the filter used here runs on the default, and `show_admin_bar()` as used in the login theme only toggles the default state, the filter can still override it.

I wasn't able to duplicate the admin bar showing on login.wordpress.org, but regardless, In https://github.com/WordPress/wordpress.org/commit/99da786734fe0b0c9c82dee870f54f6bbda4c1b4 I've switched the theme to use the filter, which should 100% work to hide it I believe.

dd32 (approved):

mu-plugins/blocks/global-header-footer/admin-bar.php
ryelle: I don't see it on the login page anymore 👍🏻

joedolson:
It appears that the 'Edit' options are no longer accessible from the keyboard since this change. @alexstine, the Make/Accessibility team rep noted this because it significantly broke his workflow to not be able to reach the 'Edit Post' submenu while posting agendas & meeting notes. 

I haven't explored thoroughly, but I think it broke the event handlers in /js/_enqueus/lib/admin-bar.js

ryelle:
Thanks for the report, can easily reproduce. I'll make a new issue to get that fixed.

PR #263, opened 2022-09-02
Header: Fix visible focus styles

### LINKED ISSUE #249: Header Menu: Improve style of visible focus indicator
When tabbing through the header nav menu, browsers will add a visible focus indicator (outline).

Currently, it's being cut off on the top/right on top-level menu items:

The focus on the dropdown toggle is awkwardly small, which also flags that the click target here is tiny.

There are also two other (different) focus styles on the header, on the search and the "Get WordPress" button.

Screen Shot 2022-08-19 at 3 51 17 PM Screen Shot 2022-08-19 at 3 51 21 PM

The dashed outline seems to be the preferred button hover, so we could use that like so:

```css
…:focus-visible {
    outline: 1px dashed currentColor;
    outline-offset: -2px;
}
```

Screen Shot 2022-08-19 at 3 56 29 PM

### LINKED ISSUE #261: Submenu hover issue on bigger touch devices
Hovering (touching) submenu links on bigger touch devices causes the page to load the parent link, preventing users from choosing a page from the submenu dropdown.

https://user-images.githubusercontent.com/62803839/188175622-50edae82-ee9f-4403-b2bf-bf7eca3a3cc8.MOV

ryelle:
Fixes #249 — This makes the focus state consistent across all header items, using the dashed outline style. The outline is inset slightly to prevent it being cut off or invisible against the background. Additionally, the spacing around the dropdown arrow has been increased so that it can be a larger click/touch target (fixes #261).

I've opted to use `focus-visible` so that the browser heuristics can show this only when a user is keyboard-navigating, not when the item is clicked on (hover/active state works for that case).

**Screenshots**

#249 has some before screenshots. 

|   | Black on White | White on Black |
|---|----------------|----------------|
| Top level focus | menu-top-level | dark-menu-top-level |
| Arrow focus | menu-arrow | dark-menu-arrow |
| Submenu focus | menu-second-level | dark-menu-second-level |
| Get WP | menu-get-wp | dark-menu-get-wp |
| Search button | menu-search | dark-menu-search |

**To test**

- Try on pages with the different header variations, the outline color should always match the text color to ensure contrast
- Navigate through the menu with a keyboard

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
@@ -11,7 +11,6 @@
 		padding: 10px 19px;
 		border-radius: 2px;
 		font-weight: 700;
-		border: 1px dashed transparent;
 
 		&:hover {
 			cursor: pointer;
@@ -24,8 +23,10 @@
 		}
 
 		&:focus,
+		&:focus-visible,
 		&:focus-within {
-			border-color: var(--wp--preset--color--white);
+			outline: 1px dashed currentcolor;
+			outline-offset: -2px;
 		}
 	}
 
@@ -68,8 +69,10 @@
 
 			&:hover,
 			&:focus,
-			&:focus-within {
+			&:focus-visible {
 				background-color: unset;
+				text-decoration: none;
+				outline: none;
 			}
 		}
 	}
--- a/mu-plugins/blocks/global-header-footer/postcss/header/logos.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/logos.pcss
@@ -38,6 +38,11 @@
 			padding-bottom: 16px;
 			padding-top: 16px;
 		}
+
+		&:focus-visible {
+			outline: 1px dashed currentcolor;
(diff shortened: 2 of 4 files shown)


javierarce:
This looks good to me, but I would just ask for one adjustment:

The top menu items have different left and right paddings… could we equalize them and add the spacing on the `li` element instead of the `a`?

Screenshot 2022-09-05 at 11 44 23

ryelle:
@javierarce Can you tell me what size spacing to use?

javierarce:
@ryelle I'm sorry, I was going to say whatever left padding we were using for the links minus 6px (which is the right padding)… but it turns out that that wouldn't work and after thinking about it for a while, I think there's not a good solution and we shouldn't do anything.

- If we reduce the left padding of the links to match the right one, that will look too narrow when hovering the items.
- If we add more padding at the right of the link, we'll increase the distance between the chevron and the text, which is bad.
- If we do the previous thing and reduce the padding of the chevron… that will make them difficult to click.

So unless someone else has a good solution that works for every case, I'd say to ignore everything I mentioned 😅 

And with that said, the new focus styles look good to me 🚀

PR #267, opened 2022-09-08
Try server-side latest posts block.

### LINKED ISSUE #265: Front-Page: Fetch latest news posts serverside, render as Dynamic Block.
**Describe the bug**
By using Gutenberg libraries on the front, users are forced to load a lot of JS libraries. 

We can probably use this closed branch:
https://github.com/WordPress/wporg-mu-plugins/pull/231

StevenDufresne:
Fixes: #265

This PR narrows the scope of `multisite-latest-posts` to use blog switching and not a front-end approach in order to save `kbs` on the front end.

# Performance
**Resources**
Removes `444kb` of JS resources.

**Page Load**
It's a bit tricky to lock this down because of how many factors exist in the sandbox and creating a unique environment to get more specific numbers is probably overkill. It _seems_ to add roughly `500ms - 800ms` of initial server response time.


mu-plugins/blocks/multisite-latest-posts/src/block.json
```diff
@@ -10,11 +10,10 @@
...
 	"editorScript": "file:./index.js",
 	"style": "file:./style.css",
-	"viewScript": "file:./front.js",
 	"attributes": {
-		"endpoint" : {
-			"type": "string",
-			"default": "https://wordpress.org/news/wp-json/wp/v2"
+		"blogId" : {
+			"type": "integer",
+			"default": -1
```
StevenDufresne: I'm not sure if I should set this at all.

ryelle:
I've pushed some changes, the biggest being renaming the block. I know we're already using this as it is in the Main blog, but we'll need to update that template regardless (because it uses a different syntax now). IMO, switching to "Latest News" also makes it clear that this content comes from News<sup>[1]</sup> (to that end, I've also removed the `blogId` attribute).

Originally we wanted to control the feed/endpoint so that we could specify categories (ex, 1 design post, 1 release, 1 security), but the content on News is diverse enough that it's not necessary.

<sup>[1]</sup> The exception will be rosetta sites, which should use the local blogs (ex, [es.w.org/news](https://es.wordpress.org/news/)), but we can figure that out in another PR.

If this is merged, the custom properties set [in the parent theme](https://github.com/WordPress/wporg-parent-2021/blob/23bb133d41504b19c5fbb510766373e56f1e9f6a/source/wp-content/themes/wporg-parent-2021/theme.json#L442-L455) will need to be updated too.

mu-plugins/blocks/latest-news/latest-news.php
```diff
@@ -0,0 +1,123 @@
...
+			)
+		);
+
+		if ( is_wp_error( $posts ) ) {
+			return $posts->get_error_message();
+		}
+
+		if ( empty( $posts ) ) {
+			return __( 'No posts found.', 'wporg' );
+		}
```
ryelle, line 57: I moved these into the `if` here so that we don't cache errors.

StevenDufresne: 👍

StevenDufresne:
Yep, I don't mind reducing its scope/functionality.

I have created the associated PRs:
https://github.com/WordPress/wporg-parent-2021/pull/46
https://github.com/WordPress/wporg-main-2022/pull/122

Were there any other changes you wanted to make?

ryelle:
> Were there any other changes you wanted to make?

Nope, all three PRs look good to merge now.

PR #272, opened 2022-09-15
Fix arrows that are not using correct font

### LINKED ISSUE #266: Arrows are not using the correct font
I think when the font was split for different language subsets (#255), we lost the arrow character(s). The links with inline arrows (for example, the list on /download) now have arrows that don't match the style of the main font.

| Trunk | 0ec221990d20b895b9244b373839b3de607023e5 (correct) |
|-------|----------|
|  |  |

You can also test this out by putting this code anywhere in the patterns

```
<!-- wp:paragraph {"fontSize":"extra-large"} -->
<p class="has-extra-large-font-size">← ↑ → ↓ ↔ ↕ </p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"fontSize":"extra-large"} -->
<p class="has-extra-large-font-size">↖ ↗ ↘ ↙</p>
<!-- /wp:paragraph -->
```

Result — some arrows look correct (up, down, upper left, lower right), others don't. Which is strange, since these should be consecutive code points.

Screen Shot 2022-09-05 at 5 01 45 PM

cc @renintw

outdoor2kode:
Fixes #266 

| Before | After |
| ------ | ----- |
| image image | image image |

#### Root Cause:
Since the way we subset font was based on [Google's Stylesheet](https://fonts.googleapis.com/css2?family=Inter&display=swap), we can find that within the `unicode-range` of `Latin`, it only specifies `U+2191` and `U+2193`.  These two are ↑ and ↓ ([ref](https://jrgraphix.net/r/Unicode/2190-21FF)).

To solve this issue, we can generate a font subset for arrows to include the missing arrows and adjust the stylesheet accordingly. In this PR, `U+2190-2199` is added ([ref](https://jrgraphix.net/r/Unicode/2190-21FF)). We don't include all kinds of arrows because [Inter](https://rsms.me/inter/glyphs/) itself doesn't support all of [them](https://jrgraphix.net/r/Unicode/2190-21FF). So here we just add in the commonly used ones.

#### Discussion:
There are probably other symbols excluded from [Google's Stylesheet](https://fonts.googleapis.com/css2?family=Inter&display=swap) as it's based on what they thought about the glyphs that are mostly used and then made their own customized version sheet. 
1. Do we want to adjust unicode-range every time we spot incorrect fonts? 
2. Just roll back to the original font files?
3. Try to include all the symbols that the fonts we use have provided. (Didn't know if there's an easy way to find out the unicode-range of all symbols that a font has provided if the font supplier doesn't tell it explicitly)


ryelle (approved):
Tested on sandbox, looks good 👍🏻

ryelle:
> There are probably other symbols excluded from [Google's Stylesheet](https://fonts.googleapis.com/css2?family=Inter&display=swap) as it's based on what they thought about the glyphs that are mostly used and then made their own customized version sheet.

I noticed that when I briefly looked into this. The version of Inter we use is known to have more characters than the version from Google (one reason we're using that one), so by subsetting like this we might be loosing that. I wonder if we could just simplify the unicode ranges? Rather than pulling in random characters, break it up into groups of [unicode blocks](https://en.wikipedia.org/wiki/Unicode_block)?  I'm not sure if that's the best way to break this down, but running a few tests with different subsets and rosetta sites on a sandbox should give us a better idea.

(if you haven't seen it, [this presentation](https://www.unicodeconference.org/presentations-42/S5T3-Sheeter.pdf) was interesting, but doesn't actually answer "how does Google fonts pick **these** ranges")

outdoor2kode:
> I wonder if we could just simplify the unicode ranges? Rather than pulling in random characters, break it up into groups of [unicode blocks](https://en.wikipedia.org/wiki/Unicode_block)?

I've roughly checked how [it's](https://fonts.googleapis.com/css2?family=Inter&display=swap) broken into different groups. For `Latin`, `Latin-ext`, `Vietnamese`, `Greek`, `Greek-ext`, we've already included in the stylesheet the corresponding blocks listed in the wiki. 

For example, the range of Latin in our sheet is `U+0000-00FF`, matching the block: [Basic Latin](https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block))[[g]](https://en.wikipedia.org/wiki/Unicode_block#cite_note-Basic_Latin-10) and [Latin-1 Supplement](https://en.wikipedia.org/wiki/Latin-1_Supplement)[[h]](https://en.wikipedia.org/wiki/Unicode_block#cite_note-Latin1_Supplement-11). The inconsistency I found that exists in the sheet happens on `Cyrillic`, its range should be `U+0400..U+04FF`, but in the sheet, it is `U+0400-045F`, a bit different.

And the other small problem of just directly breaking it up into groups of unicode blocks is that, if the font we use doesn't provide certain glyphs, but in the stylesheet, we include the unicode of those certain glyphs into the unicode-range, the browser would download the sheet anyway. 

For example, the [block range](https://jrgraphix.net/r/Unicode/2190-21FF) of arrow is `2190-21FF`, and all the arrows that `Inter` has provided can be found [here](https://rsms.me/inter/glyphs/) (Not sure if the site has provided the groups or categories of its glyphs, I couldn't find one, so I just used dev tools to roughly check whether a glyph exists) If we specify a subset called `Arrow` with unicode range `2190-21FF` in our stylesheet, and when we visit a site that has only one arrow, say `U+21B2`, which is not provided by `Inter`, the browser would still download the whole `Arrow` subset font file and still be unable to render the correct font style.

> (if you haven't seen it, [this presentation](https://www.unicodeconference.org/presentations-42/S5T3-Sheeter.pdf) was interesting, but doesn't actually answer "how does Google fonts pick **these** ranges")

Thanks for sharing this! I haven't seen this, will take some time to read it.

ryelle:
> If we specify a subset called Arrow with unicode range 2190-21FF in our stylesheet, and when we visit a site that has only one arrow, say U+21B2, which is not provided by Inter, the browser would still download the whole Arrow subset font file and still be unable to render the correct font style.

I see. I hadn't thought about that case. Thanks for explaining it!

Issue #275, opened 2022-09-19
The header search bar button is out of place.


StevenDufresne:
The search bar closing button seems broken when viewing https://wordpress.org/about/.

<img width="428" alt="Screen Shot 2022-09-19 at 10 20 01 AM" src="https://user-images.githubusercontent.com/1657336/190936769-875116a0-100d-4ad8-9809-7c9f28732a12.png">

## Environment
MacOs
Chrome: Version 105.0.5195.125 (Official Build) (x86_64)

ryelle:
This looks to be related to the fix I pushed last week for mobile styles: https://github.com/WordPress/wporg-mu-plugins/commit/385f134f7c131d6ffaa6ec3e426079d5f24d70f2

StevenDufresne:
Is it possible that this introduced a different issue? 

I am unable to open the search bar on https://wordpress.org/showcase/.

Affected Browsers:
Safari, Firefox, Chrome

PR #276, opened 2022-09-19
fixed: header search bar close button alignment

hz-tyfoon:
This PR should fix issue: [275](https://github.com/WordPress/wporg-mu-plugins/issues/275)

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
@@ -92,7 +92,7 @@ html {
 
 /* The classic themes don't use the same modal view, so the offset needs to be changed. */
 body.is-classic-theme .wp-block-group.global-header .wp-block-navigation__responsive-container-close {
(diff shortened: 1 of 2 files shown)


hz-tyfoon:
Looks like the 'X' button shouldn't be aligned with the search field rather it should replace the top right 'search icon'. Trying to do that now.

hz-tyfoon:
in [70eb144](https://github.com/WordPress/wporg-mu-plugins/pull/276/commits/70eb14402457d1568fc0e96e3e738ff24aa29c67), moved the css inside media query to make them only work from tablet devices (& not desktop/wider than 890px devices). Because those css was added in the PR [111](https://github.com/WordPress/wporg-mu-plugins/pull/111) to fix issue: [110](https://github.com/WordPress/wporg-mu-plugins/issues/110). that issue was happening in safari on mobile device.

outdoor2kode:
Thanks for helping fix it!
It's fixed in https://github.com/WordPress/wporg-mu-plugins/pull/280.

PR #277, opened 2022-09-19
Solve header search bar button Issue

varshildoshi151189:
Solved #275 

CleanShot 2022-09-19 at 12 39 59

PFA Screenshot after applying patch.


outdoor2kode:
Thanks for helping fix it!
It's fixed in https://github.com/WordPress/wporg-mu-plugins/pull/280.

PR #278, opened 2022-09-19
Solve header search bar button Issue

varshil-doshi:
Solved #275 
image
PFA Screenshot after applying patch.


outdoor2kode:
Fixed in https://github.com/WordPress/wporg-mu-plugins/pull/280

PR #280, opened 2022-09-19
Menu: Fix the search close button on desktop screens

### LINKED ISSUE #275: The header search bar button is out of place.

ryelle:
Fixes #275.

Reverts some of 385f134f7c131d6ffaa6ec3e426079d5f24d70f2, and sets the search button to always be visible.

In 385f134f7c131d6ffaa6ec3e426079d5f24d70f2 I added some code that split classic theme and block theme styles, but that doesn't seem to be needed and is having an unexpected effect on the desktop search dropdown.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
@@ -90,11 +90,6 @@ html {
 	}
 }
 
-/* The classic themes don't use the same modal view, so the offset needs to be changed. */
-body.is-classic-theme .wp-block-group.global-header .wp-block-navigation__responsive-container-close {
-	top: var(--wp-admin--admin-bar--height, 0);
-}
-
 /*
  * Gutenberg Patches
  *
--- a/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
@@ -314,8 +314,7 @@
 }
 
 /* Gutenberg bug: Close button is not visible in Safari; override it locally */
-.global-header__navigation .is-menu-open,
-.global-header__search .is-menu-open {
+.global-header__navigation .is-menu-open {
 	overflow: visible;
 
 	& > div {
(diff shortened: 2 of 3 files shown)


mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
```diff
@@ -314,8 +314,7 @@
 }
 
 /* Gutenberg bug: Close button is not visible in Safari; override it locally */
-.global-header__navigation .is-menu-open,
-.global-header__search .is-menu-open {
+.global-header__navigation .is-menu-open {
 	overflow: visible;
 
 	& > div {
```
ryelle, line 320: I'm not sure if the Safari fix is still necessary here, but the `global-header__navigation .is-menu-open { overflow: visible; }` is, so I've just reverted the change where I added `global-header__search`.

mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
```diff
@@ -90,11 +90,6 @@ html {
 	}
 }
 
-/* The classic themes don't use the same modal view, so the offset needs to be changed. */
```
ryelle, line 93: It's possible this is still necessary after all - I've tested on wp.org/about and wp.org and this works, but when I tried wp.org/support the X was missing again. I'd like to merge & deploy this to fix the desktop issue, though.

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
outdoor2kode: Should we also change the comment here? Since this seems not related only to Safari anymore.
Not a blocker, but I spent some time figuring out what this really means.

Perhaps `Close button is not visible when opening nav menu in the tablet view` or just remove it?

outdoor2kode (approved):
Looks good 👍 Tested on sandbox with Chrome/FF in Desktop/Mobile mode. They all work correctly on my side.

Issue #282, opened 2022-09-21
Profile Bar is not looking proper after login


varshil-doshi:
Hello Team,

When we Login and open the Profile bar, there is a gap coming between 2 sections. For better understanding I have attached the screenshot below:

![CleanShot 2022-09-21 at 16 57 44](https://user-images.githubusercontent.com/94299018/191492605-5d0a5a7d-fbd7-4b82-9c3b-5c632aa4ff95.png)

Thank you.

ryelle:
Fixed in #283

PR #283, opened 2022-09-21
Solved Profile Bar is not looking proper after login

varshil-doshi:
Solved #282 

After applying patch.
CleanShot 2022-09-21 at 18 18 11


ryelle (approved):
Looks good, thanks for the fix!

PR #286, opened 2022-09-23
Fix the padding and font size on the search field

### LINKED ISSUE #285: The search form looks different depending on the page
This is what the search form looks like when you open it from the home, news, or download pages.

Screenshot 2022-09-23 at 15 21 10

If you open it from any other page, it looks like this (without a left padding/margin):

Screenshot 2022-09-23 at 15 21 26

This problem also appears on mobile: 

Screenshot 2022-09-23 at 15 37 17

Tested in Chrome, Safari, and Firefox.

ryelle:
There were some visual differences in the search bar for block-based themes vs classic themes. In the classic themes, the variable used for padding was undefined, so it had no padding. I'm not totally sure why the font sizes were different, but I've set them specifically for desktop and mobile, so new changes shouldn't creep in.

Fixes #285 

| Homepage | News | Patterns |
|------------|-------|----------|
|  |  |  |
|  |  |  |

(PRing since it's late friday, anyone can review/merge, otherwise I'll deploy monday morning)

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -50,6 +50,7 @@ html {
 	--wp--custom--alignment--edge-spacing: clamp(24px, calc(100vw / 18), 80px);
 	--wp--custom--body--typography--line-height: 1.9;
 	--wp--custom--margin--vertical: 30px;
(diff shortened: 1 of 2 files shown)


javierarce:
Thanks for this fix! 

There's a super tiny difference in the classic theme's search form: the left side of the input field is not rounded.

PR #292, opened 2022-10-12
Screenshot-Preview: Add in a css loader.

StevenDufresne:
See: https://meta.trac.wordpress.org/ticket/6059#comment:18

This PR adds a CSS loader to the ScreenShot preview block. Since we are still choosing to keep the theme page light and exclude extra `@wordpress` dependencies, I chose to attempt to recreate the `@wordpress/components/Spinner` component.

When the theme directory undergoes _blockification_, we should lean towards using more `@wordpress` controls, but until then it think this is a fair approach.

## Screenshots
Screen Shot 2022-10-12 at 9 20 25 AM

### Comparing of the two loaders
**Left**: This implementation
**Right**: `@wordpress/components/spinner`
Screen Shot 2022-10-12 at  9 28 51 AM
The circle outline width is not reproducible with CSS. It can either be thicker or thinner than the `<svg/>`.

### DIFF
--- a/mu-plugins/blocks/screenshot-preview/src/screenshot.js
+++ b/mu-plugins/blocks/screenshot-preview/src/screenshot.js
@@ -105,7 +105,11 @@ function ScreenShotImg( { queryString, src, isReady = false } ) {
 	}
 
 	if ( isLoading ) {
-		return <div className="wporg-screenshot wporg-screenshot__loading">{ __( 'Loading …', 'wporg' ) }</div>;
+		return (
+			<div className="wporg-screenshot wporg-screenshot-loader">
+				<span />
+			</div>
+		);
 	}
 
 	if ( hasError || hasAborted ) {
--- a/mu-plugins/blocks/screenshot-preview/src/screenshot.js
+++ b/mu-plugins/blocks/screenshot-preview/src/screenshot.js
@@ -105,11 +105,7 @@ function ScreenShotImg( { queryString, src, isReady = false } ) {
 	}
 
 	if ( isLoading ) {
-		return (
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/screenshot-preview/src/screenshot.js
```diff
@@ -105,7 +105,11 @@ function ScreenShotImg( { queryString, src, isReady = false } ) {
 	}
 
 	if ( isLoading ) {
-		return <div className="wporg-screenshot wporg-screenshot__loading">{ __( 'Loading …', 'wporg' ) }</div>;
+		return (
+			<div className="wporg-screenshot wporg-screenshot-loader">
+				<span />
```
adamwoodnz: Could you use a pseudo element instead of a `span`?

adamwoodnz (commented):
LGTM, just one minor question

mu-plugins/blocks/screenshot-preview/src/screenshot.js
StevenDufresne: I don't see why we couldn't. Let me try.

StevenDufresne: Done: https://github.com/WordPress/wporg-mu-plugins/pull/292/commits/c5b7f9af5dfd243da11dbede2de53375ec2844de

StevenDufresne: Thanks for the suggestion.

adamwoodnz (approved):
😎

PR #294, opened 2022-10-14
Update horizontal slider and image to be more accessible.

StevenDufresne:
This PR updates the `horizontal-slider` & the `screenshot-preview` block based on feedback from the #accessibility wordpress.org slack channel:
https://wordpress.slack.com/archives/C02RP4X03/p1665538949143739

It still doesn't leave us with a perfect `tablist` implementation because there is a lack of content to be used for describing style variations: (https://wordpress.slack.com/archives/C02RP4X03/p1665538927860869)

Additionally, the tabs are dynamically created in `wporg-themes`. Long term, the tabs (thumbnails), should be moved out of the `wporg-themes` javascript and added to this control for better control. While this component was originally supposed to be just a horizontal slider, we probably want to extend it to handle the thumbnails if new theme directory designs work similarly.

**Example of HTML output:**
```
<ul role="tablist">
	<li role="presentation">
		<a class="wporg-screenshot-card" 
			href="https://wp-themes.com/twentytwentytwo" 
			aria-controls="Default" 
			aria-selected="false" 
			role="tab">
			<img src={...} alt="blue style variation">
		</a>
	</li>
</ul>
```
I wanted to change out `<a>` for a `<button>` but too many default styles interfere and based on [Advanced ARIA tip #1: Tabs in web apps](https://www.marcozehe.de/advanced-aria-tip-1-tabs-in-web-apps/), it isn't completely necessary, although probably a more ideal.

### DIFF
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -95,13 +95,9 @@ function Block( { items, title } ) {
 	const height = width * aspectRatio;
 
 	return (
-		<div aria-labelledby="sliderHeading">
+		<div>
 			<div className="horizontal-slider-header">
-				<span>
-					<h3 id="sliderHeading" className="horizontal-slider-title">
-						{ title }
-					</h3>
-				</span>
+				<h3 className="horizontal-slider-title">{ title }</h3>
 				{ ( canNext || canPrevious ) && (
 					<span className="horizontal-slider-controls">
 						<Handle
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -129,7 +129,7 @@ function Block( { items, title } ) {
 							isReady={ true }
 							tagProps={ {
 								'aria-controls': item.title,
-								'aria-selected': 'true',
+								'aria-selected': false,
 								role: 'tab',
 							} }
 						/>
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -95,10 +95,12 @@ function Block( { items, title } ) {
 	const height = width * aspectRatio;
 
 	return (
-		<div>
+		<div aria-labelledby="sliderHeading">
 			<div className="horizontal-slider-header">
 				<span>
-					<h3 className="horizontal-slider-title">{ title }</h3>
+					<h3 id="sliderHeading" className="horizontal-slider-title">
+						{ title }
+					</h3>
 				</span>
 				{ ( canNext || canPrevious ) && (
 					<span className="horizontal-slider-controls">
@@ -115,19 +117,25 @@ function Block( { items, title } ) {
 					</span>
 				) }
 			</div>
-			<div className="horizontal-slider-wrapper" ref={ outerRef }>
+			<ul className="horizontal-slider-wrapper" ref={ outerRef } role="tablist">
 				{ items.map( ( item ) => (
-					<ScreenShot
-						key={ item.title }
-						{ ...item }
-						width={ `${ width }px` }
-						height={ `${ height }px` }
-						aspectRatio={ aspectRatio }
-						queryString={ `?vpw=${ initialWidth * 10 }&vph=${ initialHeight * 10 }` }
-						isReady={ true }
-					/>
+					<li key={ item.title } role="presentation">
+						<ScreenShot
+							{ ...item }
+							width={ `${ width }px` }
+							height={ `${ height }px` }
+							aspectRatio={ aspectRatio }
+							queryString={ `?vpw=${ initialWidth * 10 }&vph=${ initialHeight * 10 }` }
+							isReady={ true }
+							tagProps={ {
+								'aria-controls': item.title,
+								'aria-selected': 'true',
+								role: 'tab',
+							} }
+						/>
+					</li>
 				) ) }
-			</div>
+			</ul>
 		</div>
 	);
 }
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -123,7 +123,7 @@ function Block( { items, title } ) {
 							aspectRatio={ aspectRatio }
 							queryString={ `?vpw=${ initialWidth * 10 }&vph=${ initialHeight * 10 }` }
 							isReady={ true }
-							tagProps={ {
+							anchorTagProps={ {
 								'aria-controls': item.title,
 								'aria-selected': false,
 								role: 'tab',
--- a/mu-plugins/blocks/screenshot-preview/src/block.js
+++ b/mu-plugins/blocks/screenshot-preview/src/block.js
@@ -13,14 +13,14 @@ import ScreenShot from './screenshot';
 /**
  *
  * @param {Object} props
- * @param {string} props.link        Url for anchor tag.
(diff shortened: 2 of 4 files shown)


mu-plugins/blocks/screenshot-preview/src/block.js
```diff
@@ -20,6 +20,7 @@ import ScreenShot from './screenshot';
  * @param {string} props.width       Desired with of the preview element, include unit.
  * @param {number} props.aspectRatio Aspect ratio for the preview element.
  * @param {string} props.queryString Arguments passed to screenshot service.
+ * @param {Object} props.tagProps    Props added to block tag.
```
adamwoodnz: As this object is spread directly onto the tag it would be pretty easy to mistakenly add invalid properties to this. Might be worth being explicit, for example the Typescript type would be HTMLAnchorElement. 

Maybe
```suggestion
 * @param {Object} props.anchorTagProps    HTMLAnchorElement attributes to be added to the block anchor tag.
```

mu-plugins/blocks/horizontal-slider/src/block.js
```diff
@@ -95,10 +95,12 @@ function Block( { items, title } ) {
 	const height = width * aspectRatio;
 
 	return (
-		<div>
+		<div aria-labelledby="sliderHeading">
```
adamwoodnz: In the past I've been told that using the correct heading tags (like the H3 this references) is pretty effective for screen reader users, so I'm wondering if this is necessary... did you find there wasn't enough context without it?

mu-plugins/blocks/horizontal-slider/src/block.js
```diff
@@ -95,10 +95,12 @@ function Block( { items, title } ) {
...
 			<div className="horizontal-slider-header">
 				<span>
```
adamwoodnz: Is this tag necessary? Seem like the h3 can be a block level element, and there aren't any styles targeting the span as far as I can see

Screen Shot 2022-10-17 at 7 45 42 PM

mu-plugins/blocks/horizontal-slider/src/block.js
```diff
@@ -115,19 +117,25 @@ function Block( { items, title } ) {
...
+						<ScreenShot
+							{ ...item }
+							width={ `${ width }px` }
+							height={ `${ height }px` }
+							aspectRatio={ aspectRatio }
+							queryString={ `?vpw=${ initialWidth * 10 }&vph=${ initialHeight * 10 }` }
+							isReady={ true }
+							tagProps={ {
+								'aria-controls': item.title,
+								'aria-selected': 'true',
```
adamwoodnz: When I navigate through the list with Voiceover all the variations are read as selected, which I assume would be very confusing. Are we planning to add a script later which sets the first as selected and swaps the selected state when the user selects another variation?

adamwoodnz (commented):
Good improvement adding the aria roles, but I think we're lacking the full dynamic features of a tab list, as all the variations are currently `aria-selected`. @ryelle just built something similar for the releases page: https://github.com/WordPress/wporg-main-2022/pull/134

Sorry if I'm missing some context and there is more work to come.

mu-plugins/blocks/horizontal-slider/src/block.js
StevenDufresne: To be honest, I was following along with:
https://www.w3.org/WAI/tutorials/carousels/structure/

I'm totally fine with removing it.

mu-plugins/blocks/horizontal-slider/src/block.js
StevenDufresne: This should be set to false. Good catch. 

It'll be selected true on the theme directory side for the short-term.

mu-plugins/blocks/screenshot-preview/src/block.js
StevenDufresne: Yeah, I was hoping in the future we could make the component more flexible and use something like  `as="button"` since not every use case navigates. The reason it's still an `<a>` tag is that theme directory styling makes a mess of buttons and I'm not certain patching all those styles at the moment will be the most effective approach. The theme directory design updates will lend more context to this when we get there.

But good suggestion, I'll make the change since it's a better reflection of how the code currently behaves.

StevenDufresne:
> Good improvement adding the aria roles, but I think we're lacking the full dynamic features of a tab list, as all the variations are currently `aria-selected`. @ryelle just built something similar for the releases page: [WordPress/wporg-main-2022#134](https://github.com/WordPress/wporg-main-2022/pull/134)
> 
> Sorry if I'm missing some context and there is more work to come.

Yeah, there's some theme directory code that does the _rest_ of the functionality that should be temporary.

mu-plugins/blocks/horizontal-slider/src/block.js
```diff
@@ -115,19 +113,25 @@ function Block( { items, title } ) {
...
-					/>
+					<li key={ item.title } role="presentation">
+						<ScreenShot
+							{ ...item }
+							width={ `${ width }px` }
+							height={ `${ height }px` }
+							aspectRatio={ aspectRatio }
+							queryString={ `?vpw=${ initialWidth * 10 }&vph=${ initialHeight * 10 }` }
+							isReady={ true }
+							tagProps={ {
```
ryelle:
```suggestion
							anchorTagProps={ {
```

mu-plugins/blocks/horizontal-slider/src/block.js
```diff
@@ -115,19 +113,25 @@ function Block( { items, title } ) {
...
+								'aria-controls': item.title,
+								'aria-selected': false,
```
ryelle, line 128: Is this ever updated? I don't see it change when I select the variation (on my sandbox).

ryelle (commented):
> I wanted to change out <a> for a <button> but too many default styles interfere and based on [Advanced ARIA tip #1: Tabs in web apps](https://www.marcozehe.de/advanced-aria-tip-1-tabs-in-web-apps/), it isn't completely necessary, although probably a more ideal.

Why do you want to use a button? In this case, you don't need to worry about the "does it navigate" question, since the role is the thing that we're trying to pin down with that question (`link` vs `button`), and you're setting the role to `tab`. So I'd stick with whatever's easiest to style :)

mu-plugins/blocks/horizontal-slider/src/block.js
StevenDufresne: Ah, I merged a commit suggestion that only changed the first ref 🤦 .

StevenDufresne:
I know this PR has a bit of a wild ride :), but is anything left before I merge?

adamwoodnz:
> I know this PR has a bit of a wild ride :), but is anything left before I merge?

Is the theme directory code updating whichever variation is selected to `aria-selected="true"` now?

adamwoodnz:
Last time I tested the `active` class was changing but not aria I believe...

I think Kelly was saying the same https://github.com/WordPress/wporg-mu-plugins/pull/294/commits/327c5e794220f937186077cb8e832d8b930bb260#r998745667

StevenDufresne:
> > I know this PR has a bit of a wild ride :), but is anything left before I merge?
> 
> Is the theme directory code updating whichever variation is selected to `aria-selected="true"` now?

Relevant .org code:
```
   var cards = $( '.wporg-horizontal-slider-js .wporg-screenshot-card' );
   cards.attr( 'aria-selected', false );
   $anchorLink.attr( 'aria-selected', true );
```

adamwoodnz (approved):
LGTM then!

PR #296, opened 2022-10-18
Fix issue with horizontal slider next button.

StevenDufresne:
@adamwoodnz noticed an issue where the next and previous buttons were active but nothing happened when they were clicked. This was happening because the `scrollLeftPos` value was only updated on button clicks (by design) and the new & previous values were the same, never triggering the `useEffect` that scrolls the container.

**So for example:**
- User scrolls to end of the container, by clicking "next" button,`scrollLeftPos = 300`. 
- User would navigate back with a gesture (not clicking the button), still `scrollLeftPos = 300`.
- User would click the button, new position `scrollLeftPos = 300`. 

`useEffect` doesn't get triggered.

The approach here is to just remove the `useEffect` and update the scroll position on every click. I think this is best since the component doesn't try to _remember_ positioning but lets the browser and `scrollLeft` be the source of truth.

This also fixes a secondary issue where the "next" button was being greyed out a few pixels too early.

## Screenshots

| Before | After |
| --- | --- |
| Screen Capture on 2022-10-18 at 14-31-06 | Screen Capture on 2022-10-18 at 14-29-39 |
| Screen Capture on 2022-10-18 at 14-34-06 |  Screen Capture on 2022-10-18 at 14-35-19 |

### DIFF
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -36,7 +36,6 @@ const SET_WIDTH = CARD_WIDTH * 3;
  */
 function Block( { items, title } ) {
 	const outerRef = useRef();
-	const [ scrollLeftPos, setScrollLeftPos ] = useState( 0 );
 	const [ canPrevious, setCanPrevious ] = useState( false );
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
Works for me!

PR #297, opened 2022-10-20
Split out the *-ext subsets

outdoor2kode:
Fixes #295  

On [wp.org](https://wordpress.org/), only necessary font subsets are preloaded. (`Inter-latin.woff2`, `EBGaramond-latin.woff2`, `EBGaramond-Italic-latin.woff2`)
image

PR #298, opened 2022-10-20
Preload required font subsets for rosetta sites

### LINKED ISSUE #293: Preload required font subsets for rosetta sites global header
See https://github.com/WordPress/wporg-mu-plugins/pull/247/commits/449df49d14f77a86ac75c41de671f891317bc93b#r995103206

outdoor2kode:
Fixes #293 

**Note**: This PR should either merge to `fix/prevent-preloading-latin-ext-unless-necessary`, or to `trunk` after that one has been merged to trunk.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -338,8 +338,10 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		if ( ! is_rosetta_site() ) {
-			global_fonts_preload( 'Inter', 'latin' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext.  */
+		$subsets = explode( ',', _x( 'Latin', 'Global menu font subsets, comma separated', 'wporg' ) );
+		foreach ( $subsets as $subset ) {
+			global_fonts_preload( 'Inter', $subset );
 		}
 	}
 
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -339,10 +339,8 @@ function render_global_header( $attributes = array() ) {
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
 		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext.  */
-		$subsets = explode( ',', _x( 'latin', 'Global menu font subsets, comma separated', 'wporg' ) );
-		foreach ( $subsets as $subset ) {
-			global_fonts_preload( 'Inter', $subset );
-		}
+		$subsets = _x( 'latin', 'Global menu font subsets, comma separated', 'wporg' );
+		global_fonts_preload( 'Inter', $subsets );
 	}
 
 	// The mobile Get WordPress button needs to be in both menus.
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -339,7 +339,7 @@ function render_global_header( $attributes = array() ) {
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
 		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext.  */
-		$subsets = explode( ',', _x( 'Latin', 'Global menu font subsets, comma separated', 'wporg' ) );
+		$subsets = explode( ',', _x( 'latin', 'Global menu font subsets, comma separated', 'wporg' ) );
 		foreach ( $subsets as $subset ) {
 			global_fonts_preload( 'Inter', $subset );
 		}


outdoor2kode:
@ryelle 

In this [comment](https://github.com/WordPress/wporg-main-2022/issues/136#issuecomment-1278192526),
> We might want to default to always preloading Latin? Spot-checking a few non-latin Rosetta sites there's always a little latin text.

I've tried the following snippet at the start of [this function](https://github.com/WordPress/wporg-mu-plugins/blob/9d7a2f8467edd6705974b5c07bd6a17edc81d308/mu-plugins/global-fonts/index.php#L82)
```php
// Always preload latin. Since it seems there's always a little latin text on non-latin Rosetta sites.
$preloaded_fonts = [ 'inter latin', 'eb garamond latin', 'eb garamond latin italic' ];
foreach ( $preloaded_fonts as $preloaded_font ) {
$preload[] = [
	'href'        => get_font_url( $preloaded_font ),
	'as'          => 'font',
	'crossorigin' => 'crossorigin',
	'type'        => 'font/woff2',
];
}
```
But I didn't add it to this PR yet because I'm thinking that it kind of does not make sense. 
`Latin` perhaps does exist on every non-latin rosetta site more or less, but not all of them would use every `font family` and `font style` of `Latin`. In regard to their sizes, it seems not a good idea to download them if we're not using them.
image

Perhaps we just let users judge which font family and language to use like mu-plugins/blocks/global-header-footer/blocks.php did?
(Or we can still preload `Inter Latin` anyway as it is now preloaded when `render_global_header`, and I guess basically every site has the header on the top, it doesn't seem to have much difference from always preloading `Latin` - except translators might overwrite it)

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -338,8 +338,10 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		if ( ! is_rosetta_site() ) {
-			global_fonts_preload( 'Inter latin' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+		$subsets = _x( 'latin', 'Global menu font subsets, comma separated', 'wporg' );
+		foreach ( (array) $subsets as $subset ) {
```
ryelle: This `$subsets` value will always be a string, converted to an array of length 1, even if the translator specifies `cyrillic, cyrillic-ext`. You can `explode( ',', $subsets );` to split it into the subset parts. Or maybe `global_fonts_preload` should take a second parameter which is the comma-separated subsets, and `preload_font` can handle splitting apart the subsets?

ryelle:
> not all of them would use every font family and font style of Latin. In regard to their sizes, it seems not a good idea to download them if we're not using them.

Yes, that's true — I was only thinking we'd load Inter Latin when I said that, not EB Garamond. Besides, the benefit we'd get for preloading the file is pretty small since it might only be used for a few words on the page. We can let translators decide if they want latin or not, just like the other subsets.

mu-plugins/blocks/global-header-footer/blocks.php
outdoor2kode: OH! I only tested with:
```
$subsets = 'cyrillic, cyrillic-ext';
foreach ( (subsets) $ as $subset ) {
    echo $subset;
} 
```
And it prints out `cyrillic, cyrillic-ext`. 
Apparently the structure is:
```
Array
(
    [0] => cyrillic, cyrillic-ext
)
```
Thanks for pointing it out!

mu-plugins/global-fonts/index.php
```diff
@@ -69,8 +69,8 @@ function preload_font( $fonts, $subsets ) {
 
 	$preload = $style->extra['preload'] ?? [];
 	$new_preload = [
-		'fonts' => array_unique( $fonts ),
-		'subsets' => array_unique( $subsets ),
+		'fonts' => $fonts,
+		'subsets' => $subsets,
```
outdoor2kode: ~`array_unique` is removed for the cases:
`global_fonts_preload( 'Inter, Inter', 'Latin, Latin-ext' );`
`global_fonts_preload( 'Inter, EB Garamond', 'Latin, Latin' );`~

No longer available after [Update preload_font logic. See README for details.](https://github.com/WordPress/wporg-mu-plugins/pull/297/commits/7851a09f6dde6db8684893ac1cff25aa35c30330)

mu-plugins/global-fonts/index.php
```diff
@@ -89,9 +89,12 @@ function maybe_preload_font( $preload ) {
 		return $preload;
 	}
 
-	$combined_fonts_subsets = array_combine( (array) $style->extra['preload']['fonts'], (array) $style->extra['preload']['subsets'] );
+	$fonts = (array) $style->extra['preload']['fonts'];
+	$subsets = (array) $style->extra['preload']['subsets'];
```
outdoor2kode: ~`array_combine` would fail the case:
`global_fonts_preload( 'Inter, Inter', 'Latin, Latin-ext' );` as the key (i.e., `'Inter'`) can't be duplicated.~

No longer available after [Update preload_font logic. See README for details.](https://github.com/WordPress/wporg-mu-plugins/pull/297/commits/7851a09f6dde6db8684893ac1cff25aa35c30330)

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -338,8 +338,10 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		if ( ! is_rosetta_site() ) {
-			global_fonts_preload( 'Inter', 'latin' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext.  */
+		$subsets = explode( ',', _x( 'Latin', 'Global menu font subsets, comma separated', 'wporg' ) );
+		foreach ( $subsets as $subset ) {
```
outdoor2kode: ~`explode`  + `foreach` are used here for the case that translators might update to more than one subset, for example:
`global_fonts_preload( 'Inter', 'Cyrillic, Cyrillic-ext' );`
If `$subsets` is used directly, it would fail to `get_font_url` for `Cyrillic-ext` as its `$font` is empty. See [this](https://github.com/WordPress/wporg-mu-plugins/pull/298/files#r1005970000) and [this](https://github.com/WordPress/wporg-mu-plugins/pull/298/files#diff-3e55f21b4a7e4e1adfa13b4471290f4cce80b2f75fe9664fa35213977f7fb49cR98). 
Therefore, it can't 
`global_fonts_preload( 'Inter', 'Cyrillic, Cyrillic-ext' );`
it has to get it by using 
`global_fonts_preload( 'Inter', 'Cyrillic');` and `global_fonts_preload( 'Inter', 'Cyrillic-ext');` 
respectively.
Or there's another alternative. 
Check whether `count($subsets)` is greater than `count($font)` in `global_fonts_preload`, and make up for the missing part. But with this, we can't just 
`global_fonts_preload( 'Inter, EB Garamond', 'Latin, Latin' );` 
as it might be modified by translators to something like 
`global_fonts_preload( 'Inter, EB Garamond', 'Cyrillic, Cyrillic-ext, Greek, Greek-ext, Latin, Latin-ext' );`
To deal with the case, it needs to call 
`global_fonts_preload( 'Inter', 'Latin');` and `global_fonts_preload( 'EB Garamond', 'Latin);` 
respectively.~

No longer available after [Update preload_font logic. See README for details.](https://github.com/WordPress/wporg-mu-plugins/pull/297/commits/7851a09f6dde6db8684893ac1cff25aa35c30330)

mu-plugins/global-fonts/index.php
```diff
@@ -69,8 +69,8 @@ function preload_font( $fonts, $subsets ) {
 
 	$preload = $style->extra['preload'] ?? [];
 	$new_preload = [
```
outdoor2kode: Case that translators might make it happen.
e.g., `new_preload` = [ `'fonts'` => [ 'Inter' ], `'subsets'` => [ 'Cyrillic, Cyrillic-ext' ] ]

mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -338,8 +338,10 @@ function render_global_header( $attributes = array() ) {
 
 	// Preload the menu font.
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		if ( ! is_rosetta_site() ) {
-			global_fonts_preload( 'Inter', 'latin' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext.  */
+		$subsets = explode( ',', _x( 'Latin', 'Global menu font subsets, comma separated', 'wporg' ) );
```
ryelle: We should keep this lowercase just for consistency.

```suggestion
		$subsets = explode( ',', _x( 'latin', 'Global menu font subsets, comma separated', 'wporg' ) );
```

mu-plugins/global-fonts/index.php
ryelle: Why do you need to specify "Inter" twice, or "latin" twice?

From the current code, I would expect these outcomes:

| Function call | to preload |
|---------------|------------|
| `global_fonts_preload( 'Inter',  'latin' )` | Inter-latin |
| `global_fonts_preload( 'Inter, EB Garamond',  'latin' )` | Inter-latin and EBGaramond-latin |
| `global_fonts_preload( 'Inter',  'latin, cyrillic' )` | Inter-latin and Inter-cyrillic |
| `global_fonts_preload( 'Inter, EB Garamond', 'latin, latin-ext' )` | Inter-latin, Inter-latin-ext, EBGaramond-latin, and EBGaramond-latin-ext |

Maybe requiring just a single font in the first parameter would help, if the aim is to allow for different subsets per font.

outdoor2kode: Because originally the function works like this:
| Function call | to preload |
|---------------|------------|
| `global_fonts_preload( 'Inter',  'latin' )` | Inter-latin |
| `global_fonts_preload( 'Inter, EB Garamond',  'latin' )` | Would fail, because EB Garamond has nothing to pair |
| `global_fonts_preload( 'Inter',  'latin, cyrillic' )` | Would fail, because Cyrillic has nothing to pair |
| `global_fonts_preload( 'Inter, EB Garamond', 'latin, latin-ext' )` | Inter-latin, EBGaramond-latin-ext |

So if we want to preload `Inter latin` and `Inter latin-ext`,  it needs to be `global_fonts_preload( 'Inter, Inter', 'Latin, Latin-ext' );`

I've changed to your idea, it's smarter.
Although there's a case that it needs to call `global_fonts_preload` twice when we want to preload `latin` and `latin-ext` for `Inter` and only `latin` for `EB Garamond`, such a case should never happen as subsets required for a certain page should be the same.

ryelle (approved):

PR #299, opened 2022-10-31
Tools: Try adding infrastructure for PHPUnit tests

ryelle:
I think it would be helpful to add unit tests for the global-font changes in #297, but we need the infrastructure in place first. This forks out the phpunit code from #204, and rebases it all to the current `trunk`.

To test:

- `yarn` to install the wp-env dep
- `composer update` to install the new composer deps
- `yarn test:php` to run the tests (currently just the demo test)

The actions run on this PR should also include a unit-test action.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,9 @@
 		"phpcompatibility/phpcompatibility-wp": "*",
 		"phpunit/phpunit": "^9.5",
 		"wp-coding-standards/wpcs": "2.*",
-		"wporg/wporg-repo-tools": "dev-trunk"
+		"wp-phpunit/wp-phpunit": "^5.9",
+		"wporg/wporg-repo-tools": "dev-trunk",
+		"yoast/phpunit-polyfills": "^1.0"
 	},
 	"config": {
 		"allow-plugins": {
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -17,14 +17,15 @@
...
-		"composer/installers": "~1.0"
-	},
 	"require-dev": {
+		"composer/installers": "~1.0",
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
 		"phpcompatibility/phpcompatibility-wp": "*",
+		"phpunit/phpunit": "^9.5",
 		"wp-coding-standards/wpcs": "2.*",
-		"wporg/wporg-repo-tools": "dev-trunk"
+		"wp-phpunit/wp-phpunit": "~6.0",
```
ryelle, line 26: I'm pretty sure we don't need `phpunit/phpunit` or `wp-phpunit/wp-phpunit` since it's using the `wp-env` container & wp installation, but I've left it in from #204 in case it's necessary for running phpunit locally.

outdoor2kode (approved):
Tested and looks good 👍

PR #302, opened 2022-11-18
Header: Update spacing around logo, menus, buttons

### LINKED ISSUE #301: Improve the alignment of navigation items
The secondary or local navigation is currently centered on the page, following the style of the content below. 

For some screen resolution using this system produces off alignments between the global navigation, the local navigation and the content, so I'd like to propose a change in alignment that would make the secondary navigation behave like the global one:

image

In the image above the first item in the local navigation (About) is displayed right under the first item in the global navigation (News). In order to achieve this alignment, we'll need to also reduce the distance between the WordPress logo and the left side of the screen.

The proposed design also suggests changes to the mobile navigation.

Here's a [Figma page](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=851%3A2108) where we are exploring these changes.

ryelle:
Fixes https://github.com/WordPress/wporg-mu-plugins/issues/301 — Adjust the spacing in the header to match the latest design iteration.

This PR only changes the header, I'll have another PR for the parent theme to adjust the subpage navigation bar (and then another PR for the child themes that re-implement it… If this changes again, we should figure out a better way of inheriting these styles).

**Screenshots**

|  | WordPress.org | Page with subnav |
|---|------|-------|
| 480px | black-480 | white-480 |
| 600px | black-600 | white-600 |
| 1200px | black-1200-tall | white-es-1200-tall |

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -52,7 +52,7 @@ html {
 	--wp--custom--margin--vertical: 30px;
 	--wp--custom--margin--horizontal: 30px;
 	--wp--custom--form--padding: calc(0.5 * var(--wp--custom--margin--horizontal));
-	--wp--style--block-gap: 24px;
+	--wp--style--block-gap: 20px;
 
 	/* This is a subset of the colors provided by `theme.json`, set here to prevent conflicts across other block themes. */
 	--wp--preset--color--charcoal-1: #1e1e1e;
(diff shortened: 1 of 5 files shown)


StevenDufresne:
The proportions seem to match the originating ticket. 👍

StevenDufresne:
Informed by work on wporg-showcase, I think we need to move to the same approach with the footer and content.

Because we are using fluid widths for the content and the footer, it falls out of align with the header and subnav easily. Is that something that you had considered?

Here is an example:
margin (1)

PR #305, opened 2022-11-28
Blocks: Add the "Site Breadcrumbs" block from showcase

### LINKED ISSUE #304: Move Site Breadcrumbs block to here from Showcase
The breadcrumb nav pattern will be used across multiple sites, so the block should be decoupled from the Showcase theme (https://github.com/WordPress/wporg-showcase-2022/pull/45). For example, it's needed for the Documentation site (https://github.com/WordPress/wporg-documentation-2022/issues/8).

I think the default behavior should be fine, but we can also add a filter for themes to use if they need to override the breadcrumb display in any way.

ryelle:
Fixes #304 — This copies over the Site Breadcrumbs block from the showcase theme (https://github.com/WordPress/wporg-showcase-2022/pull/45) for use on all sites.

Editor view has a simple placeholder, just to get an idea of how the settings will apply (color, font, etc)

On the frontend, it looks the same as the current block in showcase.

The main difference is the addition of a `wporg_block_site_breadcrumbs` filter. This can be used in individual sites' `functions.php` to render slightly different breadcrumbs as needed, since the exact hierarchy changes in each site (for example, showcase only ever has two levels of breadcrumb, but [docs can have 3](https://github.com/WordPress/wporg-documentation-2022/issues/8))

**To test**

- Try this in combination with the showcase PR https://github.com/WordPress/wporg-showcase-2022/pull/65
- Or try using it in a template/page and make sure the customization options work

### DIFF
--- a/mu-plugins/blocks/site-breadcrumbs/index.php
+++ b/mu-plugins/blocks/site-breadcrumbs/index.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Block Name: Site Breadcrumbs
+ * Description: Display breadcrumbs of the site.
+ *
+ * @package wporg
+ */
+
+namespace WordPressdotorg\MU_Plugins\Site_Breadcrumbs;
+
+add_action( 'init', __NAMESPACE__ . '\init' );
+
+/**
+ * Registers the block using the metadata loaded from the `block.json` file.
+ * Behind the scenes, it registers also all assets so they can be enqueued
+ * through the block editor in the corresponding context.
+ *
+ * @see https://developer.wordpress.org/reference/functions/register_block_type/
+ */
+function init() {
+	register_block_type(
+		__DIR__ . '/build',
+		array(
+			'render_callback' => __NAMESPACE__ . '\render_block',
+		)
+	);
+}
+
+/**
+ * Render the block content.
+ *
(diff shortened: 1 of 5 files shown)


mu-plugins/blocks/site-breadcrumbs/index.php
```diff
@@ -0,0 +1,88 @@
...
+ * Render the block content.
+ *
+ * @param array    $attributes Block attributes.
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string Returns the block markup.
+ */
+function render_block( $attributes, $content, $block ) {
+	if ( ! isset( $block->context['postId'] ) ) {
```
StevenDufresne: Can we remove this line? It's causing the control to not show up on the search page.

ryelle: Huh, it's strange that it worked before. In any case, this block uses the global page context, not a local `postId` so that can be removed entirely. Done in https://github.com/WordPress/wporg-mu-plugins/pull/305/commits/730275e166989a137e5283e566bfed5c52bb3b9f.

outdoor2kode (commented):
Tested and looks good 👍 👍 👍

PR #307, opened 2022-12-08
Add dark black header and footer variation

bengreeley:
Currently, the header and footer blocks don't support the charcoal-1 color. This variation is needed for the Enterprise landing page.

To test:
- Open a page in the editor
- Add or edit the 'Global Header' and 'Global Footer' blocks
- Set the variation to 'White on dark black'
- Verify the header/footer blocks change color in the editor.
- Preview and verify the blocks change color on the front end.

Screenshot 2022-12-08 at 12 44 33 PM
Screenshot 2022-12-08 at 12 39 38 PM

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
@@ -308,7 +308,7 @@
 	& .wp-block-navigation__container {
 
 		@media (--tablet) {
-			background: var(--wp-global-header--background-color);
+			background: inherit;
 		}
 	}
 }
--- a/mu-plugins/blocks/global-header-footer/src/footer/index.js
+++ b/mu-plugins/blocks/global-header-footer/src/footer/index.js
@@ -14,7 +14,7 @@ import metadata from './block.json';
 
 const variations = [
 	{ label: __( 'White on charcoal-1', 'wporg' ), value: 'white-on-dark-black' },
-	{ label: __( 'White on charcoal-2 (default)', 'wporg' ), value: 'white-on-black' },		
+	{ label: __( 'White on charcoal-2 (default)', 'wporg' ), value: 'white-on-black' },
 	{ label: __( 'White on blueberry-1', 'wporg' ), value: 'white-on-blue' },
 	{ label: __( 'Black on white', 'wporg' ), value: 'black-on-white' },
 ];
--- a/mu-plugins/blocks/global-header-footer/src/footer/index.js
+++ b/mu-plugins/blocks/global-header-footer/src/footer/index.js
@@ -14,7 +14,7 @@ import metadata from './block.json';
 
 const variations = [
 	{ label: __( 'White on black (default)', 'wporg' ), value: 'white-on-black' },
-	{ label: __( 'White on dark black', 'wporg' ), value: 'white-on-black-dark' },
+	{ label: __( 'White on dark black', 'wporg' ), value: 'white-on-dark-black' },
 	{ label: __( 'Black on white', 'wporg' ), value: 'black-on-white' },
 	{ label: __( 'White on blue', 'wporg' ), value: 'white-on-blue' },
 ];
--- a/mu-plugins/blocks/global-header-footer/src/footer/index.js
+++ b/mu-plugins/blocks/global-header-footer/src/footer/index.js
@@ -14,6 +14,7 @@ import metadata from './block.json';
 
 const variations = [
 	{ label: __( 'White on black (default)', 'wporg' ), value: 'white-on-black' },
+	{ label: __( 'White on dark black', 'wporg' ), value: 'white-on-black-dark' },
 	{ label: __( 'Black on white', 'wporg' ), value: 'black-on-white' },
 	{ label: __( 'White on blue', 'wporg' ), value: 'white-on-blue' },
 ];
--- a/mu-plugins/blocks/global-header-footer/src/footer/index.js
+++ b/mu-plugins/blocks/global-header-footer/src/footer/index.js
@@ -13,10 +13,10 @@ import { useBlockProps } from '@wordpress/block-editor';
 import metadata from './block.json';
 
 const variations = [
-	{ label: __( 'White on black (default)', 'wporg' ), value: 'white-on-black' },
-	{ label: __( 'White on dark black', 'wporg' ), value: 'white-on-dark-black' },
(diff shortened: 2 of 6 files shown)


mu-plugins/blocks/global-header-footer/src/footer/index.js
```diff
@@ -14,6 +14,7 @@ import metadata from './block.json';
 
 const variations = [
 	{ label: __( 'White on black (default)', 'wporg' ), value: 'white-on-black' },
+	{ label: __( 'White on dark black', 'wporg' ), value: 'white-on-dark-black' },
 	{ label: __( 'Black on white', 'wporg' ), value: 'black-on-white' },
 	{ label: __( 'White on blue', 'wporg' ), value: 'white-on-blue' },
```
adamwoodnz: This reminds me of colour palettes I've seen where eventually one ends up adding superlatives for different shades, eg. dark grey, darker grey, darkest grey, super dark grey, etc.

To avoid this I wonder if we should reference the actual name of the preset, ie. 'White on Charcoal 1': `white-on-charcoal-1`, 'White on Blueberry 1': `white-on-blueberry-1`. I realise there is an intended simplification here with abstracting away the underlying color, but we're already seeing expansion of the palette, and number based naming is designed to address this problem, as there is an infinite scale built in.

bengreeley: I think that's a great idea. I was cringing a little when I was trying to think of a name, but sticking to the actual preset name would simplify things. I'm happy to update this PR if others agree.

ryelle: Agreed. When I first named these, I assumed this was the total set, so I went with more human-friendly names. But since we're adding more, using the actual color names would be better.

ryelle:
This is probably more a question for Design (@jasmussen), but if we're introducing a new header style, do the hover colors need to be updated too? The other styles have a noticeably different hover/submenu background color, but the default black blends in with charcoal-1 (#1e1e1e vs #1c2024).

| Default | Blue | White | This PR |
|-----|------|------|-----|
| Screenshot 2022-12-09 at 10 19 14 AM | Screenshot 2022-12-09 at 10 18 31 AM | Screenshot 2022-12-09 at 10 18 50 AM | Screenshot 2022-12-09 at 10 16 40 AM |

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
```diff
@@ -308,7 +308,7 @@
 	& .wp-block-navigation__container {
 
 		@media (--tablet) {
-			background: var(--wp-global-header--background-color);
+			background: inherit;
```
ryelle, line 311: Rather than changing this, you can add another "case" to `_common.pcss` to override `--wp-global-header--background-color`. See how it's done for the other styles here: https://github.com/WordPress/wporg-mu-plugins/blob/2ca5e5ba0046f5fa384a504ffe07df0b2b29a6c4/mu-plugins/blocks/global-header-footer/postcss/_common.pcss#L67-L90

mu-plugins/blocks/global-header-footer/src/footer/index.js
```diff
@@ -13,9 +13,10 @@ import { useBlockProps } from '@wordpress/block-editor';
 import metadata from './block.json';
 
 const variations = [
-	{ label: __( 'White on black (default)', 'wporg' ), value: 'white-on-black' },
+	{ label: __( 'White on charcoal-1', 'wporg' ), value: 'white-on-dark-black' },
```
adamwoodnz, line 16: Nothing is using this value yet, correct? So it's not too late to try to improve the naming? For me the range of values is where it's going to get messy if/when we add more variations...

```suggestion
	{ label: __( 'White on charcoal-1', 'wporg' ), value: 'white-on-charcoal-1' },
```

Also I wonder if it's still manageable to migrate the existing usages to new values while we are still relatively early in the overall redesign. All the pages are saved as patterns in source control in the various repos, so we could replace eg. `white-on-black` with `white-on-charcoal-2`, and deploy all at once.

Or maybe I'm worrying too much about this... feel free to give me a reality check 😂

adamwoodnz: For `white-on-black`
1 usage in wporg-main-2022
2 usages in wporg-showcase-2022
2 usages in wporg-parent-2021
Several in this repo

adamwoodnz: Usages are even better for `white-on-blue`

jasmussen:
>  but if we're introducing a new header style, do the hover colors need to be updated too? The other styles have a noticeably different hover/submenu background color, but the default black blends in with charcoal-1 (#1e1e1e vs #1c2024).

Good question. I guess it depends on whether we can use charcoal 1 or 2 right? (related, #308)

I'd be happy to reuse the existing charcoal for simplicity, unless @panchovm has an opportunity to chime in otherwise.

mu-plugins/blocks/global-header-footer/src/footer/index.js
bengreeley: @adamwoodnz Perhaps for a migration strategy we could do the following:
1. Add both the legacy and new values to the [switch statement](https://github.com/WordPress/wporg-mu-plugins/pull/307/files#diff-460e50b82495cab49b3ce3a702572bec919fae2cbec691937f8abbe7de6e2c33R1031) so we can support both the legacy and new new names.
2. Create an issue so we can update the content someday.
3. Eventually remove the support from the switch statements.

What do you think? I'm going to hold off until we find out whether we are going to use charcoal-1 or charcoal-2, but I think it's a good idea to implement someday.

adamwoodnz: Yeah I was a bit hesitant to suggest we reuse the same color. I think that makes sense for simplicity's sake.

Having both values in the switch could work.

ryelle: Would a [deprecation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/) work? It might only apply to JS blocks.

If so, I think the proposed migration strategy sounds good.

fcoveram:
Agree with @jasmussen on using the existing header scheme.

bengreeley:
Sounds good. In that case, I'm going to close this PR. If we want to implement the new naming convention, we always can in the future.

PR #315, opened 2022-12-20
Header: Add alerts banner

iandunn:
This adds a banner under the header for any messages that need to be presented to the user. For example, https://github.com/WordPress/wporg-two-factor/ will remove capabilities from Committers who haven't enabled 2FA, and needs to inform them why that's been done and how they can enable it.

You can test with something like:

```php
add_filter( 'wporg_global_header_alert_markup', function() {
	return '<p>Your account requires two-factor authentication, which adds an extra layer of protection against hackers. You cannot make any changes to this site until you <a href="#">enable it</a> </p>.';
} );
```

<hr />

Screenshot 2022-12-20 at 9 30 20 AM
<hr />

Screenshot 2022-12-20 at 9 31 17 AM
<hr />

Screenshot 2022-12-20 at 9 31 47 AM
<hr />

Screenshot 2022-12-20 at 9 32 25 AM

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
@@ -90,6 +90,39 @@ html {
 	}
 }
 
+.global-header__alert-banner {
+	--wp-global-header--link-color: #3e58e1;
+
+	background-color: #eff2ff;
+	color: #1e1e1e;
+	font-size: 12px;
+	line-height: 1.2;
+	text-align: center;
+	margin: 0;
+	padding: 10px 20px;
+
+	& p {
+		color: inherit; /* Override some Classic themes. */
+
+		&:first-child {
+			margin-top: 0;
+		}
+
+		&:last-child {
+			margin-bottom: 0;
+		}
+	}
+
+	& a:hover,
+	& a:focus {
+
+		/* !important is needed because the build process wraps some of the selectors in `_common.pcss` in an
+		 * `:is()` pseudo-class, which increases their specificity.
+		 */
+		color: var(--wp-global-header--link-color) !important;
+	}
+}
+
 /*
  * Gutenberg Patches
  *
--- a/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
@@ -94,7 +94,7 @@ html {
 	--wp-global-header--link-color: #3e58e1;
 
 	background-color: #eff2ff;
-	color: #1e1e1e;
+	color: var(--wp--preset--color--charcoal-1);
 	font-size: 12px;
 	line-height: 1.2;
 	text-align: center;
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -734,6 +734,23 @@ function get_download_url() {
 }
 
 /**
+ * Render a banner that other plugins can add alerts to.
+ */
+function render_header_alert_banner() {
+	$markup = '';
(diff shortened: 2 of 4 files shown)


mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
```diff
@@ -90,6 +90,38 @@ html {
 	}
 }
 
+.global-header__alert-banner {
+	--wp-global-header--link-color: #3e58e1;
+
+	background-color: #eff2ff;
+	color: #1e1e1e;
```
iandunn: There's a var for this in the News and Parent themes (albeit with different names), but nothing available universally. The other colors are similar.

I haven't seen a canonical location where all colors get defined for custom blocks, block themes, the global header/footer, and classic themes. LMK if I missed it, though.

mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
```diff
@@ -90,6 +90,38 @@ html {
...
+			margin-bottom: 0;
+		}
+	}
+
+	& a:hover,
+	& a:focus {
+		/* !important is needed because the build process wraps some of the selectors in `_common.pcss` in an
+		 * `:is()` pseudo-class, which increases their specificity.
+		 */
+		color: var(--wp-global-header--link-color) !important;
```
iandunn, line 122: This was odd, since I would have thought that, if anything, PostCSS would do the opposite and convert `:is()`  to comma-separated selectors, ala https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-is-pseudo-class. 

I didn't see a better solution to this without going down a tooling rabbit hole, but LMK if you have any ideas.

beafialho:
This looks good to me :+1:

mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
StevenDufresne: Right, I don't think we've agreed on a strategy for that. I assume that the parent theme is supposed to be the eventual source of truth.

What about having a fallback for now: `var(--wp--preset--color--charcoal-1, #1e1e1e)`?

iandunn: Doh, it looks like that already exists in `_common.pcss` and I just missed it. fixed in 7198958. thanks!

ryelle: > I assume that the parent theme is supposed to be the eventual source of truth.

That's correct, and those colors are documented here: https://github.com/WordPress/wporg-parent-2021/#colors. Since the global header/footer need to be rendered on old sites too, the base colors we need are redefined in `_common.pcss`. So you could also move over `--wp--preset--color--blueberry-4` for the `#eff2ff` value.

mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
```diff
@@ -90,6 +90,39 @@ html {
 	}
 }
 
+.global-header__alert-banner {
+	--wp-global-header--link-color: #3e58e1;
```
ryelle, line 94: Should this color be `#3858e9` (`--wp--preset--color--blueberry-1`)?

iandunn: Yeah, thanks, fixed in ac660e528fbc409c2b2b9c3abc859aeeba68c89e

mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
iandunn: 👍🏻  fixed in ac660e528fbc409c2b2b9c3abc859aeeba68c89e

Issue #318, opened 2023-01-10
Add link to 'Swag' store in global footer


coffee2code:
The [WordPress Swag Store](https://mercantile.wordpress.org/) long had a link on the homepage of w.org until the homepage redesign launched in Aug 2022. As it stands now, there isn't a notable link to the Swag store. Per matt, a link to the store should be added to the footer.

The current footer is balanced across its 5 columns with 4 links each, which would become unbalanced by the addition of this link. As a solution to that, I am also in favor of the proposal in #134 that adds a link each to the first 4 columns: Jobs, Mobile(?), Photos, Openverse (in that order). (_I think at the time "Mobile" was present and "Learn" wasn't, but that has since changed. Not sure if it should be reintroduced._) However, the Jobs link should instead go into the last column. The new links would all be conceptually appropriate for each column. (Discussion for these changes can continue in #134.)

The first column (left empty in the aforementioned ticket after my proposed change) would ideally accommodate "Swag".

PR #319, opened 2023-01-11
Footer: Add links to Mobile, Photos, Openverse, and Job Board

### LINKED ISSUE #134: Adding the Jobs link to the footer
Hallo!

@courtneyr-dev opened this [trac ticket]( https://meta.trac.wordpress.org/ticket/6068) to add a `Jobs` link in the global footer. I think this is a great idea. I'm aware that adding an additional link would throw the footer aesthetic off balance a bit but I've been exploring an option to make it look right and still be able to add that additional helpful link:
Screen Shot 2022-02-01 at 12 34 40 PM

Here we've added a Jobs link in the first column, Mobile in the second, Photos in the third, and Openverse in the fourth.

It's a pretty easy fix and I'm happy to take care of it, once there's agreement from all the right folks! Let me know. :)

coffee2code:
Adds links to notable resources and balances footer columns after introduction of 'Swag' link for #318.

Fixes #134.


StevenDufresne:
@coffee2code Nobody is assigned to this Review. Are we waiting on stakeholder opinion of some sort?

coffee2code:
_Updates to PR branch were simply to update Openverse URL._

coffee2code:
@StevenDufresne: Apologies for the lack of reply. I wrote one awhile ago but apparently never submitted it. Anyhow, this was mostly left open for consideration from others before being committed. In light of #364, there is more discussion regarding the footer links, so the two should be handled in tandem. I've referenced this PR and its proposal in a [comment there](https://github.com/WordPress/wporg-mu-plugins/issues/364#issuecomment-1487596523).

PR #320, opened 2023-01-13
Allow passing blog id to latest news block.

StevenDufresne:
For the redesign of developer.wordpress.org, we need to use this block but make requests to the developer news blog and not the news blog.

### DIFF
--- a/mu-plugins/blocks/latest-news/latest-news.php
+++ b/mu-plugins/blocks/latest-news/latest-news.php
@@ -33,8 +33,10 @@ function render_block( $attributes ) {
 
 	// Check if we can switch to the News blog.
 	// @todo Prevent switch if Rosetta, Rosetta should use local posts.
-	if ( should_switch_to_blog() && defined( 'WPORG_NEWS_BLOGID' ) ) {
-		switch_to_blog( WPORG_NEWS_BLOGID );
+	if ( should_switch_to_blog() && ( defined( 'WPORG_NEWS_BLOGID' ) || isset( $attributes['blogId'] ) ) ) {
+		// Prefer the attribute blogId over the constant.
+		$blog_id = isset( $attributes['blogId'] ) ? $attributes['blogId'] : WPORG_NEWS_BLOGID;
+		switch_to_blog( $blog_id );
 		$blog_switched = true;
 	}
 
--- a/mu-plugins/blocks/latest-news/latest-news.php
+++ b/mu-plugins/blocks/latest-news/latest-news.php
@@ -33,10 +33,8 @@ function render_block( $attributes ) {
 
 	// Check if we can switch to the News blog.
 	// @todo Prevent switch if Rosetta, Rosetta should use local posts.
-	if ( should_switch_to_blog() && ( defined( 'WPORG_NEWS_BLOGID' ) || isset( $attributes['blogId'] ) ) ) {
-		// Prefer the attribute blogId over the constant.
-		$blog_id = isset( $attributes['blogId'] ) ? $attributes['blogId'] : WPORG_NEWS_BLOGID;
-		switch_to_blog( $blog_id );
+	if ( should_switch_to_blog() && ( isset( $attributes['blogId'] ) && 0 !== $attributes['blogId'] ) ) {
+		switch_to_blog( (int) $attributes['blogId'] );
 		$blog_switched = true;
 	}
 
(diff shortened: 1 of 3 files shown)


mu-plugins/blocks/latest-news/latest-news.php
```diff
@@ -33,8 +33,10 @@ function render_block( $attributes ) {
 
 	// Check if we can switch to the News blog.
 	// @todo Prevent switch if Rosetta, Rosetta should use local posts.
-	if ( should_switch_to_blog() && defined( 'WPORG_NEWS_BLOGID' ) ) {
-		switch_to_blog( WPORG_NEWS_BLOGID );
+	if ( should_switch_to_blog() && ( defined( 'WPORG_NEWS_BLOGID' ) || isset( $attributes['blogId'] ) ) ) {
+		// Prefer the attribute blogId over the constant.
+		$blog_id = isset( $attributes['blogId'] ) ? $attributes['blogId'] : WPORG_NEWS_BLOGID;
```
ryelle: Since you're providing a default value (0), this is always going to be set. Maybe check if it's set and not 0?

This is what I see on the main homepage when loaded on a sandbox:
Screenshot 2023-01-16 at 4 49 37 PM

mu-plugins/blocks/latest-news/src/edit.js
```diff
@@ -21,14 +21,20 @@ import ServerSideRender from '@wordpress/server-side-render';
...
 
 	const onPerPageChange = ( value ) => setAttributes( { perPage: value * 1 } );
+	const onBlogIdChange = ( value ) => setAttributes( { blogId: Number( value ) } );
 
 	return (
 		<div { ...useBlockProps() }>
 			<InspectorControls>
 				<PanelBody title={ __( 'Settings', 'wporg' ) }>
+					<NumberControl
+						label={ __( 'Blog Id', 'wporg' ) }
```
ryelle:
```suggestion
						label={ __( 'Blog ID', 'wporg' ) }
```

mu-plugins/blocks/latest-news/src/block.json
```diff
@@ -11,6 +11,10 @@
 	"editorScript": "file:./index.js",
 	"style": "file:./style.css",
 	"attributes": {
+		"blogId": {
+			"type": "integer",
+			"default": 0
```
ryelle, line 16: If you made the default 8, that might also fix the "no posts found" issue.

mu-plugins/blocks/latest-news/src/edit.js
```diff
@@ -21,14 +21,20 @@ import ServerSideRender from '@wordpress/server-side-render';
...
+						onChange={ onBlogIdChange }
+						value={ blogId }
```
ryelle, line 36: Maybe add these for some quick references?
```suggestion
						value={ blogId }
						help={ __(
							'For example, 8 for wordpress.org/news, 719 for developer.wordpress.org/news',
							'wporg'
						) }
```

Screenshot 2023-01-16 at 5 09 34 PM

ryelle (commented):
Thanks for adding this! It's also needed for the developer blog, see https://github.com/WordPress/wporg-developer-blog/issues/30 (though in that case it also needs a tag filter, but we can do that later).

I ran into an issue with the default attribute state, but I suggested two fixes, whichever you'd rather use.

mu-plugins/blocks/latest-news/latest-news.php
StevenDufresne: Right, It was originally undefined but that created some react issues with controlled/uncontrolled inputs. I'll check for `0`.

ryelle: I haven't spun anything up to test it, but will you need to update the block on the homepage when this is merged? Or will it still default to the News blog?

StevenDufresne: We'll need to update the block on the [homepage](https://github.com/WordPress/wporg-main-2022/blob/d3fc83080ed67056133537c2229da69619f12db2/source/wp-content/themes/wporg-main-2022/patterns/home.php#L297). We could alternatively just default it to `8` but having no default probably makes sense. It seems to be showing up more in different contexts. 

Since we now check that it isn't `0` it should fix the "no post found" issues as it won't switch blogs.

ryelle (approved):
Looks good, and I'm fine with a default of `0`, just make sure to update the homepage when you deploy it 🙂

StevenDufresne:
Updated in: https://github.com/WordPress/wporg-main-2022/commit/d97c9abd091ed07409a56ec59373ff5a4d4b876d

PR #322, opened 2023-01-18
Add a Notice Block

ryelle:
The new redesign uses a different style of callout box from the hardcoded `.callout` or shortcode previously used. This PR adds a "notice" block to render the 4 different notice styles.

- Tip, green background with ℹ️  icon
- Info, blue background with ❓ icon
- Alert, yellow background with ⚠️  icon
- Warning, red background with ❌  icon

I forked this block from [one I'd written as a demo](https://github.com/ryelle/notice-block/) for [a WordCamp talk](https://ryelle.codes/wcmontclair-2022/).

Having a Notice block will let us keep the 4 different notice styles consistent, rather than hardcoding in background colors on group blocks, etc. This also restricts the content that can be in a notice, to act as guardrails to keep notices to the point.

| Frontend | Editor |
|------|-------|
| notices | Screenshot 2023-01-17 at 7 03 57 PMScreenshot 2023-01-17 at 6 52 29 PM |

We can use it in code/templates like this:

```html
<!-- wp:wporg/notice {"type":"alert"} -->
<div class="wp-block-wporg-notice is-alert-notice">
<div class="wp-block-wporg-notice__icon"></div>
<div class="wp-block-wporg-notice__content"><p>[Notice content here.]</p></div>
</div>
<!-- /wp:wporg/notice -->
```

To change the type, you need to change both the attribute and the class name. To change the text, just enter content in the paragraph. There can be multiple `p`s (though it's not likely).

(It's not a dynamic block because it was built editor-first, but if we hate that it can be switched to dynamic)

**To test:**

- Try adding the block to some post content, it should not break
- Changing the type should work
- Try adding the block to a template
- It should continue to work

The block does support padding, margin, and different font sizes, just in case we need those.


mu-plugins/blocks/notice/src/icon/index.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Internal dependencies
+ */
+import { ReactComponent as Alert } from './library/alert.svg';
+import { ReactComponent as Info } from './library/info.svg';
+import { ReactComponent as Tip } from './library/tip.svg';
+import { ReactComponent as Warning } from './library/warning.svg';
+
+export default function ( { type /*, label */ } ) {
```
StevenDufresne:
```suggestion
export default function ( { type } ) {
```

ryelle: 👍🏻  — I thought I'd go back and add an aria label, but changed approach.

StevenDufresne:
I'm having issues building:

```
ERROR in ./mu-plugins/blocks/notice/src/icon/index.js 9:0-66
Module not found: Error: Can't resolve 'url-loader' in '/Users/stevendufresne/dev/wp-github/wporg-mu-plugins'
 @ ./mu-plugins/blocks/notice/src/edit.js 13:0-26 68:28-32
 @ ./mu-plugins/blocks/notice/src/index.js 10:0-43 14:8-12 16:14-29
```

Didn't have a chance to debug, but surely related to importing `svg`s.

ryelle:
Huh, `url-loader` should be part of wp-scripts. Could be a version mismatch? or something. I'll look into it.

ryelle:
Well, looks like it's been part of wp-scripts for 3 years, so it should be there… but the action is also having the same problem: https://github.com/WordPress/wporg-mu-plugins/actions/runs/3945370268/jobs/6752203752

ryelle:
Well, I'm guessing it has to do with how the `build` script works here, but adding it as a local dependency (https://github.com/WordPress/wporg-mu-plugins/pull/322/commits/1e09f1db6eecd191643c0379456cd2b1ea0c2320) seems to have fixed the action at least. Does that work for you?

mu-plugins/blocks/notice/postcss/style.pcss
```diff
@@ -0,0 +1,78 @@
+.wp-block-wporg-notice {
+	--wp--custom--wporg-notice--color--background: var(--wp--preset--color--acid-green-3);
+	--wp--custom--wporg-notice--color--text: var(--wp--preset--color--charcoal-1);
```
StevenDufresne, line 3: When I look at other notice-like components, the links are still the default link color:

Screen Shot 2023-01-19 at 9 28 54 AM

Should we also use that color? Or maybe make the component support link color updates?

StevenDufresne:
This looks great. Merge when ready.

mu-plugins/blocks/notice/postcss/style.pcss
ryelle: I wouldn't call the language suggestion a notice, since it doesn't have an icon, the default text is smaller, and it's a full-width stripe.

I was referencing another mockup when I made this decision, I thought it was Themes but I don't see a notice with links there. Found this one on Jobs:

Screenshot 2023-01-19 at 9 19 16 AM

I think this is correct, but if we end up with a notice that needs a different link color we can update the block.

PR #332, opened 2023-02-02
Allow for content filtering before we retrieve that content.

StevenDufresne:
Currently, if we have a template full of blocks and no post-content in [wporg-developer](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/patterns/single.php). As a result, we have to inject the blocks into the `the_content` (i think :)). 

However, filtering the content in the theme seems to be done too late. @dd32 suggested that the best strategy is to do the following:

`$content = apply_filters( 'the_content', get_the_content( $post ), $post );`

This does in fact seem to work.

### DIFF
--- a/mu-plugins/blocks/table-of-contents/index.php
+++ b/mu-plugins/blocks/table-of-contents/index.php
@@ -36,7 +36,9 @@ function init() {
  * @return string Returns the block markup.
  */
 function render( $attributes, $content, $block ) {
-	$items = get_headings( get_the_content() );
+	$post = get_post();
+	$post_content = apply_filters( 'the_content', get_the_content( $post ), $post );
+	$items = get_headings( $post_content );
 	if ( ! $items ) {
 		return '';
 	}
@@ -69,8 +71,8 @@ function render( $attributes, $content, $block ) {
 	// Use the parsed headings & IDs to inject IDs into the post content.
 	add_filter(
 		'the_content',
-		function( $content ) use ( $items ) {
-			return inject_ids_into_headings( $content, $items );
+		function() use ( $items, $post_content ) {
+			return inject_ids_into_headings( $post_content, $items );
 		},
 		5 // Run early, before special character handling, so the items match.
 	);
--- a/mu-plugins/blocks/table-of-contents/index.php
+++ b/mu-plugins/blocks/table-of-contents/index.php
@@ -37,7 +37,7 @@ function init() {
  */
 function render( $attributes, $content, $block ) {
 	$post = get_post();
-	$post_content = apply_filters( 'the_content', get_the_content( $post ), $post );
+	$post_content = apply_filters( 'the_content', get_the_content( $post ) );
 	$items = get_headings( $post_content );
 	if ( ! $items ) {
 		return '';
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/table-of-contents/index.php
```diff
@@ -69,8 +71,8 @@ function render( $attributes, $content, $block ) {
 	// Use the parsed headings & IDs to inject IDs into the post content.
 	add_filter(
 		'the_content',
-		function( $content ) use ( $items ) {
-			return inject_ids_into_headings( $content, $items );
+		function() use ( $items, $post_content ) {
+			return inject_ids_into_headings( $post_content, $items );
```
dd32: This one should remain unchanged, as we want to operate on the $content that's being passed through the filter, not what we retrieved earlier, I think..

StevenDufresne: Hmm, it wasn't returning the content properly without this change.

mu-plugins/blocks/table-of-contents/index.php
```diff
@@ -36,7 +36,9 @@ function init() {
  * @return string Returns the block markup.
  */
 function render( $attributes, $content, $block ) {
-	$items = get_headings( get_the_content() );
+	$post = get_post();
+	$post_content = apply_filters( 'the_content', get_the_content( $post ), $post );
```
ryelle: It looks like `the_content` only takes the `$content` as a param, not post. I don't think this breaks anything, but for consistency [with core](https://developer.wordpress.org/reference/hooks/the_content/), it should probably be 

```suggestion
	$post_content = apply_filters( 'the_content', get_the_content( $post ) );
```

mu-plugins/blocks/table-of-contents/index.php
```diff
@@ -36,7 +36,9 @@ function init() {
  * @return string Returns the block markup.
  */
 function render( $attributes, $content, $block ) {
-	$items = get_headings( get_the_content() );
+	$post = get_post();
```
ryelle: I should have used the `postId` from the context in this block, oops. If you want to, you could make that change too (see https://github.com/WordPress/wporg-developer/pull/181/commits/105875f5d3880791ac01594daa6c236b976b5fae). Not a blocker though, that doesn't change the work you've got here.

mu-plugins/blocks/table-of-contents/index.php
ryelle: Probably because of the priority — this had to be priority 5, but your update to the content in https://github.com/WordPress/wporg-developer/pull/194 happens later, at prio 10. If I switch the filter priority in 194 to `4`, it works.

StevenDufresne: Ah that makes sense. I'll revert this out.

mu-plugins/blocks/table-of-contents/index.php
StevenDufresne: Done in https://github.com/WordPress/wporg-mu-plugins/pull/332/commits/ca79e8c4ffcdf80828246e3a99bff975765518f9. 👍

StevenDufresne:
I'll merge this making sure to test documentation before deploying.

PR #335, opened 2023-02-09
Table of contents block: Include the top margin in fixed sidebar detection

### LINKED ISSUE #333: Table Of Contents content is hidden off screen.
Content in the Table of Contents control seem to be hidden 

Page: https://developer.wordpress.org/redesign-test/rest-api/reference/page-revisions/

## Screenshot

Screen Shot 2023-02-06 at 12 50 20 PM

ryelle:
Fixes #333. The math to determine when the sidebar is fixed vs static was not including the top margin & space needed for the "Back to top" link, so as long as the viewport had enough space for the blue box + search, the fixed sidebar would be triggered. This meant that on screens within ~100px of the sidebar, it would incorrectly trigger and the bottom of the ToC would be offscreen.

**To test**

1. Try on both developer and documentation
2. Find a page with a ToC that fits on one page, and shrink your browser so the bottom of the screen hits in the pink area:
    Screen Shot 2023-02-09 at 15 29 02
3. The sidebar should not be fixed, it stays aligned to the top content when scrolling the page
4. Expand your viewport to beyond the pink area and reload
5. The ToC should fit on screen, so when you scroll the page it stays visible on screen
6. Ensure footer collision still works by scrolling to the bottom of the page
7. The ToC should not overlap the footer content

### DIFF
--- a/mu-plugins/blocks/table-of-contents/src/view.js
+++ b/mu-plugins/blocks/table-of-contents/src/view.js
@@ -67,12 +67,19 @@ function onScroll() {
 
 function init() {
 	const container = document.querySelector( '.wp-block-wporg-table-of-contents' );
+	// Margin offset from the top of the sidebar.
+	const gap = getCustomPropValue( '--wp--preset--spacing--edge-space' );
 
 	if ( container ) {
+		// Usable viewport height.
 		const viewHeight = window.innerHeight - FIXED_HEADER_HEIGHT;
+		// Get the height of the sidebar, plus the top margin and 50px for the
+		// "Back to top" link, which isn't visible until `is-fixed-sidebar` is
+		// added, therefore not included in the parentNode.offsetHeight value.
+		const sidebarHeight = container.parentNode?.offsetHeight + gap + 50;
 		// If the table of contents sidebar is shorter than the view area, apply the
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/table-of-contents/src/view.js
```diff
@@ -67,12 +67,19 @@ function onScroll() {
...
+	// Margin offset from the top of the sidebar.
+	const gap = getCustomPropValue( '--wp--preset--spacing--edge-space' );
 
 	if ( container ) {
+		// Usable viewport height.
 		const viewHeight = window.innerHeight - FIXED_HEADER_HEIGHT;
+		// Get the height of the sidebar, plus the top margin and 50px for the
+		// "Back to top" link, which isn't visible until `is-fixed-sidebar` is
+		// added, therefore not included in the parentNode.offsetHeight value.
+		const sidebarHeight = container.parentNode?.offsetHeight + gap + 50;
```
StevenDufresne, line 79: Change to the font size and the margin of the "Back to Top" link could throw this off, but I think this is an immediate improvement.

StevenDufresne (approved):
It's working for the example in #333 and I tested it on my sandbox, seems to work for all the examples I tested.

PR #337, opened 2023-02-10
Try: Override callout shortcodes in favor of the notice block

### LINKED ISSUE #315: Update callout design to reflect the WordPress.org Design Library
The callouts still have the "old" design. The new design is available [here](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=1264-4116&mode=design&t=PXxc0jZv4uQ2fdXW-0), which should ideally be used across all wordpress.org. 

| Current | New Design |
|-|-|
| image | image|

I am not sure if this issue should be in this repo or another. @adamwoodnz if the later, let me know where, and I will move this over. Thanks!

ryelle:
Closes https://github.com/WordPress/wporg-developer/issues/315

Remove the existing "callout" shortcodes, and replace them with the Notice block.

docs-tip

docs-alert-info

Some of the shortcode content can be very long
docs-long-info

This shortcode is also used on classic themes, since it's part of the handbook code. I've updated the CSS to include fallbacks for the custom properties, so it should still work. For example, on make/core's handbook.

make-info

Is the sanitization + fitlering necessary? I grabbed it from [the existing callout shortcode](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/handbook/inc/callout-boxes.php#L116).

### DIFF
--- a/mu-plugins/blocks/notice/index.php
+++ b/mu-plugins/blocks/notice/index.php
@@ -6,7 +6,8 @@
 
 namespace WordPressdotorg\MU_Plugins\Notice;
 
-add_action( 'init', __NAMESPACE__ . '\init' );
+// Run after `WPorg_Handbook_Callout_Boxes` registers the shortcodes.
+add_action( 'init', __NAMESPACE__ . '\init', 11 );
 
 /**
  * Registers the block using the metadata loaded from the `block.json` file.
@@ -17,4 +18,47 @@ add_action( 'init', __NAMESPACE__ . '\init' );
  */
 function init() {
 	register_block_type( __DIR__ . '/build' );
+
+	foreach ( [ 'info', 'tip', 'alert', 'tutorial', 'warning' ] as $shortcode ) {
+		remove_shortcode( $shortcode );
+		add_shortcode( $shortcode, __NAMESPACE__ . '\render_shortcode' );
+	}
+}
+
+/**
+ * Display the callout shortcodes using the notice block.
+ *
+ * @param array|string $attr    Shortcode attributes array or empty string.
+ * @param string       $content Shortcode content.
+ * @param string       $tag     Shortcode name.
+ *
+ * @return string Shortcode output as HTML markup.
+ */
+function render_shortcode( $attr, $content = '', $tag ) {
+	$shortcode_mapping = array(
+		'info'     => 'info',
+		'tip'      => 'tip',
+		'alert'    => 'alert',
+		'tutorial' => 'tip',
+		'warning'  => 'warning',
+	);
+
+	$type = $shortcode_mapping[ $tag ] ?: 'tip';
+
+	// Sanitize message content.
+	$content = wp_kses_post( $content );
+	// Temporarily disable o2 processing while formatting content.
+	add_filter( 'o2_process_the_content', '__return_false', 1 );
+	$content = apply_filters( 'the_content', $content );
+	remove_filter( 'o2_process_the_content', '__return_false', 1 );
+
+	$block_markup = <<<EOT
+<!-- wp:wporg/notice {"type":"$type"} -->
+<div class="wp-block-wporg-notice is-{$type}-notice">
+<div class="wp-block-wporg-notice__icon"></div>
+<div class="wp-block-wporg-notice__content">$content</div></div>
+<!-- /wp:wporg/notice -->
+EOT;
+
+	return do_blocks( $block_markup );
 }
--- a/mu-plugins/blocks/notice/index.php
+++ b/mu-plugins/blocks/notice/index.php
@@ -52,13 +52,21 @@ function render_shortcode( $attr, $content = '', $tag ) {
 	$content = apply_filters( 'the_content', $content );
 	remove_filter( 'o2_process_the_content', '__return_false', 1 );
 
+	// Create a unique placeholder for the content.
+	// Directly processing `$content` with `do_blocks` can lead to buggy layouts on make.wp.org.
+	// See https://github.com/WordPress/wporg-mu-plugins/pull/337#issuecomment-1819992059.
+	$placeholder = '<!-- CONTENT_PLACEHOLDER -->';
+
 	$block_markup = <<<EOT
 <!-- wp:wporg/notice {"type":"$type"} -->
 <div class="wp-block-wporg-notice is-{$type}-notice">
 <div class="wp-block-wporg-notice__icon"></div>
-<div class="wp-block-wporg-notice__content">$content</div></div>
+<div class="wp-block-wporg-notice__content">$placeholder</div></div>
 <!-- /wp:wporg/notice -->
 EOT;
 
-	return do_blocks( $block_markup );
+	$processed_markup = do_blocks( $block_markup );
+	$final_markup = str_replace( $placeholder, $content, $processed_markup );
+
+	return $final_markup;
 }
--- a/mu-plugins/blocks/notice/postcss/style.pcss
+++ b/mu-plugins/blocks/notice/postcss/style.pcss
@@ -1,17 +1,17 @@
 .wp-block-wporg-notice {
-	--wp--custom--wporg-notice--color--background: var(--wp--preset--color--acid-green-3);
-	--wp--custom--wporg-notice--color--text: var(--wp--preset--color--charcoal-1);
+	--wp--custom--wporg-notice--color--background: var(--wp--preset--color--acid-green-3, #e2ffed);
+	--wp--custom--wporg-notice--color--text: var(--wp--preset--color--charcoal-1, #1e1e1e);
 
 	&.is-info-notice {
-		--wp--custom--wporg-notice--color--background: var(--wp--preset--color--blueberry-4);
+		--wp--custom--wporg-notice--color--background: var(--wp--preset--color--blueberry-4, #eff2ff);
 	}
 
 	&.is-alert-notice {
-		--wp--custom--wporg-notice--color--background: var(--wp--preset--color--lemon-3);
+		--wp--custom--wporg-notice--color--background: var(--wp--preset--color--lemon-3, #fffdd6);
 	}
 
 	&.is-warning-notice {
-		--wp--custom--wporg-notice--color--background: var(--wp--preset--color--pomegrade-3);
+		--wp--custom--wporg-notice--color--background: var(--wp--preset--color--pomegrade-3, #ffe9de);
 	}
 
 	display: grid;
@@ -33,6 +33,10 @@
 		margin-block-end: 0;
 	}
 
+	& br:first-child {
+		display: none;
+	}
+
 	& a:link,
 	& a:visited,
 	& a:hover,
@@ -43,7 +47,7 @@
 
 	&.alignleft,
 	&.alignright {
-		max-width: calc(var(--wp--style--global--content-size) * 0.66);
+		max-width: calc(var(--wp--style--global--content-size, 680px) * 0.66);
 	}
 }
 
--- a/mu-plugins/blocks/notice/postcss/style.pcss
+++ b/mu-plugins/blocks/notice/postcss/style.pcss
@@ -38,14 +38,6 @@
 		display: none;
 	}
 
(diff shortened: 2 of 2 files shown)


adamwoodnz:
Tried this in sandbox on https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/ and it breaks quite badly. I'm not sure what exactly, but seems to be do with content parsing. The page is very slow to load and looks like this when fully loaded:

Screenshot 2023-11-21 at 12 53 39 PM

pkevan:
@adamwoodnz I had a quick look at this, and it appears that the `info` shortcode gets into some kind of loop, when you remove it from the `foreach` the page renders quickly, so probably some bug in there.

outdoor2kode:
> Is the sanitization + fitlering necessary? I grabbed it from [the existing callout shortcode](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/handbook/inc/callout-boxes.php#L116).

I think the [sanitization](https://github.com/WordPress/wporg-mu-plugins/pull/337/files#diff-4c217b8869c069d64cb900a20ba183d4665d01b417024ec0b67e672958a80f09R49) is still needed as the shortcode accepts users' input, but not so sure about the [filter](https://github.com/WordPress/wporg-mu-plugins/pull/337/files#diff-4c217b8869c069d64cb900a20ba183d4665d01b417024ec0b67e672958a80f09R51).

So I've checked a bit if `o2` is activated on `wp.org`, `make.wp.org`,  and `developer.wp.org`, and only `make.wp.org` has `o2 cross-Posting Access` network-activated; no other o2-related plugins are activated on any of these three sites (search with 'o2'). Does that mean we can remove the filter safely? 

@dd32 I noticed that you have made several commits in the o2 plugin, so perhaps you have some thoughts on this. 🙂

pkevan:
> Does that mean we can remove the filter safely?

Do we envisage this not being used on make.wp.org? Given it's a mu-plugin, it's probably best to keep the filters in there, since it would automatically be loaded anyway. You'd ideally check the occurrence of the legacy shortcodes too.

outdoor2kode:
The root cause is [this line](https://github.com/WordPress/wporg-mu-plugins/pull/337/files#diff-4c217b8869c069d64cb900a20ba183d4665d01b417024ec0b67e672958a80f09R52). 
This filter would execute all functions hooked onto it and then formats `$content`, it ends up adding various extra tags, `\n`, etc. Something among them seems to cause `do_blocks` to get into a loop. 

Removing this filter directly resolves the issue. (i.e., comment out [these three lines](https://github.com/WordPress/wporg-mu-plugins/pull/337/files#diff-4c217b8869c069d64cb900a20ba183d4665d01b417024ec0b67e672958a80f09R51-R53))

Additionally, it also needs to remove the `<p>` tags from within the shortcode content, as Gutenberg interprets them as blank lines. [PHP: PHPUnit](https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/) ⬇️

| With `<p>` tag | Without `<p>` tag + Add a newline |
|-|-|
| image image | image image |

[Adv Admin landing page](https://developer.wordpress.org/redesign-test/wp-admin/post.php?post=143559&action=edit) works correctly.
image

@adamwoodnz I'll run a few more tests based on this tomorrow to see if there are no issues. It'd be nice if you could also take a look. The scope of this change seems quite extensive, it's great to have extra eyes to help check.

outdoor2kode:
> Do we envisage this not being used on make.wp.org

The filter seems there for the plugin `o2`, but it doesn't seem to be activated anymore.

> You'd ideally check the occurrence of the legacy shortcodes too.

Probably a dumb question. What does a legacy shortcode refer to there, any examples?

But since [the_content](https://github.com/WordPress/wporg-mu-plugins/pull/337/files#diff-4c217b8869c069d64cb900a20ba183d4665d01b417024ec0b67e672958a80f09R52) seems to be the root cause here, once it's removed, `o2_process_the_content` filter is no longer required.

pkevan:
> What does a legacy shortcode refer to there, any examples?

The ones being replaced by this code is what I was attempting to refer to 😄 `[ 'info', 'tip', 'alert', 'tutorial', 'warning' ]`

outdoor2kode:
I realized that we cannot remove the 'the_content' filter, as this would lead to losing some features like auto-linking or small features like tooltips.

image

Since earlier I placed the same content that's broken on production locally and it did not break, I guess that the problem might be on `$content`. After being processed by `do_blocks`, it is not compatible with the traditional theme of make.wp.org. 

So eventually I just move $content outside the `do_blocks` to prevent it from being processed by the function.
I've tested several pages on both Make and Developer sites while sandboxed, and it seems like there were no issues. 

I guess the remaining task is first to convert the [content that contains](https://github.com/search?q=repo%3AWordPress%2Fgutenberg+callout+language%3AMarkdown&type=code&l=Markdown) `<div class="callout callout-warning">` into shortcodes. Both make.wp.org and developer.wp.org still have content using this format. 
Then include @ndiego's [idea](https://github.com/WordPress/wporg-mu-plugins/issues/549) of extending the logic here to cover the GitHub markdown.

Reference: https://github.com/WordPress/wporg-developer/issues/315#issuecomment-1791563305

adamwoodnz:
Created a test post https://developer.wordpress.org/redesign-test/2023/12/12/callouts/

### Screenshots

| Before | After |
|-|-|
| Screenshot 2023-12-13 at 9 44 10 AM | Screenshot 2023-12-13 at 9 44 30 AM |

adamwoodnz:
Current Developer pages look ok

Screenshot 2023-12-13 at 9 46 44 AM

adamwoodnz:
New developer pages look good

Screenshot 2023-12-13 at 9 50 52 AM

Screenshot 2023-12-13 at 9 52 00 AM

adamwoodnz:
> Tried this in sandbox on https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/ and it breaks quite badly.

This page is fixed, apart from the new line issue 🎉 

Screenshot 2023-12-13 at 9 49 24 AM

@renintw what do you propose for the empty p tags, does this involve a lot of handbook editing?

Some have the issue, others don't

Screenshot 2023-12-13 at 9 56 27 AM

ndiego:
Colors and icons look great! Can the links be in blueberry and a border-radius of 2px be applied? The [Figma design](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=1264-4116&mode=design&t=fZncTwxBKjdc0lXs-0) did not show links (that I can find), but we use blueberry in other similar content areas.

adamwoodnz:
There's an alert on https://make.wordpress.org/core/handbook/about/release-cycle/releasing-major-versions/ which isn't being captured for some reason

Screenshot 2023-12-13 at 10 03 47 AM

The alert on https://make.wordpress.org/core/handbook/about/organization/ is ok though

Trying some other handbooks and found https://make.wordpress.org/test/handbook/test-reports/ has on old style info box too

Screenshot 2023-12-13 at 10 09 44 AM

adamwoodnz:
> There's an alert on https://make.wordpress.org/core/handbook/about/release-cycle/releasing-major-versions/ which isn't being captured for some reason

> Trying some other handbooks and found https://make.wordpress.org/test/handbook/test-reports/ has on old style info box too

Turns out these are both using `<div class="callout"...` rather than shortcodes, the same as the Block Editor handbook does.

I think we'll need to do a secondary PR to update styles for these in the Developer repo, as a short term fix before we add support for the GitHub markdown approach previously discussed.

For these Make sites I'm not sure how concerned we should be about the inconsistency if we merge this. I don't think it's a major. Thoughts @ndiego @renintw ?

adamwoodnz (commented):
Works for me; tested on old and new Developer sites, Make Core, Make Test and Make Meta. One question above regarding inconsistencies with handbooks using callout markup rather than shortcodes.

ndiego:
> Works for me; tested on old and new Developer sites, Make Core, Make Test and Make Meta. One question above regarding inconsistencies with handbooks using callout markup rather than shortcodes.

Will all the Developer sites be consistent?

adamwoodnz:
> Will all the Developer sites be consistent?

Do you mean all the callouts/shortcodes on the Developer site? Not without further work for the Block Editor Handbook. I think our short term options are:

1. Add some styles for the existing `.callout` markup to either the parent theme (if we think this will stick around, or the Documentation site also uses this) or the Developer theme
2. Migrate the Block Editor Handbook callout markup to shortcodes to take advantage of this PR.

I think I prefer 1 as 2 feels like a backward step.

Medium term I think we should push forward with the plan of using GitHub markdown and coverting that to Notice blocks.

adamwoodnz:
> 1. Add some styles for the existing `.callout` markup to either the parent theme (if we think this will stick around, or the Documentation site also uses this) or the Developer theme

PR raised: https://github.com/WordPress/wporg-developer/pull/429

outdoor2kode:
> @renintw what do you propose for the empty p tags, does this involve a lot of handbook editing?

It doesn't need a lot of editing on a single handbook post, but it would need a lot on all the posts.
It simply needs to remove the `<p>` tags from within the shortcode content, as Gutenberg interprets them as blank lines. [PHP: PHPUnit](https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/) ⬇️

| With `<p>` tag | Without `<p>` tag + Add a newline |
|-|-|
| image image | image image |

We can write a regex to replace all the p tags inside the shortcode, and also starting from the second p tag, in addition to getting rid of it, we also need to add a new line before it. 
But I would more prefer to possibly share with the community, asking folks to see and report back, and then make modifications.

outdoor2kode:
> I think I prefer 1 as 2 feels like a backward step.

Thanks for opening the PR for it, 1 does make more sense than 2.

outdoor2kode:
> Colors and icons look great! Can the links be in blueberry and a border-radius of 2px be applied? The [Figma design](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=1264-4116&mode=design&t=fZncTwxBKjdc0lXs-0) did not show links (that I can find), but we use blueberry in other similar content areas.

I've changed the border-radius to `2px`.

For links color, they should've been blueberry, but somehow the styles [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/notice/postcss/style.pcss#L36-L42) are applied, even though those actions have not been performed on links. Removing those pseudo-classes does make it render blueberry.

Do you run into this as well @adamwoodnz?

adamwoodnz:
Looks like we should change the notice font size to `--wp--preset--font-size--small` too, based on this [comment](https://github.com/WordPress/wporg-developer/pull/429#issuecomment-1854552136) and the [designs](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?node-id=1264%3A4116&mode=dev).

adamwoodnz:
> For links color, they should've been blueberry, but somehow the styles [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/notice/postcss/style.pcss#L36-L42) are applied, even though those actions have not been performed on links. Removing those pseudo-classes does make it render blueberry.
> 
> Do you run into this as well @adamwoodnz?

Yep same here. Shall we add a new `--wp--custom--wporg-notice--color--link` var and update those link styles then?

adamwoodnz (commented):
2 suggestions inline then I think we can merge this 👍

outdoor2kode:
> Yep same here. Shall we add a new --wp--custom--wporg-notice--color--link var and update those link styles then?

Huh, I see the problem now. It shouldn't have added `a:link`, as it conflicts with `a:visited`. The former applies to all unvisited hyperlinks, while the latter is for links that have already been visited. 
I think I'll just remove all actions here, unless we need to change the text to black after an action, which I don't see in the design.

PR #340, opened 2023-02-14
Update Openverse URL in global header

### LINKED ISSUE #339: Update Openverse URL in global header
Could someone update the Openverse url in [w.org](http://w.org/)’s primary navigation from [wordpress.org/openverse](http://wordpress.org/openverse) to https://openverse.org/?

It’s working fine as is, but we might as well skip the 301 redirect for users.

adamwoodnz:
Closes #339 

Updates the Openverse URL in the global header to https://openverse.org/, which avoids a 301 redirect.


zackkrida (approved):
Looks good 👍

PR #344, opened 2023-02-22
Check for reference before removing event.

### LINKED ISSUE #343: Horizontal Slider is causing js errors in theme directory.
The slider component is used within the Backbone js code in the theme directory to render style variations. When users click "Return to Themes List" the view gets destroyed and triggers errors in the block which wants to gracefully remove the event listeners but since the view is abruptly removed, references are lost and a js error occurs.

**Error**
```
react-dom.min.js?ver=18:1 TypeError: Cannot read properties of null (reading 'removeEventListener')
    at index.js?ver=b5ae14b0e3cd188089f4:1:3091
    at Or (react-dom.min.js?ver=18:1:73042)
    at Ir (react-dom.min.js?ver=18:1:73228)
    at El (react-dom.min.js?ver=18:1:92472)
    at pl (react-dom.min.js?ver=18:1:85955)
    at e.unmountComponentAtNode (react-dom.min.js?ver=18:1:128620)
    at HTMLDivElement.<anonymous> (theme.min.js?ver=1677042452:1:14982)
    at Function.each (load-scripts.php?c=0&load%5Bchunk_0%5D=jquery-core,jquery-migrate,wp-polyfill-inert,regenerator-runtime,wp-polyfill&ver=6.2-beta3-55401:2:3003)
    at e.<computed>.each (load-scripts.php?c=0&load%5Bchunk_0%5D=jquery-core,jquery-migrate,wp-polyfill-inert,regenerator-runtime,wp-polyfill&ver=6.2-beta3-55401:2:1481)
    at n.unmountReactAssets (theme.min.js?ver=1677042452:1:14957)
```

### How To Reproduce
1. Visit https://wordpress.org/themes/twentytwentytwo/
2. Scroll down and back up
3. Click "Return to Themes List"
4. Notice console error logged
5. In order to proceed you, have to click the button again.

StevenDufresne:
Fixes: #343

This PR checks to make sure the element reference exists before trying to remove the event listener.

## How to Test
1. Apply to Sandbox
2. See #343 for steps

- Expect to see no console.log errors
- Expect to only have to click once to return.

### DIFF
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -83,7 +83,9 @@ function Block( { items, title } ) {
 		outerRef.current.addEventListener( 'scroll', handleScrollEvent );
 
 		return () => {
-			outerRef.current.removeEventListener( 'scroll', handleScrollEvent );
+			if ( outerRef.current ) {
+				outerRef.current.removeEventListener( 'scroll', handleScrollEvent );
+			}
 		};
 	}, [ outerRef ] );
 
--- a/mu-plugins/blocks/horizontal-slider/src/block.js
+++ b/mu-plugins/blocks/horizontal-slider/src/block.js
@@ -83,9 +83,7 @@ function Block( { items, title } ) {
 		outerRef.current.addEventListener( 'scroll', handleScrollEvent );
 
 		return () => {
-			if ( outerRef.current ) {
-				outerRef.current.removeEventListener( 'scroll', handleScrollEvent );
-			}
+			outerRef.current?.removeEventListener( 'scroll', handleScrollEvent );
 		};
 	}, [ outerRef ] );
 


mu-plugins/blocks/horizontal-slider/src/block.js
```diff
@@ -83,7 +83,9 @@ function Block( { items, title } ) {
 		outerRef.current.addEventListener( 'scroll', handleScrollEvent );
 
 		return () => {
-			outerRef.current.removeEventListener( 'scroll', handleScrollEvent );
+			if ( outerRef.current ) {
+				outerRef.current.removeEventListener( 'scroll', handleScrollEvent );
+			}
```
adamwoodnz: [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is a nice way to handle this, looks like [browser support](https://caniuse.com/?search=optional%20chaining) is good

```suggestion
			outerRef.current?.removeEventListener( 'scroll', handleScrollEvent );
```

adamwoodnz (approved):
👍 Tested the current and suggested change, both work for me

mu-plugins/blocks/horizontal-slider/src/block.js
StevenDufresne: Yep, maybe it's time to move on to optional chaining :) Feels weird.

PR #345, opened 2023-02-23
Add DB-backed WP_Session_Tokens implementation

dd32:
WordPress sessions (Which are not used on WordPress.org currently) use user-meta to store active user sessions.

This class, based on the implementation on WordPress.com, stores these sessions in a custom table.

This is pending changes to allow the use of sessions on WordPress.org, such as dealing with Trac ~and the Codex~ the codex doesn't share cookies, only user tables..

This is preferably needed so that we can use the upstream 2fa session code for https://github.com/WordPress/wporg-two-factor/issues/43

### DIFF
--- a/mu-plugins/db-user-sessions/class-tokens.php
+++ b/mu-plugins/db-user-sessions/class-tokens.php
@@ -7,11 +7,6 @@ class Tokens extends \WP_Session_Tokens {
 
 	protected function get_sessions() {
 		$user_sessions = $this->get_all_user_sessions();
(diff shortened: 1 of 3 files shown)


dd32:
Draft support for this in the TracWordPressCookiesPlugin: https://github.com/dd32/TracWordPressCookiesPlugin/pull/2 - I haven't tested this in my test trac environment, but it looks like it'll work and is much simpler than the [user_meta implementation](https://github.com/dd32/TracWordPressCookiesPlugin/pull/1).

PR #348, opened 2023-03-03
Make the breadcrumb more readable.

### LINKED ISSUE #325: Site Breadcrumbs should collapse on mobile
We've solved the breadcrumbs wrapping on mobile in a few sites so far, https://github.com/WordPress/wporg-developer/pull/166, https://github.com/WordPress/wporg-documentation-2022/pull/19, and now https://github.com/WordPress/wporg-showcase-2022/pull/96.

Let's DRY this up and set some better defaults for the block's CSS.

StevenDufresne:
Fixes #325 

Building on conversations in https://github.com/WordPress/wporg-showcase-2022/pull/96, this PR attempts to do the following:
- Make the breadcrumb work on mobile
- Apply truncation to reduce the length of the breadcrumb on desktop
- Try to do as much as possible with CSS

I also think there's some work to be done to reduce the length of the crumbs in the data. This hopes to be a fallback in case we keep these surprisingly long crumbs.

# Mobile

The idea here is to keep everything visible but allow it to wrap naturally. In doing so, we do get a broken menu dropdown, we'll need to follow up with a PR to fix that if we agree on this approach.

## Updates crumbs `display` so it stacks on mobile

| Before | After |
| --- | --- |
| developer wordpress org_redesign-test_block-editor_how-to-guides_data-basics_4-building-a-create-page-form_ (1) |  |

## Move divider to `::after`

If the separator was an arrow, it would make sense to have them on the left. However, to me at least, the slash as the first character on a line reads weird.

| Before | AFter |
| --- | --- |
|  localhost_8888_block-editor_getting-started_devenv_docker-ubuntu_ (1) | localhost_8888_block-editor_getting-started_devenv_docker-ubuntu_ |

# Desktop

The follow rules are applied:
 -  Excluding the last item, truncate the total length of the crumb at `125px` by default and show an ellipsis.
 - Display a maximum of `4` crumbs
   - Truncate crumbs from left to right, starting at `i = 1` and `last - 3`

## _Unhiding_ the content on interaction

We want to maintain the ability to see the truncated text. As such, this PR currently displays the full text on hover & focus. 

### Hover (GIF)
Screen Capture on 2566-03-03 at 11-08-12

### Focus (GIF)
Screen Capture on 2566-03-03 at 11-08-32

### Problem

There is one problem with the current implementation; since it displays and displaces the content to the right, it sometimes forces the menu to drop down onto a new line. That is not a good experience.

### Example (GIF)
Screen Capture on 2566-03-03 at 11-19-56

**Options to fix:**
- When we display it, overlay it on top as a tooltip so the underlying menu doesn't move.
- Hide all the items to the right when interacting with a specific crumb
- Use JS to pull the container itself out of the flow temporarily so that when it grows it doesn't affect the menu on the right
- Use JS to truncate other texts to maintain a similar total width

I'm thinking the first option is probably the best one. Does anyone have any thoughts here? 

@ryelle @renintw @adamwoodnz @WordPress/meta-design

### DIFF
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -4,7 +4,7 @@
 	& span {
 		display: inline-block;
 
-		&.hidden {
+		&.is-hidden {
 			display: none;
 		}
 	}
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -76,3 +76,7 @@
 		width: auto;
 	}
 }
+
+html.no-js .wp-block-wporg-site-breadcrumbs span {
+	visibility: visible;
+}
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -1,6 +1,12 @@
 .wp-block-wporg-site-breadcrumbs {
+	white-space: nowrap;
+
 	& span {
 		display: inline-block;
+
+		&.hidden {
+			display: none;
+		}
 	}
 
 	& a {
@@ -10,60 +16,53 @@
 			text-decoration-line: underline;
 		}
 
-		@media (min-width: 800px) {
-
-			/* Show full crumb on interaction. */
-			&.truncated {
-				& span {
-					position: absolute;
-					opacity: 0;
-				}
-
-				&::before {
-					content: "...";
-				}
-
-				&:focus,
-				&:hover {
-					& span {
-						position: relative;
-						opacity: 1;
-					}
-
-					&::before {
-						display: none;
-					}
-				}
-			}
-
+		/* Show full crumb on interaction. */
+		&.truncated {
 			& span {
-				display: inline-block;
-				overflow-x: hidden;
-				white-space: nowrap;
-				max-width: 125px;
-				text-overflow: ellipsis;
-				vertical-align: middle;
+				position: absolute;
+				opacity: 0;
 			}
 
-			&:focus span,
-			&:hover span {
-				max-width: fit-content;
-				overflow-x: visible;
+			&::before {
+				content: "...";
 			}
+
+			&:focus::after,
+			&:hover::after {
+				background: var(--wp--preset--color--charcoal-1);
+				border-radius: 2px;
+				color: var(--wp--preset--color--white);
+				padding: 2px 4px;
+				position: absolute;
+				width: max-content;
+				left: -100%;
+				bottom: -40px;
+				font-size: var(--wp--preset--font-size--small);
+				content: attr(data-title);
+				z-index: 1000;
+			}
+		}
+
+		& span {
+			display: inline-block;
+			overflow-x: hidden;
+			white-space: nowrap;
+			text-overflow: ellipsis;
+			vertical-align: middle;
+		}
+
+		&:focus span,
+		&:hover span {
+			max-width: fit-content;
+			overflow-x: visible;
 		}
 	}
 
-	& > span:not(:last-child) {
-
-		/* This ensures that the square separator is reliably centered. */
-		margin-top: 0;
-
-		&::after {
-			content: "/";
-			display: inline-block;
-			font-weight: 400;
-			margin: 0 0.5rem;
-		}
+	& > span:not(:last-child)::after {
+		content: "/";
+		display: inline-block;
+		font-weight: 400;
+		margin: 0 0.5rem;
 	}
 
 	& .is-current-page {
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -33,6 +33,24 @@
 				}
 			}
 		}
+
+		&.partially-truncated {
+			& span {
+				display: inline-block;
+				overflow-x: hidden;
+				white-space: nowrap;
+				width: 100px;
+				text-overflow: ellipsis;
+				vertical-align: middle;
+				transition: width 400ms ease-in-out;
+			}
+
+			&:focus span,
+			&:hover span {
+				width: auto;
+				overflow-x: visible;
+			}
+		}
 	}
 
 	& > span:not(:last-child) {
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -17,7 +17,7 @@
 		}
 
 		/* Show full crumb on interaction. */
-		&.truncated {
+		&.is-truncated {
 			& span {
 				position: absolute;
 				opacity: 0;
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -4,9 +4,19 @@
 	& span {
 		display: inline-block;
 
+		/* Prevent visual flash on load */
+		&:not(:first-child) {
+			visibility: hidden;
+		}
+
 		&.is-hidden {
 			display: none;
 		}
+	}
+
+	/* After initial page load, show everything */
+	&.has-rendered span {
+		visibility: visible;
 	}
 
 	& a {
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -10,44 +10,44 @@
 			text-decoration-line: underline;
 		}
 
-		/* Show full crumb on interaction. */
-		&.truncated {
-			& span {
-				position: absolute;
-				opacity: 0;
-			}
+		@media (min-width: 800px) {
 
-			&::before {
-				content: "...";
-			}
-
-			&:focus,
-			&:hover {
+			/* Show full crumb on interaction. */
+			&.truncated {
 				& span {
-					position: relative;
-					opacity: 1;
+					position: absolute;
+					opacity: 0;
 				}
 
 				&::before {
-					display: none;
+					content: "...";
+				}
+
+				&:focus,
+				&:hover {
+					& span {
+						position: relative;
+						opacity: 1;
+					}
+
+					&::before {
+						display: none;
+					}
 				}
 			}
-		}
 
-		&.partially-truncated {
 			& span {
 				display: inline-block;
 				overflow-x: hidden;
 				white-space: nowrap;
-				width: 100px;
+				max-width: 125px;
 				text-overflow: ellipsis;
 				vertical-align: middle;
-				transition: width 400ms ease-in-out;
 			}
 
 			&:focus span,
 			&:hover span {
-				width: auto;
+				max-width: fit-content;
 				overflow-x: visible;
 			}
 		}
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -10,6 +10,7 @@
 	}
 
 	& a {
+		position: relative;
 		color: var(--wp--preset--color--charcoal-1);
 
 		&:hover {
@@ -35,7 +36,7 @@
 				padding: 2px 4px;
 				position: absolute;
 				width: max-content;
-				left: -100%;
+				left: 0;
 				bottom: -40px;
 				font-size: var(--wp--preset--font-size--small);
 				content: attr(data-title);
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -55,6 +55,10 @@
 
 	& .is-current-page {
 		font-weight: 700;
+
+		overflow-x: hidden;
+		text-overflow: ellipsis;
+		vertical-align: top;
 	}
 
 	/* If this is set to 1px, we can't calculate the elements natural width. */
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -56,4 +56,9 @@
 	& .is-current-page {
 		font-weight: 700;
 	}
+
+	/* If this is set to 1px, we can't calculate the elements natural width. */
+	& .screen-reader-text {
+		width: auto;
+	}
 }
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -1,6 +1,7 @@
 .wp-block-wporg-site-breadcrumbs {
-	display: flex;
-	align-items: center;
+	& span {
+		display: inline-block;
+	}
 
 	& a {
 		color: var(--wp--preset--color--charcoal-1);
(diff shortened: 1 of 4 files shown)


jasmussen:
Seems good to me. Another option is that the ... always links to the parent of the last shown breadcrumb. I.e. in the last GIF example shown, the ... links to "Block Editor Handbook" which is the parent of "How-to Guides". If we had deeper nesting, it would still just be the most immediate parent. The idea being: when you go up that high a level, on the next page you visit, you'll see further ancestors unhidden. So it would be the same amount of clicks.

StevenDufresne:
> Seems good to me. Another option is that the ... always links to the parent of the last shown breadcrumb. I.e. in the last GIF example shown, the ... links to "Block Editor Handbook" which is the parent of "How-to Guides". If we had deeper nesting, it would still just be the most immediate parent. The idea being: when you go up that high a level, on the next page you visit, you'll see further ancestors unhidden. So it would be the same amount of clicks.

Confirming I understand the idea, we omit parents as we go deeper?

**Examples**
- Given: `Home / Level 1 / Level 2 / Level 3 / Current`
  - Show `Home / Level 3 / Current`   
- Given: `Home / Level 2 / Level 3 / Level 4 / Current`    
  - Show `Home / Level 4 / Current`

I think that does work technically, and for me would make sense, however, I'm personally not familiar with how users are using the breadcrumbs on the developer & documentation sub-sites. Since we also have the Chapters sidebar, the breadcrumb probably doesn't get much use because it's further away from the content, its main use is to help users get back "home".... 🤔 

I'm going to try the tooltip approach and if we don't like it, we can reassess.

StevenDufresne:
I played with long breadcrumbs and omitting/truncating them and I honestly can't figure out a way to logically do this without ending up with __a lot__ of truncated items or removing crumbs:

For example, with a breadcrumb like the following:

`Developer / Block Editor Handbook / Getting Started / Development Environment / How to setup local WordPress environment on Ubuntu`

The only way to get things on the same line is to essentially truncate all the things, but even that will still fail before we hit our mobile breakpoint. Even if we omitted the parents, it will still break before the right-side menu collapses.

| Img |
| --- |
|   |

We could consider shrinking the text size, but since the menu is also in the same visual container, we'll have to reduce that font size too.


My instincts are telling me that a breadcrumb should not be sharing a container with another variable-width element (the menu), especially with how unpredictable our crumbs are. I know that's an opinion and runs counter to the redesign (documentation already uses this breadcrumb).


My suggestion:
 - Add the chapters sidebar to Documentation
 - Remove the breadcrumb from the documentation and developer handbooks

As I [mention above,](https://github.com/WordPress/wporg-mu-plugins/pull/348#issuecomment-1455277231) I think the breadcrumb trail is redundant in the handbooks with the presence of the chapters sidebar. They both do the same thing. We would still use the breadcrumbs for more controlled nesting (like [viewing a function](https://developer.wordpress.org/redesign-test/reference/functions/absint/)).

StevenDufresne:
@StevenDufresne
> Is this feature [chapters], not available on mobile?

@javierarce 
>  Sorry, I didn't explain this more explicitly elsewhere, but I think we could hide it there, because in that case people can use the breadcrumb navigation. 
>
>Here's the idea:
>
>If a user is reading the article on "HTML Coding Standards " (at https://developer.wordpress.org/coding->standards/wordpress-coding-standards/html) and wants to learn about "PHP Documentation Standards" (which is an >article under   "Inline Documentation Standards") they could use the breadcrumb navigation to reach the parent article (
>https://developer.wordpress.org/coding-standards/) and visit that link that is included in the content of the page.
>
>That said, we could explore how to integrate the chapter navigation on mobile if you think this solution is not right.

@javierarce 
>As [ryelle] pointed out in another comment, we can't currently >show the breadcrumb on mobile inside the local navigation >bar. So I think we have some options to solve this:
>
>a) hide the chapters on mobile (users will need to use the search function or go to the Developer homepage and then >browse through the content) 
>b) display the chapter navigation inside the article (I would need help from you to know the possible locations for the list)
>c) hide the chapter + show an independent breadcrumb navigation above the title that would only appear on mobile

Surfacing an incomplete discussion from the [developer redesign figma](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp?node-id=1246:11634#339218099).

jasmussen:
Not that this is the most important thing in the world, seems fine to hide and revisit if need be. Just for my own education, let's say we have the following full range of breadcrumbs:

`a > b > c > d > e > f > g`

Now let's say we decide for truncation purposes to truncate everything but the top level and the two bottom level items, i.e. keep only `a`, `f`, and `g`.

Can we do the following?

`a > ... > f > g`

instead of

`a > ... > ... > ... > e > f > g`

?

StevenDufresne:
> Can we do the following?
> 
> `a > ... > f > g`
> 
> instead of
> 
> `a > ... > ... > ... > e > f > g`
> 
> ?

Yep, we can and it will work for roughly `80%` of the crumbs. However, for the remaining `20%`, titles are still too long to fit within the submenu with the `4` navigational items on the right like in the developer theme.

**Some Examples:**

[Home](https://developer.wordpress.org/) / ... / [How to use JavaScript with the Block Editor](https://developer.wordpress.org/block-editor/how-to-guides/javascript/) / JavaScript Versions and Build Step

[Home](https://developer.wordpress.org/) / ... / [How to use JavaScript with the Block Editor](https://developer.wordpress.org/block-editor/how-to-guides/javascript/) / Extending the Block Editor

[Home](https://developer.wordpress.org/) / ... / [Create your First App with Gutenberg Data](https://developer.wordpress.org/block-editor/how-to-guides/data-basics/) / Part 4: Building a Create page form

[Home](https://developer.wordpress.org/) / ... / [Create a Block Tutorial](https://developer.wordpress.org/block-editor/getting-started/create-block/) / Share your Block with the World

[Home](https://developer.wordpress.org/) / ... / [Development Platform](https://developer.wordpress.org/block-editor/how-to-guides/platform/) / Building a custom block editor

StevenDufresne:
I played with collapsing everything. I think we can do for that now.

Screen Capture on 2566-03-07 at 15-12-45

Screen Capture on 2566-03-07 at 15-16-50

mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
```diff
@@ -1,30 +1,70 @@
...
 		color: var(--wp--preset--color--charcoal-1);
 
 		&:hover {
 			text-decoration-line: underline;
 		}
-	}
 
-	& > span:not(:first-child) {
+		/* Show full crumb on interaction. */
+		&.truncated {
```
ryelle: This is nitpicky, but I try to follow [the core naming standards](https://developer.wordpress.org/block-editor/contributors/code/coding-guidelines/#naming) which says that "state" class names should use an `is-` or `has-` prefix. I'm not sure if we've adopted it as a team, so you can take it or leave it 🙂

StevenDufresne: No, I appreciate that someone takes the time to call this stuff out. It's important. 

Feel no need to preface feedback with the word `nitpicky`. :)

StevenDufresne:
# Current State

## How it works
As the window shrinks, we calculate the space between the breadcrumb container and the navigation container. Once the containers are closer than a predefined value, `50`, we start by truncating items from left to right, making them visually look like `...`. Once we have truncated all the paths, excluding the home path and the current path, we then hide them. As the window expands, we do everything in reverse. 

This currently **does not** truncate the current path, which itself is sometimes too long. I'm not sure if that needs to be in this PR.

### Assumptions

#### HTML Structure

In order for this to work, the site breadcrumbs and navigation need to share a parent container and not flex 100% if broken onto 2 lines:

**IE:**
```
<div>
    <Sitebreadcrumbs />
    <Navigation />
</div>
```

This is necessary because we need to calculate the remaining space between them. This is problematic because our subnav code in most themes doesn't follow this format. To use this PR, we'll need to create a consistent wrapper block that accepts the site breadcrumbs and navigation as the only children. This is also important because in order to calculate the space remaining in the container, we'll need the elements to be `display: inline-block` otherwise, if they ever break onto separate lines the calculation will fail if the `display` is `block` or `flex` because the containers would fill the parent as stacked elements. We would calculate `0` remaining space. Having a dedicated component will allow us to control the CSS. I'll open up a PR for that.

#### Performance

The script makes calls to `getBoundingClientRect`, causing layout/reflow. This isn't ideal but I believe works well enough for the first iteration because we only calculate on `resize` which in normal conditions shouldn't be called often.
We could calculate it based on stored values, and should later, but right now that seems secondary to more readable code. Additionally, we could throttle the `resize` calls for even more performance improvements. However, I'm not sure they need to be addressed in this PR. I'm open to other opinions!

See my [last comment](https://github.com/WordPress/wporg-mu-plugins/pull/348#issuecomment-1457607684) for GIFs of it in action.

#### Flash on load
Once we merge https://github.com/WordPress/wporg-mu-plugins/pull/349, or something similar, I'll add some CSS here to make the experience nicer.

@renintw @adamwoodnz @ryelle Does this approach make sense? In terms of testing this, we'll need the wrapper component to make it work `100%`. You can test how it's roughly working by adding it to a theme. Just make sure that the site-breadcrumb and navigation share a parent element. Until we inline-block the containers, you will see some breakpoints where they stack. You can refresh and they should fix correctly.


### Edit

> This is also important because in order to calculate the space remaining in the container, we'll need the elements to be `display: inline-block` otherwise, if they ever break onto separate lines the calculation will fail if the `display` is `block` or `flex` because the containers would fill the parent as stacked elements. We would calculate `0` remaining space. Having a dedicated component will allow us to control the CSS

This assumption may be incorrect.

StevenDufresne:
I found time to improve this code and add in:
- Truncating last part
- Throttling resize events

I also combined this with a nav wrapper in https://github.com/WordPress/wporg-mu-plugins/tree/try/crumbs-with-nav. I would appreciate feedback before I do any other work. Part of me feels like this is getting overengineered :)

mu-plugins/blocks/site-breadcrumbs/src/view.js
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @return {number} The available space in pixels
+ */
+const getAvailableSpace = ( parent ) => {
+	const parentWidth = parent.getBoundingClientRect().width;
+	let occupiedSpace = 0;
+
+	for ( const child of parent.children ) {
+		occupiedSpace += child.getBoundingClientRect().width;
+	}
```
adamwoodnz, line 14: Could use reduce for this
```suggestion
	const occupiedSpace = parent.children.reduce(
		( width, child ) => width + child.getBoundingClientRect().width,
		0
	);
```

mu-plugins/blocks/site-breadcrumbs/src/view.js
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * This function collapses the breadcrumbs until the available space is greater than the breakpoint.
+ *
+ * @param {Array<HTMLElement>} arr        List of breadcrumb elements.
+ * @param {HTMLElement}        container  Breadcrumb parent container
+ * @param {number}             breakpoint The breakpoint in pixels.
+ *
+ * @return {void}
+ */
+const collapseCrumbs = ( arr, container, breakpoint ) => {
```
adamwoodnz: Name the array for what it is?
```suggestion
const collapseCrumbs = ( breadcrumbs, container, breakpoint ) => {
```

mu-plugins/blocks/site-breadcrumbs/src/view.js
```diff
@@ -0,0 +1,191 @@
...
+		if ( spanWidth >= pixelToAllocate ) {
+			return;
+		}
+
+		anchorElement.classList.remove( 'is-truncated' );
+		anchorElement.firstChild.classList.remove( 'screen-reader-text' );
+
+		pixelToAllocate -= spanWidth;
+	}
+};
```
adamwoodnz, line 147: This is a little DRYer and imo easier to read

```suggestion
	const crumbsRightToLeft = middleCrumbs.slice().reverse();
	crumbsRightToLeft.forEach( ( { firstChild, classList } ) => {
		// We don't need to do anything if the element is already expanded.
		if ( ! classList.contains( 'is-truncated' ) ) {
			continue;
		}

		const spanWidth = firstChild.getBoundingClientRect().width;

		// If this item can't fit, wait until it can.
		if ( spanWidth >= pixelToAllocate ) {
			return;
		}

		classList.remove( 'is-truncated' );
		firstChild.classList.remove( 'screen-reader-text' );

		pixelToAllocate -= spanWidth;
	}
};
```

mu-plugins/blocks/site-breadcrumbs/src/view.js
```diff
@@ -0,0 +1,191 @@
...
+	if ( hiddenEls.length ) {
+		hiddenEls[ 0 ].classList.remove( 'is-hidden' );
+		return;
+	}
+
+	/**
+	 * Loop through right to left, expand
+	 */
+	for ( let i = middleCrumbs.length - 1; i >= 0; i-- ) {
+		const anchorElement = middleCrumbs[ i ].firstChild;
```
adamwoodnz, line 128: IMO this is easier to read/understand. `slice()` creates a copy, as `reverse()` modifies the original array.

```suggestion
	const crumbsRightToLeft = middleCrumbs.slice().reverse();
	crumbsRightToLeft.forEach( ( { firstChild: anchorElement  } ) => {
```

adamwoodnz:
Nice work, this is an elegant solution imo. Few suggestions on JS inline, otherwise the only thing for me is that the throttling makes it a bit laggy when resizing. I think the only real world use case is something like rotating a device, this is what it looks like on iPad Pro

Kapture 2023-03-17 at 13 03 48

I wonder if we want to throttle using requestAnimationFrame instead of waiting with setTimeout?

StevenDufresne:
> Nice work, this is an elegant solution imo. Few suggestions on JS inline, otherwise the only thing for me is that the throttling makes it a bit laggy when resizing. I think the only real world use case is something like rotating a device, this is what it looks like on iPad Pro

Thanks for testing. I updated the delay from `200` to `50` to reduce that lag you are mentioning.

**Updated UX**


> I wonder if we want to throttle using requestAnimationFrame instead of waiting with setTimeout?

Coming back to this, I misused the word `throttle` to represent `debouncing`. Since the function is only being called once, I'm not sure I understand how using RAF to debounce would be an improvement.

adamwoodnz:
> Since the function is only being called once

On a single resize event like device rotation yes.

It will fire many times if the window is being resized manually though, ie. dragging to resize. It would be a nice-to-have to handle that gracefully so that the breadcrumb layout updates fluidly as the window width changes.

mu-plugins/blocks/site-breadcrumbs/src/view.js
StevenDufresne: I feel like the fact that you needed to explain how `slice` and `reverse` works, maybe suggests it isn't otherwise easier to understand? :) 

I don't feel strongly either way.

mu-plugins/blocks/site-breadcrumbs/src/view.js
StevenDufresne: I definitely like `.reduce` when a functions are passed in. For simple operations like this, I don't feel strongly.

StevenDufresne:
> It will fire many times if the window is being resized manually though, ie. dragging to resize. It would be a nice-to-have to handle that gracefully so that the breadcrumb layout updates fluidly as the window width changes.

Yeah that's fair, if the window is resized slowly, it will call multiple times. We could be a bit more careful. I would probably consider that a nice-to-have at this point.

mu-plugins/blocks/site-breadcrumbs/src/view.js
adamwoodnz: Yeah, looking at this again, creating a copy probably isn't the most performant suggestion.

mu-plugins/blocks/site-breadcrumbs/src/view.js
adamwoodnz: IMO it's almost always the most elegant way to take many things (`children`) and output 1 thing (`width`); no mutated `let`, everything encapsulated.

But I've come from codebases with a lot of functional patterns, and I don't feel strongly either 😂

adamwoodnz:
> I updated the delay from `200` to `50` to reduce that lag you are mentioning.

I'll test this again first thing next week. I'm sure you want to get this wrapped up :)

adamwoodnz (commented):
Tested again and lag is much better, hidden span works with JS, but does the `html.no-js` override depend on [this PR](https://github.com/WordPress/wporg-mu-plugins/pull/349#issuecomment-1536025156) which has had to be reverted?

StevenDufresne:
> Tested again and lag is much better, hidden span works with JS, but does the `html.no-js` override depend on [this PR](https://github.com/WordPress/wporg-mu-plugins/pull/349#issuecomment-1536025156) which has had to be reverted?

Yes.

I'm also going to wait until https://github.com/WordPress/wporg-mu-plugins/issues/465 is resolved.

StevenDufresne:
I think this will not be merged. I'll switch it to draft in the meantime.

adamwoodnz:
I think this is still going to be an issue on Developer, even without the content on the right

StevenDufresne:
True, but I think you can just stack them instead of complex truncating logic.

PR #349, opened 2023-03-07
Add class to html to detect js.

StevenDufresne:
There are a couple of components (chapters, breadcrumb ) that have a flash due to javascript running post initial paint. This PR adds `no-js` to the HTML element and removes it using `js` so we can set up styling to stop the _jank_.

### Use case
The breadcrumbs flash before we are able to calculate how much we need to truncate them. In an only `js` world, we could add a class that obscures them and remove it once we are done calculating. However, if the browser isn't running js, that class would never be removed and the user would never see the breadcrumbs. With the `no-js` class, we can do something like the following:

```
.breadcrumbs {
     opacity: 0;
}

.no-js .breadcrumbs {
   opacity: 1;
}
```

Note: That's just a basic example and probably not the actual solution that would be used.

### DIFF
--- a/mu-plugins/no-js/index.php
+++ b/mu-plugins/no-js/index.php
@@ -7,8 +7,8 @@
 namespace WordPressdotorg\MU_Plugins\NO_JS;
 
 // Actions & Filters
-add_filter( 'language_attributes', __NAMESPACE__ . '\add_no_js_class_to_html_tag', 10, 2 );
-add_action( 'wp_head', __NAMESPACE__ . '\add_custom_js' );
+add_filter( 'language_attributes', __NAMESPACE__ . '\add_no_js_tag', 10, 2 );
+add_action( 'wp_head', __NAMESPACE__ . '\remove_js_tag' );
 
 /**
  * Add a 'no-js' tag by default.
--- a/mu-plugins/no-js/index.php
+++ b/mu-plugins/no-js/index.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * Plugin Name: No-JS
+ * Description: This adds a 'no-js' class and removes it with JS for styling.
+ */
+
+namespace WordPressdotorg\MU_Plugins\NO_JS;
+
+// Actions & Filters
+add_filter( 'language_attributes', __NAMESPACE__ . '\add_no_js_class_to_html_tag', 10, 2 );
+add_action( 'wp_head', __NAMESPACE__ . '\add_custom_js' );
+
+/**
+ * Add a 'no-js' tag by default.
+ *
+ * @param string $output
+ * @param string $doctype
+ * @return string
+ */
+function add_no_js_tag( $output, $doctype ) {
+	if ( 'html' !== $doctype ) {
+		return $output;
+	}
+
+	$output .= ' class="no-js"';
+
+	return $output;
+}
+
+/**
+ * Remove the 'no-js' class from html using JS.
+ */
+function remove_js_tag() {
+	?>
+	<script>
+		document.documentElement.classList.remove('no-js')
+	</script>
+	<?php
+}
--- a/mu-plugins/no-js/index.php
+++ b/mu-plugins/no-js/index.php
@@ -4,7 +4,7 @@
  * Description: This adds a 'no-js' class and removes it with JS for styling.
  */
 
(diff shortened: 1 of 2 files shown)


ryelle:
Is there a reason to put the class on `html` vs `body`? If you put it on `body`, you could use the [`body_class` hook](https://developer.wordpress.org/reference/hooks/body_class/) and avoid the workaround you're doing in `language_attributes`. I guess you'd need to move the `script` to the end of the page instead, in that case.

StevenDufresne:
> Is there a reason to put the class on `html` vs `body`? If you put it on `body`, you could use the [`body_class` hook](https://developer.wordpress.org/reference/hooks/body_class/) and avoid the workaround you're doing in `language_attributes`. I guess you'd need to move the `script` to the end of the page instead, in that case.

Yeah, if it's moved to the end of the page, we are more likely to get the flicker.

mu-plugins/no-js/index.php
```diff
@@ -0,0 +1,39 @@
+<?php
+/**
+ * Plugin Name: No-JS
+ * Description: This adds a 'no-js' class and removes it with JS for styling.
+ */
+
+namespace WordPressdotorg\MU_Plugins\NO_JS;
```
ryelle:
```suggestion
namespace WordPressdotorg\MU_Plugins\No_JS;
```

ryelle (approved):
Tested on my sandbox on classic and new sites, and in the API endpoint that renders the header for trac etc — this sets the class & removes it very quickly when JS is supported.

I did a quick search for `language_attributes` in the wporg codebase and I found that the [plugin](https://github.com/WordPress/wordpress.org/blob/1bcbb531c8c4f22e0262edb5640ea89ac17fa727/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/embed-plugin.php#L21) and [theme](https://github.com/WordPress/wordpress.org/blob/1bcbb531c8c4f22e0262edb5640ea89ac17fa727/wordpress.org/public_html/wp-content/themes/pub/wporg-themes/embed-repopackage.php#L14) embed pages also add a `no-js` class to `html`, so this creates a duplicate class attribute (test on your sandbox & go to https://wordpress.org/plugins/akismet/embed/).

I think that's fine though, it seems to be the exception. As far as I can tell, the classic sites use the header from `global-header-footer/classic-header.php`, which only sets `<html <?php language_attributes(); ?>>`, and the block sites use the code from [`template-canvas.php`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/template-canvas.php) which also only sets `<html <?php language_attributes(); ?>>`.

StevenDufresne:
> Tested on my sandbox on classic and new sites, and in the API endpoint that renders the header for trac etc — this sets the class & removes it very quickly when JS is supported.
> 
> I did a quick search for `language_attributes` in the wporg codebase and I found that the [plugin](https://github.com/WordPress/wordpress.org/blob/1bcbb531c8c4f22e0262edb5640ea89ac17fa727/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/embed-plugin.php#L21) and [theme](https://github.com/WordPress/wordpress.org/blob/1bcbb531c8c4f22e0262edb5640ea89ac17fa727/wordpress.org/public_html/wp-content/themes/pub/wporg-themes/embed-repopackage.php#L14) embed pages also add a `no-js` class to `html`, so this creates a duplicate class attribute (test on your sandbox & go to https://wordpress.org/plugins/akismet/embed/).
> 
> I think that's fine though, it seems to be the exception. As far as I can tell, the classic sites use the header from `global-header-footer/classic-header.php`, which only sets `<html <?php language_attributes(); ?>>`, and the block sites use the code from [`template-canvas.php`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/template-canvas.php) which also only sets `<html <?php language_attributes(); ?>>`.

Thanks for the in-depth review. Do you have any hesitation in merging this?

ryelle:
nope, looks good to me 👍🏻

pkevan:
This was reverted on production due to a bug in the password recovery screen, whereby the password field was hidden.

StevenDufresne:
Thanks for the revert @pkevan. I'll reassess this approach. 👍

PR #350, opened 2023-03-13
Header & Footer: Use core supports for text, background, and border controls

### LINKED ISSUE #338: Border colors and behaviour in WordPress.org
A discussion came up on [this issue](https://github.com/WordPress/wporg-main-2022/issues/199) regarding the header and when it should have a border vs when it shouldn't. The thinking would be, whenever the header background is the same color as the cover background, there should be a border to act as a visual mark.

**The example for the header in the Hosting page:**

Captura de ecrã 2023-02-14, às 15 46 06

**Same for the footer.**

Captura de ecrã 2023-02-14, às 15 46 57

**Another example with the header and cover in black as well, there should be a white border with opacity between them.**

Captura de ecrã 2023-02-14, às 15 44 56

**Here's an example with coloured backgrounds:**

Captura de ecrã 2023-02-14, às 15 45 41

Captura de ecrã 2023-02-14, às 15 53 31

There may be exceptions depending on the concept of the design (such as [About](https://wordpress.org/about/)) where separators don't have opacity, or thumbnails that could require a bit more opacity, but after a few tests I think 10% opacity for header and footer could work. What do you think @WordPress/meta-design @marko-srb?

### LINKED ISSUE #341: Enable color settings on the Footer block
See https://github.com/WordPress/wporg-developer/issues/213#issuecomment-1430579150

> > Hm... with the "expressive" sites, they'll probably introduce other footer variations with other colors. Do you think we should just keep adding variations, or maybe switch to enabling colors on the footer block, and set the background/text/link colors on the block instead?
>
> Yeah, that makes sense. Let's make the block configurable.

ryelle:
Fixes #338, fixes #341 — This enables the core supports for the text color, background color, and border controls on Global Header and Global Footer, replacing the existing "variations" approach for styling. Rather than continuing to add variations for "expressive" sites, we can allow configuration and tweak the CSS as needed.

**Screenshots**

The first three screenshots match the current behavior.

| Variation | Screenshot |
|-----|-----|
| White on black (default) | default |
| Black on white | black-on-white |
| White on blue | white-on-blue |
| With a border | with-borders |
| Custom background | custom-color |

In the editor you can see the new UI

in-editor

Custom colors are supported on the header & footer. The footer handles this fine, since there are only two colors involved. Technically you can change the background color of the header to any color too, but the hover states & dropdowns are not smart enough to match arbitrary colors. Currently I've got this falling back to the grey dropdown when the text is a `charcoal`. But even the blue "current" state clashes. If/when we want new colors for the header, we'll need to add some extra CSS.

custom-color-header

**To test**

You technically can try this in the site editor, but the blocks themselves are a little wonky. It's easier to test by editing the block in a template file, for example, try any of the following (swap out `header`/`footer`, different colors, etc):

Default
```html
<!-- wp:wporg/global-header /-->
```

Black on white
```html
<!-- wp:wporg/global-header {"backgroundColor":"white","textColor":"charcoal-2"} /-->
```

Black on a custom color
```html
<!-- wp:wporg/global-footer {"textColor":"charcoal-2","style":{"color":{"background":"#6ec9ae"}}} /-->
```

Black on white with a border
```html
<!-- wp:wporg/global-header {"textColor":"charcoal-2","backgroundColor":"white","style":{"border":{"bottom":{"color":"#0000001a","width":"1px","style":"solid"}}}} /-->
```

White on blue (using `style` syntax to ensure back-compat for the sites that are currently using it)
```html
<!-- wp:wporg/global-header {"style":"white-on-blue"} /-->
```

❗  Reviewing note: The changes in `header.php` and `footer.php` look massive, but it's mostly indentation changes because the wrapper element has been moved up to `blocks.php`.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -87,6 +87,10 @@ html {
 	--wp-global-header--link-color: var(--wp--preset--color--charcoal-2);
 	--wp-global-header--link-color--active: var(--wp--preset--color--deep-blueberry);
 	--wp-global-header--text-color: var(--wp--preset--color--charcoal-2);
+
+	&.has-background[style*="background-color:"] {
+		--wp-global-header--background-color--hover: var(--wp--preset--color--light-grey-2);
+	}
 }
 
 .has-blueberry-1-background-color {
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -81,6 +81,8 @@ html {
 	--wp-global-header--text-color: var(--wp--preset--color--white);
 }
 
+.has-charcoal-0-color,
+.has-charcoal-1-color,
 .has-charcoal-2-color {
 	--wp-global-header--link-color: var(--wp--preset--color--charcoal-2);
 	--wp-global-header--link-color--active: var(--wp--preset--color--deep-blueberry);
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -377,7 +377,15 @@ function render_global_header( $attributes = array() ) {
 		$markup = ob_get_clean() . $markup;
 	}
 
-	return $markup;
+	$wrapper_attributes = get_block_wrapper_attributes(
+		array( 'class' => 'global-header wp-block-group' )
+	);
+	return sprintf(
+		'<header %1$s>%2$s</header>%3$s',
+		$wrapper_attributes,
+		$markup,
+		wp_kses_post( render_header_alert_banner() )
+	);
 }
 
 /**
@@ -822,7 +830,6 @@ function render_global_footer( $attributes, $content, $block ) {
 	$wrapper_attributes = get_block_wrapper_attributes(
 		array( 'class' => 'global-footer wp-block-group' )
 	);
-	// var_dump( $wrapper_attributes );
 	return sprintf(
 		'<footer %1$s>%2$s</footer>',
 		$wrapper_attributes,
@@ -840,7 +847,7 @@ function render_global_footer( $attributes, $content, $block ) {
 function update_block_style_colors( $block ) {
 	if (
 		! empty( $block['blockName'] ) &&
-		'wporg/global-footer' === $block['blockName'] &&
+		in_array( $block['blockName'], [ 'wporg/global-footer', 'wporg/global-header' ], true ) &&
 		! empty( $block['attrs']['style'] )
 	) {
 		if ( 'black-on-white' === $block['attrs']['style'] ) {
@@ -1062,7 +1069,7 @@ function swap_header_search_action( $block_content, $block ) {
 	return $block_content;
 }
 
-/*
+/**
  * Translate the tagline with the necessary text domain.
  */
 function get_cip_text() {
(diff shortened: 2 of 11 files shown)


StevenDufresne:
What about backward compatibility?

ryelle:
> What about backward compatibility?

Should be fine, the old style of `<!-- wp:wporg/global-header {"style":"white-on-blue"} /-->` will still work. [This function](https://github.com/WordPress/wporg-mu-plugins/blob/e8405c7d9c1656a97ba579c9d36d942146f075c9/mu-plugins/blocks/global-header-footer/blocks.php#L840-L863) converts the old `style` attribute to the new `textColor` & `backgroundColor`.

StevenDufresne (approved):
This works (provided your Gutenberg is up to date to support `skipBlockSupportAttributes`) and the code makes sense!

I'm not completely familiar with how the header/footer works in places like trac... it's probably worth checking that out before we deploy this.

ryelle:
> I'm not completely familiar with how the header/footer works in places like trac

It's synced from [an API endpoint](https://wordpress.org/wp-json/global-header-footer/v1/header/), but good catch, I'll do a bit more sandbox testing. It _should_ be fine since the defaults all still work.

ryelle:
Okay, works fine with the API endpoint, but it's actually wonky with the classic themes, since the full header output [is wrapped in `header` tags now](https://github.com/WordPress/wporg-mu-plugins/blob/e8405c7d9c1656a97ba579c9d36d942146f075c9/mu-plugins/blocks/global-header-footer/blocks.php#L369-L388), including the `<head>` code.

ryelle:
Okay, fixed that so now the opening & closing html tags & code are not wrapped in the header/footer tags on classic themes. Do you want to re-review @StevenDufresne?

mu-plugins/blocks/global-header-footer/postcss/_common.pcss
```diff
@@ -81,10 +81,16 @@ html {
...
 }
 
+.has-charcoal-0-color,
+.has-charcoal-1-color,
 .has-charcoal-2-color {
 	--wp-global-header--link-color: var(--wp--preset--color--charcoal-2);
 	--wp-global-header--link-color--active: var(--wp--preset--color--deep-blueberry);
 	--wp-global-header--text-color: var(--wp--preset--color--charcoal-2);
+
+	&.has-background[style*="background-color:"] {
```
outdoor2kode, line 91: I noticed that editing the global header background color in the editor can cause the class name to change, even if the color is the same. This would make the hover color reset here ineffective. 

Do you think adding something like `&[class*="-background-color"]` makes sense?
```suggestion
	&.has-background[style*="background-color:"],
	&[class*="-background-color"] {
```

outdoor2kode (commented):
Just caught up on the context and tested the code, and things look good 👍 

Except for the editor part that:
1. When I change the color, it doesn't change accordingly in the editor, but your screenshot shows that it does. 
2. Also, when I change the header style by modifying the template file, it only reflects on the front end, in the editor, it's still white-on-black.

Perhaps these are the wonky part that you mentioned.

StevenDufresne:
> Okay, fixed that so now the opening & closing html tags & code are not wrapped in the header/footer tags on classic themes. Do you want to re-review @StevenDufresne?

No, the changes are relatively minor. 👍

ryelle:
> Perhaps these are the wonky part that you mentioned.

Yeah, I should have been more clear 😅  — I'm not sure why it's not rendering the color in the editor anymore (it was when I took the screenshot in the description, but not now), but I think we can ignore that issue. This won't be edited in the editor anyway.

> I noticed that editing the global header background color in the editor can cause the class name to change, even if the color is the same. This would make the hover color reset here ineffective.

Can you explain, or share the block code where you're seeing this issue?

outdoor2kode:
> Can you explain, or share the block code where you're seeing this issue?

I first replaced the default globalHeader block with

`<!-- wp:wporg/global-header {"textColor":"charcoal-2","style":{"color":{"background":"#6ec9ae"}}} /-->`

in `front-page.html`, so the front-end was like:



and in the editor, it was like

Screen Shot 2023-03-16 at 11 54 34 PM

Its code was 

`<!-- wp:wporg/global-header {"style":{"color":{"background":"#6ec9ae"}},"textColor":"charcoal-2"} /-->`

If I switch to another color and then switch back, or if I unselect color `#6ec9ae` from the color palette and then select it again, the code would change to the following:

`<!-- wp:wporg/global-header {"backgroundColor":"aquamarine","textColor":"charcoal-2"} /-->`

And the front-end would be broken then.

Screen Shot 2023-03-17 at 12 00 10 AM

> This won't be edited in the editor anyway.

But since you mentioned that this wouldn't be edited in the editor, so I suppose it doesn't matter and can be skipped for now?

ryelle:
Oh, I see now. I addressed this in the description somewhat; while arbitrary background colors are somewhat supported for the header, if we actually want to implement a new color style, it will need design and custom CSS. I think that's out of scope for this PR, since we don't even know if that will be necessary (if you look at the current i3 in developer, it still uses a black global-header).

outdoor2kode:
I got it now, it's a situation that's uncertain in a couple of ways. Thank you for taking the extra time to explain it to me 🙂

PR #373, opened 2023-03-28
Navigation: Update focus style, Get WordPress button mobile style

### LINKED ISSUE #363: Navigation: Mobile overlay "Get WP" button details
Get WordPress is a standard blue button in the desktop navigation:

Screenshot 2023-03-27 at 12 45 34

In the mobile overlay, it becomes full width, and has a larger font size:

Screenshot 2023-03-27 at 12 45 24

Can we use the same style of button inside the overlay? I.e. same height, font size, perhaps left aligned?

### LINKED ISSUE #372: Navigation: Explore clearer focus style that doesn't look like an active section, and don't show on mobile

> (Mobile) News is permanently underlined, even when other instances are active.

Here are screenshots. Homepage:

Screenshot 2023-03-27 at 13 36 21

Download page:

Screenshot 2023-03-27 at 13 36 41

In both cases here, "News" is underlined. 

Technically this happens because the overlay menu is a modal dialog, which has to set focus on the first menu item. In other words, the underline represents the focus style. To rectify this, let's do two things:

**1. Don't show this focus style on a touchscreen device, where you can't use tab to navigate anyway**

We can very probably hide it using code like this:

```
@media (hover: none) {
	.classname:focus {
	   ...
	}
}
```

Let's separately explore a better focus style that can ideally be generic and the same across the entire website, and look more like a focus style. A quick mockup in the inspector:

Screenshot 2023-03-27 at 13 46 07

* The above reduces the padding of top level menu items, increases on the menu container instead
* It adds a pseudo element on focus, and inside that pseudo element draws a box-shadow.
* This is so the focus style can have a 2px rounded radius.
* Specific metrics of the shadow are: `box-shadow: 0 0 0 1.5px #7b90ff`.
* Can use light blueberry on dark backgrounds, regular blueberry on light backgrounds.

This focus style is the same inside the block editor.

ryelle:
Fixes #372, fixes #363. This updates the focus style across the header, and removes the special styling on the mobile "Get WordPress" button.

I'm using the "current item" color for the focus state, which is `--wp--preset--color--blueberry-2` on the black background, `--wp--preset--color--deep-blueberry` on the white background, and `--wp--preset--color--white` on the blue background.

| | Default | Blue bg | White bg |
|---|---|---|---|
| Top level item focus | branch-default-1-large-menu-focus | branch-blue-1-large-menu-focus | branch-white-1-large-menu-focus |
| Menu open, submenu item focus | branch-default-2-large-menu-open | branch-blue-2-large-menu-open | branch-white-2-large-menu-open |
| Get WordPress focus | branch-default-3-large-button-focus | branch-blue-3-large-button-focus | branch-white-3-large-button-focus |
| Small screen, logo focus | branch-default-4-small-logo-focus | branch-blue-4-small-logo-focus | branch-white-4-small-logo-focus |
| Small screen, menu open, top level item focus | branch-default-5-small-menu-top-level | branch-blue-5-small-menu-top-level | branch-white-5-small-menu-top-level |
| Small screen, menu open, submenu item focus | branch-default-6-small-menu-2nd-level | branch-blue-6-small-menu-2nd-level | branch-white-6-small-menu-2nd-level |

The spacing around the small screen menu items have also been tweaked, but I've kept the highlight at the edge of the screen

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
@@ -25,8 +25,9 @@
 		&:focus,
 		&:focus-visible,
 		&:focus-within {
-			outline: 1px dashed currentcolor;
-			outline-offset: -2px;
+			outline: none;
+			border-radius: 2px;
+			box-shadow: 0 0 0 1.5px var(--wp-global-header--link-color--active);
 		}
 	}
 
--- a/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/get-wordpress.pcss
@@ -3,14 +3,14 @@
  */
 
 .wp-block-group.global-header {
-
-	/* This is the `a` on desktop, and the containing `li` on mobile. */
-	& .global-header__get-wordpress {
+	& .global-header__desktop-get-wordpress,
+	& .global-header__mobile-get-wordpress a {
 		background-color: var(--wp--preset--color--blueberry-1) !important; /* Override Gutenberg's !important */
 		color: var(--wp--preset--color--white) !important; /* Override the header color scheme */
 		padding: 10px 19px;
 		border-radius: 2px;
 		font-weight: 700;
+		font-size: var(--wp--preset--font-size--small);
 
 		&:hover {
 			cursor: pointer;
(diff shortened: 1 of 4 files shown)


StevenDufresne:
I don't think i see another issue for this, but can we address the fact that the focus indicator is cut off and not aligned on the bottom (off by a couple pixels?):

Screenshot 2023-03-29 at 10 31 26 AM

Seems to be related to a border bottom on the `li`.

I would also prefer if the outline was closer to the text, but that could probably be done later.

jasmussen:
Thanks for the PR, and excellent catch Steve! The box shadow needs to be inset, that also corrects the border radius. So instead of:

> box-shadow: 0 0 0 1.5px var(--wp-global-header--link-color--active);

it should be 

> box-shadow: inset 0 0 0 1.5px var(--wp-global-header--link-color--active);

That should fix the crop.

ryelle:
I've switched the box-shadow to inset, so it's not cut off anymore:

Screenshot 2023-03-29 at 4 54 41 PM

Will merge this now to keep things moving, but if we need to iterate more we can.

PR #374, opened 2023-03-28
Footer: Align "wordpress.org" baseline with "code is poetry"

### LINKED ISSUE #362: Footer: Align logo with social links vertically
Currently, the logo in the footer is not correctly aligned:

Screenshot 2023-03-28 at 11 58 44

It should be nudged upwards to be more like:

Screenshot 2023-03-28 at 12 43 01

This also makes sure code is poetry is on the baseline as the "WordPress.org" word. The above inspector change:

* adds a -2px top margin to the logo
* sets the social icons to be 28x28 instead of 29x29
* sets the top position of the Code is Poetry image to be 6px instead of the about 8.666 that it is now

iandunn:
Fixes #362

Screenshot 2023-03-28 at 3 49 53 PM

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
@@ -48,6 +48,8 @@
 
 		@media (--tablet) {
 			display: inline;
+			position: relative;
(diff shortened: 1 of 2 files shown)


jasmussen:
Thanks for the screenshot, I haven't been able to set up the dev env to test. But eyeballing it, this ooks correct, thank you!

PR #375, opened 2023-03-28
Header: Make search icon consistent between desktop and mobile

### LINKED ISSUE #369: Search on mobile: Align icons
On desktop, the Search and Close icons look correct, and align properly:

desktop search

On mobile, this, the search icon is misaligned, and appears too big:

mobile search

Here's a visual of the metrics of the search icon on desktop:

Screenshot 2023-03-27 at 11 01 27

Here's a visual of the metrics of the search icon on mobile:

Screenshot 2023-03-27 at 11 01 35

Visible when comparing those two, it seems like the search icon (which appears to be used as a background image on mobile instead of the inline SVG) appears larger. It should remain 24x24.

iandunn:
Fixes #369

This keeps the desktop styles, and makes the mobile icon smaller and center-positioned. I think the size is correct, but I'm not sure what the precise alignment should be. Let me know if you'd like changes.

Screenshot 2023-03-28 at 4 22 10 PM

### DIFF
--- a/mu-plugins/blocks/global-header-footer/images/search-for-light-bg.svg
+++ b/mu-plugins/blocks/global-header-footer/images/search-for-light-bg.svg
@@ -1,4 +1,6 @@
-<svg width="20" height="19" viewBox="0 0 20 19" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M1.5 17.5001L7.16667 12.5417" stroke="#1e1e1e" stroke-width="2"/>
-<path d="M18.9168 8.29167C18.9168 12.0426 15.8761 15.0833 12.1252 15.0833C8.37423 15.0833 5.3335 12.0426 5.3335 8.29167C5.3335 4.54073 8.37423 1.5 12.1252 1.5C15.8761 1.5 18.9168 4.54073 18.9168 8.29167Z" stroke="#1e1e1e" stroke-width="2"/>
+<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" style="enable-background:new 0 0 24 24" viewBox="0 0 24 24">
+	<path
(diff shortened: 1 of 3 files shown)


jasmussen:
Thanks for the PR, it certainly looks better. I delved into the inspector more, and it appears that the SVG is used as a background image, and comes with a somewhat nonstandard size (20x19) compared to the [new WordPress icon library](https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library) (24x24). The menu icon is correct, though. 

Here's a comparison. The two right most are from the library:

Frame 3

Here it is, zoomed in:

Screenshot 2023-03-29 at 09 10 58

So this needs more work on my end:

* The core icon is the one we should use, it has the correct 24x24 footprint and uses the right stroke width.
* However it's not optically centered horizontally.

jasmussen:
Upstream fix: https://github.com/WordPress/gutenberg/pull/49425

jasmussen:
There's a little discussion in the upstream fix, ~but I think it might result in this SVG~ it did result in this SVG:

```
<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" style="enable-background:new 0 0 24 24" viewBox="0 0 24 24"><path d="M13 5c-3.3 0-6 2.7-6 6 0 1.4.5 2.7 1.3 3.7l-3.8 3.8 1.1 1.1 3.8-3.8c1 .8 2.3 1.3 3.7 1.3 3.3 0 6-2.7 6-6S16.3 5 13 5zm0 10.5c-2.5 0-4.5-2-4.5-4.5s2-4.5 4.5-4.5 4.5 2 4.5 4.5-2 4.5-4.5 4.5z" style="fill-rule:evenodd;clip-rule:evenodd;fill:#1e1e1e"/></svg>
```

iandunn:
Thanks for that, I've switched out the SVG and it's using the `24x24` footprint now:

Screenshot 2023-03-29 at 3 48 47 PM

Screenshot 2023-03-29 at 3 48 40 PM

The reason it's a background image is because we're using the Search block inside the Navigation block:

https://github.com/WordPress/wporg-mu-plugins/blob/b32eb9e1fe74396c0d2fca60c58f68e3ada247de/mu-plugins/blocks/global-header-footer/header.php#L96-L103

AFAIK we can't change the markup of the Search block to swap in a different image, so we used `background-image` instead. One downside is that we need to fork the Core image in order to set the `fill`, but that seems much simpler to me than the alternative. Do you see a compelling reason to change that?

jasmussen:
Looks good, go for it!

iandunn:
👍🏻

PR #378, opened 2023-03-30
Search: Update search styles to match parent theme

### LINKED ISSUE #368: Search: X to clear is out of place, and wrong icon
When a search field is focused, in Chrome and Safari (and probably other webkit browsers), there's an X to clear button. Chrome:

Screenshot 2023-03-27 at 11 06 45

Screenshot 2023-03-27 at 11 06 53

Safari:
Screenshot 2023-03-27 at 11 10 23

Screenshot 2023-03-27 at 11 10 35

It appears to be the `::-webkit-search-cancel-button` pseudo element. Its position and appearance isn't a good match for the rest of the interface. It seems we should:

* Style it to correctly position the X, and make it use an icon from the set
* Hide it if that's not possible.

Here's a mockup:

Mockup

* Charcoal 4 placeholder color
* Charcoal search term color
* 14px Inter, same as breadcrumb font size. 
* When on a background that has more than 3:0 contrast, don't show a border.
* Always 24x24px icons, never larger or smaller, even on mobile.
* 16px font size on mobile.

I expect search fields across the site to be updated to match. A separate linked issue will be opened for that.

ryelle:
This copies over the styles from the parent theme's `/sass/blocks/_search.scss`, expanding out mixins and converting to mostly-vanilla CSS. Since this can also be used on Classic sites that don't set the theme.json variables, I also had to set those `--wp--custom--form--*` properties.

See https://github.com/WordPress/wporg-main-2022/issues/228, Fixes #368.

**Screenshots**

| | Default | Blue | White | Classic |
|---|---|---|---|---|
| Desktop, search open | branch-default-1 | branch-blue-1 | branch-white-1 | branch-old-1 |
| Desktop, button focused | branch-default-2 | branch-blue-2 | branch-white-2 | branch-old-2 |
| Mobile, search open | branch-default-3 | branch-blue-3 | branch-white-3 | branch-old-3 |
| Mobile, button focused | branch-default-4 | branch-blue-4 | branch-white-4 | branch-old-4 |

~This fixes part of #368 by updating the button margin, but since it doesn't change the icon it's not a full fix.~ Icon has been added.


jasmussen:
Thanks for working on this. Thanks for differentiating between the light contexts, where the input field border just grows thicker, and the darker contexts. 

Can we try an inset shadow for the latter context? I.e. have both the white halo, and the blue focus style, grow inwards? This is most visible on the search field where input and button are separate. An alternative could potentially be to have the search button visually sit "inside" the search field, so the focus style for the input encompassed both.

jasmussen:
An alternative, and I regret this will be a bit more complex, potentially, to implement:

* By default, inputs on a light background, change the input border on focus to a 1.5px blue stroke 
* For inputs on a dark or black background, change the input border on focus to a 1.5px blue stroke — the blue should have more than 3:1 contrast ratio against both white and black.
* For inputs on a blue background, change the input border on focus to a 1.5px black stroke — which similarly should retain a 3:1 contrast ratio against both white and blue.

fcoveram:
I tested the search component with the three themes' backgrounds, and here is an idea for the focus ring

image

The idea covers the following:
* Input focus has a double ring of 1.5px on all backgrounds to ensure high contrast. On light backgrounds, the external ring is blueberry and the internal ring white. On dark backgrounds, the external ring is white and internal is blueberry.
* "Clear" and "Search" actions are 24px with a col-gap of 4px. The focus style is one outer ring of blueberry.
* The resting state of inputs matches the background color, except for light backgrounds. Here I reduced the background opacity to see the input's border color in dark backgrounds.

image

jasmussen:
Love it Francisco, especially the inner clear and search elements are much stronger here.

However I would omit the white halo on light backgrounds, it wouldn't be necessary there.

ryelle:
I like the design, and we can use that focus style in general, but we can't outline things the way it's mocked up. 

The X isn't a real button, so it doesn't receive keyboard focus at all.

For focus when in the search box, it can't easily be full width. If we used `focus-within` to highlight the full border, it would stay highlighted when the button was focused too. So for that, it needs to stay restricted to the input itself:
    Screenshot 2023-03-31 at 10 29 08 AM

Would shrinking the focus of the search button also shrink it's clickable area, or is that outline inset?

> The resting state of inputs matches the background color, except for light backgrounds. Here I reduced the background opacity to see the input's border color in dark backgrounds.

I'm not sure I understand this. We have two styles of search— one on dark backgrounds and one on light backgrounds, so we can have two default border colors. On dark backgrounds, that's no border, and light backgrounds it's `charcoal-1`.  Is the default border color changing?

fcoveram:
I see. Now it makes more sense.

> Would shrinking the focus of the search button also shrink it's clickable area, or is that outline inset?

The idea was shifting to a smaller button, so yes, a smaller clickable area. But with the `focus-within` context now I'm hesitant.

> I'm not sure I understand this. We have two styles of search— one on dark backgrounds and one on light backgrounds, so we can have two default border colors. On dark backgrounds, that's no border, and light backgrounds it's charcoal-1. Is the default border color changing?

Sorry I didn't explain it correctly. I was referring to the second screenshot where background colors are soften to show the focus style effect. Not intending to change the focus style but showing my idea through a different example.

---

I think the example you attached has an outer ring of 1.5px. If so, the thick difference makes it look blowing and odd. Here is a different idea that taps into the same but simulating an *internal* ring through 2 border rings.

image

The ring is always inner.
* Light background: one `1.5px` ring of blueberry
* Dark background: one `1.5px` ring of white, one `1.5px` ring of blueberry. Diagram attached below.



What do you think @jasmussen?

ryelle:
Okay, I've tried that in the latest commit. Here are some screenshots with focus, input first, button second.

| Default (on black) | Default (on blue) | Secondary (on light) |
|---|---|---|
| branch-default-1 | branch-blue-1 | branch-white-1 |
| branch-default-2 | branch-blue-2 | branch-white-2 |

I also updated the clear icon, which should fully close #368. 

Screenshot 2023-03-31 at 3 46 38 PM

ryelle:
I just realized my last screenshot doesn't actually show the clear icon, here's a better screenshot:

Screenshot 2023-04-03 at 12 07 55 PM

Mobile:
Screenshot 2023-04-03 at 12 08 51 PM

ryelle:
I'm going to merge this without design approval since @StevenDufresne discovered [a bug](https://github.com/WordPress/wporg-parent-2021/pull/79) on production due to these styles be out of sync. if there are any other problems, we can follow up in a new issue.

jasmussen:
I may have just given conflicting feedback in https://github.com/WordPress/wporg-main-2022/issues/228#issuecomment-1502952018, where I suggested a 1px radius when inset. That wasn't my intent. @panchovm what do you think? Happy to drop it if you prefer the square.

ryelle:
@jasmussen When you & @panchovm come to a consensus here, can you open a new issue? (if a change is necessary)

jasmussen:
It's a tiny detail, if it comes back up we'll be sure to open an issue.

PR #379, opened 2023-03-30
Header: Add background to logo on hover

### LINKED ISSUE #366: Header: Add hover style behind logo
We should add a hover style to the logo. Currently, the hover style is that of a dark background, and underlined text. We could apply that:

hover option 1

- [ ] Add background color on logo hover

This is likely the best and simplest route to go, as it leans into established patterns. Be mindful that the left/right padding of the navigation menu itself needs to be increased a little, so that the logo button can have uniform padding left and right. At the moment, the logo has additional padding to the left, for positioning purposes, which would make the hover style awkward.

ryelle:
Fixes #366 — Add the hover color behind the logo.

| Default | Blue | White | Classic |
|---|---|---|---|
| branch-default-1 | branch-blue-1 | branch-white-1 | branch-old-1 |
| branch-default-2 | branch-blue-2 | branch-white-2 | branch-old-2 |

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/logos.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/logos.pcss
@@ -40,10 +40,15 @@
 			padding-top: 16px;
 		}
 
+		&:hover {
+			background-color: var(--wp-global-header--background-color--hover);
+		}
+
 		&:focus-visible {
 			outline: none;
 			border-radius: 2px;
 			box-shadow: inset 0 0 0 1.5px var(--wp-global-header--link-color--active);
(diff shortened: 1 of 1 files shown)


jasmussen:
Thanks for doing this. It's a bit hard for me to compare the before and after, here's my best attempt:

Screenshot 2023-03-31 at 08 55 57

From the inspector, this seems like it should be fine, because the padding of the logo button is uniform:

Screenshot 2023-03-31 at 08 57 01

So it looks good (ship it).

When I initially created 366, though, I thought I recalled a non-uniform padding that might have needed finessing. But that may have been from the figma mockups:

Screenshot 2023-03-31 at 08 58 54

PR #380, opened 2023-03-30
Header: Add cachebuster to search icons.

iandunn:
The old images appear to be permanently cached on production, even when not being pulled via the CDN. This is a temporary fix until I can track down and fix the root cause.

See #375


iandunn:
Ah, I think it was because I was assuming the mid-deploy cachebuster applied to images, but it's only CSS/JS (and only ones that are registered with `wp_register_{script|style}`). 

Images still need to be done the "old" way:

1. commit/deploy the change to the image
1. bump the cachebuster in the URL, and commit/deploy that

So, this PR is all that's needed to fix it. I'll update the docs to clarify the distinction, though.

PR #381, opened 2023-03-30
Prevent scrolling behind search modal

### LINKED ISSUE #371: Search: Mobile overlay does not properly lock the screen in place
Demo video: 

```
html.has-modal-open {
    overflow: hidden;
}
```

```
html.has-modal-open {
    overflow: hidden;
    position: fixed;
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
}
```

iandunn:
Fixes #371
Closes #382

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -42,6 +42,7 @@ html {
 .global-header,
 .global-header__alert-banner,
 .global-footer,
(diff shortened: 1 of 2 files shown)


iandunn:
This is still a WIP. I found a few edge cases after applying the hotfix and fixed those, but there's at least 1 more that I'm working on.

iandunn:
I tried another approach in #382 and that seems more promising at the moment. I'll keep this open until that's finalized, though.

iandunn:
#392 will fix this

PR #382, opened 2023-03-31
Prevent scrolling behind search modal - Alternate

### LINKED ISSUE #371: Search: Mobile overlay does not properly lock the screen in place

iandunn:
Fixes #371
Closes #381

This is an alternate approach, because applying the hotfix in #381 caused uninteded consequences. Fixing those caused [more problems, which in turn had to be fixed, causing more problems](https://github.com/WordPress/wporg-mu-plugins/pull/381/commits).

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -10,6 +10,7 @@
  */
 
 /* Breakpoints should be synced with `wporg-parent-2021` (or `wporg-news-2021` until parent exists). */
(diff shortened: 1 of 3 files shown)


iandunn:
This is just a WIP at the moment, and needs more testing. So far though it seems more promising than #381, though.

iandunn:
#392 will fix this

PR #383, opened 2023-04-04
Admin Bar: Hide the admin bar for logged out users by default

### LINKED ISSUE #355: Navigation: Hide the adminbar in most cases, explore how to better integrate the register/login links
From WordPress/wporg-main-2022#221:

>   (Mobile) When a category is active (ie:Learn), the whole category of nav instances move position.

This shift is likely caused by the adminbar. Compare the homepage:

Screenshot 2023-03-27 at 16 10 33

with the Learn section:

Screenshot 2023-03-27 at 16 10 41

This issue is the same on desktop.

In general, the admin-bar is an affordance created for editors or members of the site, which presumably would be very few. In that light, it seems we should hide the admin-bar by default, and only show it in a few notable exceptions. 

For example, it might be fine to show it if you are a member, and are logged in. In this case, it should be shown across the whole site, except if it breaks the layout. But it should be shown with a #24282e background color, instead of the default #1b2024.

For some of the community sections like the Make sites, In the forums, or where community members expect to see the existing interface, it could also be okay to show it. In those sections, mainly the Log in and Register links are pertinent for logged out users, and as we get a chance to redesign those sections, it would be good to omit the adminbar in that state, and show the login/register links in a different context and affordance which might also afford better context for when and why you should. As an example, these are mockups made for the Pattern Directory redesign, which integrate the login link into the breadcrumb bar. Logged out:

Header, logged out

Logged in:

Group 1369

Outside of the typographic orphan caused by the word "improved?" in this screenshot (which we should fix separately, can just add a linebreak after "helpful?"), [this section of the documentation](https://wordpress.org/documentation/article/first-steps-with-wordpress-block-editor/) is nevertheless a good example of how a log-in link can be shown with context, without the need for an adminbar:

Screenshot 2023-03-27 at 16 18 14

ryelle:
Fixes #355 — Hides the admin bar unless you're logged in and have permissions on the site. Individual sites can override this by activating the Logged Out Admin Bar plugin, which is already active on many sites (see https://github.com/WordPress/wporg-mu-plugins/issues/355#issuecomment-1496286517).

Additionally, this fixes the background color of the admin bar, so that it matches the black version of the header bar.

**Screenshots**

| Small | Large |
|---|---|
| Screen Shot 2023-04-04 at 12 53 48 | Screen Shot 2023-04-04 at 12 54 00 |

**To test**

- Be logged out
- View the /about/ page or any child page
- There should be no admin bar
- Log in
- Reload that page
- Now there is an admin bar

If you test on a sandbox, you will still see the admin bar on the various directories, make.w.org/*, support forums, etc. even when logged out.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/admin-bar.php
+++ b/mu-plugins/blocks/global-header-footer/admin-bar.php
@@ -11,39 +11,16 @@ add_action( 'admin_bar_menu', __NAMESPACE__ . '\filter_admin_bar_links', 500 ); 
 add_filter( 'show_admin_bar', __NAMESPACE__ . '\should_show_admin_bar' );
 
 /**
- * Show the admin bar most of the time.
+ * Hide the admin bar for logged out users.
+ *
+ * The admin bar can be shown to logged out users by activating the Logged Out Admin Bar plugin.
  *
  * @param bool $show_admin_bar Whether the admin bar should be shown.
  * @return bool
  */
 function should_show_admin_bar( $show_admin_bar ) {
-	$can_access_admin = is_super_admin() || current_user_can( 'read' );
-
 	// Folks who can access wp-admin should always see the bar, for convenience.
-	if ( $can_access_admin ) {
-		return true;
-	}
-
-	// w.org/showcase visitors are unlikely to have/want an account.
-	// w.org/showcase/submit-a-wordpress-site/ has a login link for that edge case.
-	if ( 718 === get_current_blog_id() ) {
-		return false;
-	}
-
-	$is_download_page = get_permalink() === get_download_url();
-
-	/*
-	 * Visitors are confused by the admin bar on w.org/home and w.org/download, and it generates a large
-	 * number of support requests. Folks mistakenly assume they have to register for a w.org account in
-	 * order to download WP or set up a site.
-	 */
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/global-header-footer/admin-bar.php
```diff
@@ -11,39 +11,16 @@
...
+ *
+ * The admin bar can be shown to logged out users by activating the Logged Out Admin Bar plugin.
  *
  * @param bool $show_admin_bar Whether the admin bar should be shown.
  * @return bool
  */
 function should_show_admin_bar( $show_admin_bar ) {
-	$can_access_admin = is_super_admin() || current_user_can( 'read' );
-
 	// Folks who can access wp-admin should always see the bar, for convenience.
```
StevenDufresne: Maybe this comment should be updated.

StevenDufresne:
This makes sense to me. I think @iandunn's has better context to approve this PR.

iandunn (approved):
This looks good to me 👍🏻

mu-plugins/blocks/global-header-footer/admin-bar.php
ryelle: I'll remove it then, it's covered by the function comment just a few lines up.

outdoor2kode (approved):
Tested and everything looks good 👍

PR #384, opened 2023-04-05
Menu: Remove section-level menu highlight

### LINKED ISSUE #370: Navigation: Remove navigation highlights when on subpages/posts

> When in subpages, the top nav highlights the main page. Worth removing to not mislead the user with wayfinding.

As an example, in the main news section, the item is highlighted:

Screenshot 2023-03-27 at 12 08 24

When in subpages or permalinks, it remains highlighted (parent):

Screenshot 2023-03-27 at 12 08 35

This is similar in About and About subpages:

Screenshot 2023-03-27 at 12 09 42

Screenshot 2023-03-27 at 12 09 53

While the ability to highlight the parent and ancestor of a page or post is a pattern we're leveraging here, it hasn't materially aided wayfinding, and has caused a few issues on its own. 

Should we hide this highlight? It seems especially an issue on mobile:

Screenshot 2023-03-27 at 12 36 02

Screenshot 2023-03-27 at 12 40 54

ryelle:
Fixes #370 — This removes the previous "section" highlights (where a page matching anything in one of the submenu's selections would highlight every level to the top). Now the menu items only highlight when they are an exact match for the current page.

**Screenshots**

| | Large screen | Small screen |
|---|---|---|
| Top level match | large-top-level-item | small-top-level |
| 2nd level match | large-sub-item | small-sub-item |
| 3rd level, no match in menu | large-sub-inner<br />Single plugin | small-sub-inner<br />About - Accessibility page |

Subpages in News also no longer highlight the "News" item:

large-news-inner

**To test**

Due to how the menu works, you need a sandbox to test this. Just make sure the menu items are correctly highlighted, while not highlighting the top-level items.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -943,100 +943,26 @@ function is_wporg_network() {
  * @return array The altered menu items.
  */
 function set_current_item_class( $menu_items ) {
-	$current_url = get_menu_url_for_current_page( $menu_items );
+	$host        = strtolower( $_SERVER['HTTP_HOST'] ); // phpcs:ignore
+	$uri         = strtolower( $_SERVER['REQUEST_URI'] ); // phpcs:ignore
+	$current_url = "https://{$host}{$uri}";
 
 	foreach ( $menu_items as & $item ) {
-		$sub = false;
 		if ( ! empty( $item['submenu'] ) ) {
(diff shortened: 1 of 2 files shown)


StevenDufresne:
Working as expected. 👍 

Tested:
- Main site
- Meta Trac
- fr.wordpress.org

StevenDufresne (approved):

PR #385, opened 2023-04-06
Better ARIA support for site breadcrumbs

alexstine:
Not sure if it is valid to do it this way or not, may cause block validation errors.

### DIFF
--- a/mu-plugins/blocks/site-breadcrumbs/index.php
+++ b/mu-plugins/blocks/site-breadcrumbs/index.php
@@ -80,5 +80,5 @@ function render_block( $attributes, $content, $block ) {
 	}
 
 	$wrapper_attributes = get_block_wrapper_attributes();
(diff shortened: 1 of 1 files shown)


alexstine:
CC @ryelle

mu-plugins/blocks/site-breadcrumbs/index.php
```diff
@@ -80,5 +80,5 @@ function render_block( $attributes, $content, $block ) {
 	}
 
 	$wrapper_attributes = get_block_wrapper_attributes();
-	return sprintf( '<div %s>%s</div>', $wrapper_attributes, $content );
+	return sprintf( '<div role="navigation" aria-label="%s" %s>%s</div>', esc_attr_e( 'Breadcrumbs', 'wporg' ), $wrapper_attributes, $content );
```
ryelle:
```suggestion
	return sprintf( '<div role="navigation" aria-label="%s" %s>%s</div>', esc_attr__( 'Breadcrumbs', 'wporg' ), $wrapper_attributes, $content );
```

ryelle (approved):
Looks good. No block validation errors since this is a fully dynamic block. Thanks for the PR!

PR #386, opened 2023-04-07
Signal external links

### LINKED ISSUE #359: Navigation: Signal external links
At the moment, the main navigation contains a mix of internal and external links. In this case, internal means the link takes you to a section that maintains the global WP.org header, and external means it takes you to one that does not. 

We should sort items here according to their destination, and include an indicator when they are external. Eg. instead of this:

Screenshot 2023-03-28 at 10 06 05

We should do this:

Download

Additional sections:

Learn

Community

About

Download

That's the unicode northeast arrow: `↗`.

The above do not preclude separate changes to the menus, we may want to rename or move a few items around still, a separate issue will be opened if need be. But just as a baseline, this would make the navigation clearer.

outdoor2kode:
Fixes: #359

This PR adds northeast arrows to external links and adds the `white-space: nowrap` to prevent `WordPress Swag Store ↗︎` from being wrapped.

https://user-images.githubusercontent.com/18050944/230524196-3e5afbbc-5761-4301-84bb-9f82c06ce46b.mp4

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
@@ -187,6 +187,7 @@
 		padding-top: calc(var(--wp--style--block-gap) / 2);
 		padding-bottom: calc(var(--wp--style--block-gap) / 2);
 		border: none;
+		white-space: nowrap;
 
 		@media (--tablet) {
 			font-size: inherit;
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -443,11 +443,6 @@ function get_global_menu_items() {
 					'type'  => 'custom',
 				),
 				array(
-					'title' => esc_html_x( 'Openverse', 'Menu item title', 'wporg' ),
-					'url'   => 'https://openverse.org/',
-					'type'  => 'custom',
-				),
-				array(
 					'title' => esc_html_x( 'Mobile', 'Menu item title', 'wporg' ),
 					'url'   => 'https://wordpress.org/mobile/',
 					'type'  => 'custom',
@@ -455,6 +450,11 @@ function get_global_menu_items() {
 				array(
 					'title' => esc_html_x( 'Hosting', 'Menu item title', 'wporg' ),
 					'url'   => 'https://wordpress.org/hosting/',
+					'type'  => 'custom',
+				),
+				array(
+					'title' => esc_html_x( 'Openverse ↗︎', 'Menu item title', 'wporg' ),
+					'url'   => 'https://openverse.org/',
 					'type'  => 'custom',
 				),
 			),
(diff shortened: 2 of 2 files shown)


renintw:
@ryelle Wondering if you have any idea about where this stylesheet comes from?

image
image

as I was trying to adjust that min-width through [here](https://github.com/WordPress/wporg-mu-plugins/pull/386/files#diff-98560942f00c361d922f8ce0e2feac743fe8f46df9c2b95e40ea2d8bab936d0dR254), but it didn't work. And I'm not sure how to make [global-header__overflow-menu](https://github.com/WordPress/wporg-mu-plugins/pull/386/files#diff-98560942f00c361d922f8ce0e2feac743fe8f46df9c2b95e40ea2d8bab936d0dR213) and the [second & .wp-block-navigation__submenu-container](https://github.com/WordPress/wporg-mu-plugins/pull/386/files#diff-98560942f00c361d922f8ce0e2feac743fe8f46df9c2b95e40ea2d8bab936d0dR251) show up. They seem to make [min-width: 200px](https://github.com/WordPress/wporg-mu-plugins/pull/386/files#diff-98560942f00c361d922f8ce0e2feac743fe8f46df9c2b95e40ea2d8bab936d0dR254) not loaded.

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
```diff
@@ -187,6 +187,7 @@
 		padding-top: calc(var(--wp--style--block-gap) / 2);
 		padding-bottom: calc(var(--wp--style--block-gap) / 2);
 		border: none;
+		min-width: 210px;
```
StevenDufresne: I think you can accomplish the same by using `white-space: nowrap`.

StevenDufresne:
Just noting that this PR correctly and intentionally moves the external links to the bottom of the list.

mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
outdoor2kode: yeah, you're right. Thanks.

ryelle (approved):
Looks good 👍🏻

fcoveram (approved):
It looks great to me ✨

jasmussen:
Looks great, nice work!

One note, white-space: nowrap; may be appropriate in the dropdown menus, but it's not likely to be appropriate always, as in some contexts it will just expand the width of a column.

Another older trick we've used in the past, is to prevent word-breaks using non breaking spaces. The following would allow the arrow to wrap onto a new line:

`WordPress Swag Store ↗`

But the following would only allow "Store ↗" to wrap onto a new line:

`WordPress Swag Store&nbsp;↗`

This is an older hack, so I'm actually unsure whether this is appropriate in terms of screen readers, if any of you have insights I'd love to learn. Voiceover seems to read that in a not too great way.

ryelle:
I've merged this with the `white-space: nowrap` fix since there aren't any issues with the About dropdown being slightly larger.

> This is an older hack, so I'm actually unsure whether this is appropriate in terms of screen readers, if any of you have insights I'd love to learn. Voiceover seems to read that in a not too great way.

It should not read the space, since it's just a space. Is the issue you're hearing the arrow? It's read as "WordPress Swag Store Northeast arrow with variation selector 15" and that is correct given the emoji arrow. That's how it's read anywhere the arrow is used in the content.

PR #389, opened 2023-04-10
Table of Contents: Add toggle

alexstine:
Adds a toggle for the Table of Contents block.

See https://github.com/WordPress/wporg-documentation-2022/pull/58 for additional context.

### DIFF
--- a/mu-plugins/blocks/table-of-contents/postcss/style.pcss
+++ b/mu-plugins/blocks/table-of-contents/postcss/style.pcss
@@ -6,9 +6,9 @@
 	& h2 {
 		margin: 0 !important;
 		color: var(--wp--preset--color--charcoal-1);
-		line-height: var(--wp--custom--body--large--typography--line-height);
 		font-family: var(--wp--preset--font-family--inter);
-		font-size: var(--wp--preset--font-size--large);
+		font-size: var(--wp--preset--font-size--small);
+		line-height: 1.6;
 	}
 
 	& button {
@@ -26,6 +26,12 @@
 			border-radius: 2px;
 		}
 	}
+}
+
+.wporg-table-of-contents__list {
+	padding-left: 1em;
+	font-size: var(--wp--preset--font-size--small);
+	line-height: var(--wp--custom--body--small--typography--line-height);
 }
 
 .wporg-table-of-contents__list a {
--- a/mu-plugins/blocks/table-of-contents/postcss/style.pcss
+++ b/mu-plugins/blocks/table-of-contents/postcss/style.pcss
@@ -1,3 +1,13 @@
+:where(.wp-block-wporg-table-of-contents) {
+	padding-top: 15px;
+	padding-right: var(--wp--preset--spacing--20);
+	padding-bottom: 15px;
+	padding-left: var(--wp--preset--spacing--20);
+	background: var(--wp--preset--color--blueberry-4);
+	color: var(--wp--preset--color--blueberry-1);
+	font-size: var(--wp--preset--font-size--small);
+}
+
 .wporg-table-of-contents__header {
 	display: grid;
 	grid-template-columns: 1fr auto;
@@ -39,6 +49,8 @@
 
 		& ul {
 			margin-top: 3px;
+			padding-left: 1.5em;
+			list-style-type: disc;
 		}
 
 		&:last-child {
--- a/mu-plugins/blocks/table-of-contents/postcss/style.pcss
+++ b/mu-plugins/blocks/table-of-contents/postcss/style.pcss
@@ -29,17 +29,30 @@
 }
 
 .wporg-table-of-contents__list {
+	margin-bottom: 0;
 	padding-left: 1em;
 	font-size: var(--wp--preset--font-size--small);
-	line-height: var(--wp--custom--body--small--typography--line-height);
-}
+	line-height: 1.6;
 
-.wporg-table-of-contents__list a {
-	&:focus-visible {
-		outline: 1.5px solid var(--wp--preset--color--blueberry-1);
-		outline-offset: 2px;
-		box-shadow: none;
-		border-radius: 2px;
+	& li {
+		margin-bottom: 3px;
+
+		& ul {
+			margin-top: 3px;
+		}
+
+		&:last-child {
+			margin-bottom: 0;
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


alexstine:
@ryelle Here is the basic code but not sure how to inject JavaScript on the front-end for this. Is there a way with blocks or do I just need to call `wp_register_script(), wp_enqueue_script()`?

ryelle:
There's already a front end script for this block, it's at `table-of-contents/src/view.js`, currently it just handles the positioning to make the sidebar stay stuck to the screen while scrolling the page. If we move away from that layout (of having a sidebar) we can probably remove most of that code… but it will depend on the design. If you implement the toggle here, I can work with @jasmussen on how it should look 👍🏻

alexstine:
@ryelle Got it working.

ryelle:
I've updated this PR with the styles suggested by @fcoveram. The table of contents is collapsed by default so that it's easy to skip. It's collapsed at all sizes, even when it's shifted to the sidebar, so that it is still skipped for screen reader users. This makes for kind of an awkward visual, with the only thing in the sidebar a toggle button. The alternative would be to have the table of contents expanded by default when the screen is wide enough (1200px). The list would still be collapsible, so a screen reader user can collapse it to get by. Would that be acceptable, @alexstine?

Screenshots for design review:

| | Collapsed (default) | Expanded |
|---|---|---|
| Small screen (450) |  |  |
| Mid-sized screen (960) |  |  |
| Large screen (1400) |  |  |

mu-plugins/blocks/table-of-contents/postcss/style.pcss
```diff
@@ -1,11 +1,45 @@
...
+			outline: 1.5px solid var(--wp--preset--color--blueberry-1);
+			outline-offset: -0.5px;
+			box-shadow: none;
+			border-radius: 2px;
+		}
+	}
+}
+
+.wporg-table-of-contents__list {
+	padding-left: 1em;
```
StevenDufresne, line 43:
```suggestion
	margin-bottom: 0;
	padding-left: 1em;
```

I think we can remove the bottom margin since it's doubled up.

StevenDufresne: I also think we need to increase the line-height to match the designs.

mu-plugins/blocks/table-of-contents/src/view.js
```diff
@@ -65,26 +65,58 @@ function onScroll() {
...
+
+			// After toggle, see if we need to update the sidebar classes.
+			if ( isSidebarWithinViewport( container ) ) {
+				container.parentNode.classList.add( 'is-fixed-sidebar' );
+			} else {
+				container.parentNode.classList.remove( 'is-fixed-sidebar' );
+				window.scrollTo( { top: 0, left: 0, behavior: 'instant' } );
+			}
+			container.parentNode.classList.remove( 'is-bottom-sidebar' );
+			const isBottom = onScroll();
```
StevenDufresne: It's not exactly clear what (or why) this would return a value. Can we add a documentation block to the `onScroll` function since I think this is the first time the return value is assigned?

jasmussen:
Looks good to me. My only question would be whether we can have the table of contents be expanded by default, if you're on a large (two columns) screen.

mu-plugins/blocks/table-of-contents/src/view.js
ryelle: yeah… I don't love the workaround I landed on here. I'll add in some docs, and maybe think about better naming.

mu-plugins/blocks/table-of-contents/postcss/style.pcss
ryelle: > I also think we need to increase the line-height to match the designs.

The designs were updated in [this comment](https://github.com/WordPress/wporg-documentation-2022/pull/58#issuecomment-1516077452). Initially I thought I had the wrong line height so I increased it… but a closer look at the figma, I see there's a 3px margin between each list item. So it should be 1.6 with that margin, and it is now in https://github.com/WordPress/wporg-mu-plugins/pull/389/commits/ecdf69799e84330d0b305ef6bf982c6816904dd9.

Screenshot 2023-05-08 at 3 22 04 PM

mu-plugins/blocks/table-of-contents/src/view.js
ryelle: Added more docs in https://github.com/WordPress/wporg-mu-plugins/pull/389/commits/3f6ed32fba62899956e857c2ec86fac1e43c9a0e

fcoveram (commented):
I notice an unbalance spacing in the sides. Sorry that I can't point out specific values and what might cause them, but here is a screenshot highlighting the areas.

TOC component in small and large screen viewports. Paddings of the component are highlighted

ryelle:
@fcoveram Is that what it currently is, and it should be different? Can you tell me the values each side should be?

alexstine:
Collapsed or expanded by default is fine, whichever way works better visually.

Thanks.

fcoveram:
@ryelle I designed it with 15px for top and bottom paddings and 20px for left and right. But I understand that the outcome doesn't necessarily match Figma design, so adding a mockup with the spacing guides for reference.

Mockup of TOC component with spacing guides for the sides

ryelle:
@fcoveram I think you're looking at the production site, the branch here has cleaned up some of those spacing issues. I just fixed it up a bit more so it should match your design now.

Screenshot 2023-05-10 at 4 47 14 PM

(once this is merged, the templates in documentation & developers will need to updated to remove the attributes, since they're not necessary)

ryelle:
> Looks good to me. My only question would be whether we can have the table of contents be expanded by default, if you're on a large (two columns) screen.

> Collapsed or expanded by default is fine, whichever way works better visually.

Okay, I've set it to be expanded by default when the screen is large enough for the table of contents to be visually in the sidebar, otherwise it's collapsed by default. The toggle is there for all sizes.

fcoveram:
It looks great, thanks @ryelle. I was looking at the images you shared in [this message](https://github.com/WordPress/wporg-mu-plugins/pull/389#issuecomment-1536870977) 🤔 . Nonetheless, the last update works excellent.

fcoveram (approved):

PR #390, opened 2023-04-12
Add Encryption methods.

dd32:
This adds a new set of Encryption methods based off of other implementations of it.

There are two methods to use it, the namespaced methods which throw specific exceptions, or the globally-defined functions that operate as "standard PHP" (which are detailed below)

It uses Additional Data ciphers (AEAD with XChaCha20-Poly1305) via the Sodium library and supports multiple keys

Using `$context` (Additional Data) is required, and should be either set to a per-user/site/post value or to a per-feature unique token.

For example, two keys could be defined
 - WPORG_ENCRYPTION_KEY: .... = accessed via `wporg_encrypt( $string, (string) $user_id )` & `wporg_decrypt( $encrypted_value, (string) $user_id )`
 - WPORG_PLUGINS_ENCRYPTION_KEY: ... = accessed via `wporg_encrypt( $string, (string) $user_id, 'plugins' )` & `wporg_decrypt( $encrypted_value, (string) $user_id, 'plugins' )`

This is based off a combination of https://github.com/WordPress/two-factor/pull/389 and an Automattic implementations of it.
~This has not been unit tested, and does not include any yet.~
This has not been written with the intention of running on older versions of WordPress or PHP.
~This has not been peer-reviewed.~ This may contain egregious errors on my part.
This is being submitted so as to provide the basis of https://github.com/WordPress/wporg-two-factor/pull/103

### DIFF
--- a/mu-plugins/encryption/class-hiddenstring.php
+++ b/mu-plugins/encryption/class-hiddenstring.php
@@ -1,0 +1,164 @@
+<?php
+namespace WordPressdotorg\MU_Plugins\Encryption;
+/**
+ * Class HiddenString. This is a copy of https://github.com/paragonie/hidden-string without the additional dependencies.
+ *
+ * The purpose of this class is to encapsulate strings and hide their contents
+ * from stack traces should an unhandled exception occur.
+ *
+ * The only things that should be protected:
+ * - Passwords
+ * - Plaintext (before encryption)
+ * - Plaintext (after decryption)
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+final class HiddenString
+{
+	/**
+	 * @var string
+	 */
+	protected $internalStringValue = '';
+
+	/**
+	 * Disallow the contents from being accessed via __toString()?
+	 *
+	 * @var bool
+	 */
+	protected $disallowInline = false;
+
+	/**
+	 * Disallow the contents from being accessed via __sleep()?
+	 *
+	 * @var bool
+	 */
+	protected $disallowSerialization = false;
+
+	/**
+	 * HiddenString constructor.
+	 * @param string $value
+	 * @param bool $disallowInline
+	 * @param bool $disallowSerialization
+	 *
+	 * @throws \TypeError
+	 */
+	public function __construct(
+		string $value,
+		bool $disallowInline = true,
+		bool $disallowSerialization = true
+	) {
+		$this->internalStringValue = self::safeStrcpy($value);
+		$this->disallowInline = $disallowInline;
+		$this->disallowSerialization = $disallowSerialization;
+	}
+
+	/**
+	 * @param HiddenString $other
+	 * @return bool
+	 * @throws \TypeError
+	 */
+	public function equals(HiddenString $other)
+	{
+		return \hash_equals(
+			$this->getString(),
+			$other->getString()
+		);
+	}
+
+	/**
+	 * Hide its internal state from var_dump()
+	 *
+	 * @return array
+	 */
+	public function __debugInfo()
+	{
+		return [
+			'internalStringValue' =>
+				'*',
+			'attention' =>
+				'If you need the value of a HiddenString, ' .
+				'invoke getString() instead of dumping it.'
+		];
+	}
+
+	/**
+	 * Wipe it from memory after it's been used.
+	 * @return void
+	 */
+	public function __destruct()
+	{
+		if (\is_callable('\sodium_memzero')) {
+			try {
+				\sodium_memzero($this->internalStringValue);
+				return;
+			} catch (\Throwable $ex) {
+			}
+		}
+	}
+
+	/**
+	 * Explicit invocation -- get the raw string value
+	 *
+	 * @return string
+	 * @throws \TypeError
+	 */
+	public function getString(): string
+	{
+		return self::safeStrcpy($this->internalStringValue);
+	}
+
+	/**
+	 * Returns a copy of the string's internal value, which should be zeroed.
+	 * Optionally, it can return an empty string.
+	 *
+	 * @return string
+	 * @throws \TypeError
+	 */
+	public function __toString(): string
+	{
+		if (!$this->disallowInline) {
+			return self::safeStrcpy($this->internalStringValue);
+		}
+		return '';
+	}
+
+	/**
+	 * @return array
+	 */
+	public function __sleep(): array
+	{
+		if (!$this->disallowSerialization) {
+			return [
+				'internalStringValue',
+				'disallowInline',
+				'disallowSerialization'
+			];
+		}
+		return [];
+	}
+
+	/**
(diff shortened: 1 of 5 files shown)


pkevan (approved):
Tested via https://github.com/WordPress/wporg-two-factor/pull/103#issuecomment-1505216060

mu-plugins/encryption/exports.php
```diff
@@ -0,0 +1,100 @@
...
+ * Encrypt a value.
+ *
+ * Unlike the Encryption plugin, this function simply returns false for any errors, and
+ * HiddenStrings that can be cast to string as needed.
+ *
+ * @param string $value The plaintext value.
+ * @param string $key   The key to use for encryption. Optional.
+ * @return string|false The encrypted value, or false on error.
+ */
+function wporg_encrypt( string $value, string $key = '' ) {
```
xknown: I think it'd be better to rename `$key` to something more appropriate, it looks like this only indicates which constant is going to be used to get the encryption key.

Also, it'd be good to only add/use `wporg_authenticated_encrypt` (or rename it to `wporg_encrypt`) and always enforce a non emtpy `$additional_data`.

mu-plugins/encryption/exports.php
```diff
@@ -0,0 +1,100 @@
...
+	try {
+		$value = \WordPressdotorg\MU_Plugins\Encryption\encrypt( $value, '', $key );
+
+		return new HiddenString( $value->getString(), false );
```
xknown: I don't think we need to use `HiddenString` here.

mu-plugins/encryption/class-hiddenstring.php
```diff
@@ -0,0 +1,164 @@
...
+	/**
+	 * HiddenString constructor.
+	 * @param string $value
+	 * @param bool $disallowInline
+	 * @param bool $disallowSerialization
+	 *
+	 * @throws \TypeError
+	 */
+	public function __construct(
+		string $value,
```
xknown, line 49: We can probably add the `#[\SensitiveParameter]` annotation here, althought it's only supported in PHP 8.2 :) (https://www.php.net/manual/en/class.sensitiveparameter.php)

mu-plugins/encryption/index.php
```diff
@@ -0,0 +1,151 @@
...
+	if ( $value instanceOf HiddenString ) {
+		$value = $value->getString();
+	}
+
+	$key       = get_encryption_key( $key );
+	$encrypted = sodium_crypto_aead_xchacha20poly1305_ietf_encrypt( $value, $additional_data, $nonce, $key->getString() );
+
+	sodium_memzero( $value );
+
+	return new HiddenString( PREFIX . sodium_bin2hex( $nonce . $encrypted ) );
```
xknown: As mentioned above, we don't probably need to use `HiddenString` here since there's nothing "sensitive" in an encrypted blob.

mu-plugins/encryption/index.php
```diff
@@ -0,0 +1,151 @@
...
+
+	if ( $key ) {
+		$constant = 'WPORG_' . str_replace( '-', '_', strtoupper( $key ) ) . '_ENCRYPTION_KEY';
+	}
+
+	if ( ! defined( $constant ) ) {
+		throw new Exception( sprintf( 'Encryption key "%s" not defined.', $constant ) );
+	}
+
+	return new HiddenString( sodium_hex2bin( constant( $constant ) ) );
```
xknown: It'd be probably better if these constants would be already instances of `HiddenString`.

mu-plugins/encryption/exports.php
dd32: > rename $key to something more appropriate

In my mind, it was "Which key should we use?" not "Here's an Encryption key", I'm not sure what the best alternative would for renaming that, "$key_slug"?. If presented an invalid value it throws an Exception, so a developer can't really misuse it.

> it'd be good to only add/use wporg_authenticated_encrypt (or rename it to wporg_encrypt)

That sounds reasonable.

mu-plugins/encryption/index.php
dd32: I err'd on the side of using it here just incase for some reason it was exposed somewhere, but you're right, it's encrypted, so there's little reason it should be - and if it's exposed accidentally, it's encrypted so it shouldn't be of great concern.

Changed with https://github.com/WordPress/wporg-mu-plugins/pull/390/commits/3caed5629abf96edb798faba3bf8a03ce198c7a7

mu-plugins/encryption/index.php
dd32: I agree, unfortunately WordPress.org is currently running PHP 7.4 which only supports scalar constants.

One alternative would to not store them within a Constant, but within instances of HiddenString returned by a function. That would remove the keys from the constant values and make them only accessible as the return value of a function.
I'll look into this.

dd32: I've shifted it to calling a function, which appears to work well in the tests.

https://github.com/WordPress/wporg-mu-plugins/pull/390/commits/3d5b2f6fdc39a80a6419971e4b3bc9973c348276#diff-313484cc480a64323ee0d3493c3e06055224c186fcb2e366f5e4dffe80cc4387R17-R27

mu-plugins/encryption/exports.php
dd32: Went with `$key_name` in 6468c2ddf987572f2966b895add6c909fea4fada

Removed the non-auth'd variants in https://github.com/WordPress/wporg-mu-plugins/pull/390/commits/6f8d488765aab108b4c4197a25ba5ff0d4e7b7ef

mu-plugins/encryption/class-hiddenstring.php
dd32: Added in https://github.com/WordPress/wporg-mu-plugins/pull/390/commits/6c9d58adc00f5dc4509e759a0859a5a88558ae0c

dd32:
Thanks for the review @xknown, I've made changes to reflect that.

mu-plugins/encryption/exports.php
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Decrypt a value, with authentication.
+ *
+ * Unlike the Encryption plugin, this function simply returns false for any errors, and
+ * HiddenStrings that can be cast to string as needed.
+ *
+ * @param string $value    The encrypted value.
+ * @param string $context  Additional, authenticated data. This is used in the verification of the authentication tag appended to the ciphertext, but it is not encrypted or stored in the ciphertext.
+ * @param string $key_name The name of the key to use for decryption. Optional.
+ * @return string|false The decrypted value, or false on error.
```
xknown: Nitpick: the return type should be `HiddenString|false` I think?

mu-plugins/encryption/index.php
```diff
@@ -0,0 +1,155 @@
...
+		throw new Exception( 'Invalid cipher text.' );
+	}
+
+	return new HiddenString( $plaintext );
+}
+
+/**
+ * Check if a value is encrypted.
+ *
+ * @param string $value Value to check.
```
xknown: Nitpick: the type of `$value` should probably be `HiddenString|string` or `mixed`.

xknown (approved):
LGTM.

PR #392, opened 2023-04-25
Search: Prevent visible page scroll when the search modal is open on iOS

### LINKED ISSUE #371: Search: Mobile overlay does not properly lock the screen in place

ryelle:
On iOS, opening the search modal in the header triggers the keyboard to open, which re-enables page scrolling. When scrolling down & back up, the page contents are visible in the gap behind the modal. This PR fixes the header bar on scroll, so that it hides that scrolled content.

It is still strange that the page can be scrolled, but the other fixes cause larger layout issues (see https://github.com/WordPress/wporg-mu-plugins/issues/371#issuecomment-1505719345), and this only happens when the keyboard is open (and only on iOS) this seems like the best solution.

Fixes #371, see #381, see #382.

Note: this PR also includes a small fix to prevent the icon-buttons from appearing active when the tab is closed, I only noticed it on mobile, so fixed it here.

**Screenshots**

| | Production | PR |
|---|---|---|
| Home: block theme |  |  |
| Patterns: classic theme, has admin bar |  |  |
| Download: block theme, white header style |  |  |
| Download, with keyboard closed, modal behaves correctly (no scroll) | no change |  |

**To test**

I used Charles.app's DNS spoofing to send *.wordpress.org traffic to my sandbox, then set my iPad to use my computer as a proxy. This let me load my sandboxed environment on my iPad. Then using the vertical orientation, I tested the header across different wporg pages.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/_common.pcss
@@ -10,6 +10,7 @@
  */
 
 /* Breakpoints should be synced with `wporg-parent-2021` (or `wporg-news-2021` until parent exists). */
+@custom-media --mobile-only (max-width: 889px);
 @custom-media --tablet (min-width: 890px);
 @custom-media --desktop (min-width: 1152px);
 @custom-media --desktop-wide (min-width: 1440px);
--- a/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/menu.pcss
@@ -321,3 +321,22 @@
 		overflow: scroll;
 	}
 }
+
+/*
+ * On small screens, when the nav modal is open, position the header to overlay
+ * the page content.
+ *
+ * Technically, this should not scroll - GB sets `overflow: hidden` - but when
+ * the keyboard is open, iOS overrides the scroll behavior. This solution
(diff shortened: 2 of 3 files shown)


iandunn (approved):
Looks good to me 👍🏻 

One of the reasons I haven't circled back to this yet is that the upstream fix is still changing, so whatever we do here might not work in the near future. Since this is already working, though, I don't see any reason to avoid merging it.

jasmussen:
I'm unable to test this in my local environment, but for those able to, I've had good success with the free Xcode simulator to test. If it works, ship it.

I also want to note that I appreciate that [you shared your experiences on the upstream fix](https://github.com/WordPress/gutenberg/pull/49369#issuecomment-1522385651), so we can get this fixed at the core! Thank you 🙏

fcoveram:
I see the search icon twice. The one in the placeholder should not exist, as my understanding. Is that a OS behavior?

jasmussen:
Good catch, Francisco. I think I actually have an upstream fix for that: https://github.com/WordPress/gutenberg/issues/48259 and https://github.com/WordPress/gutenberg/pull/49710

PR #400, opened 2023-05-17
#6945: add a props block to sync mentions to wp.org

gaambo:
Adds a "props" block as mentioned in [#6945](https://meta.trac.wordpress.org/ticket/6945).

## How
- Adds a new block `wporg/props` which is based on the core paragraph block (and similar to the notice block already existing in this repo). 
- Block supports are copied from the core paragraph as of Gutenberg version 15.1 which includes text align, colors, typograph and spacing controls
- On `save_post` it checks for existance of this block (also supports nested blocks) and parses all mentions in this props block.
- The props are then sent to WordPress.org via `WordPressdotorg\Profiles\add_activity`.
- Synced props are stored as post meta and on updating a post they are compared and only new props are synced.
- It supports multiple props blocks in a post and also props blocks nested in others (like groups).
- It supports all post types atm.
- I've added the `classnames` dependency since its used in gutenberg and it's my go-to. But wasn't added in this package until yet and I don't think the DependencyExtractionPlugin of core handles that package. If we want to avoid this package for the simple two lines in this block, I can remove that. 

## Questions:
- [ ] How can I test the `add_activity` call? Is there a test instance or a local instance I can get when running via wp-env?
- [ ] Which values does the `type` parameter of `add_activity` allow? 
- [ ] I'm assuming existing Make blogs use bbPress and therefore `bbp_find_mentions` to parse mentions. I'm reusing this function and provide a fallback if this is not available. If bbPress and this function will always be available, we can remove the fallback.
- [ ] It does not handle removing props (because I don't know if there's a `remove_activity` function?
- [ ] I'm using the generic `save_post` hook here but if we only want to support the post post type, I can switch that to `save_post_post`.

## How to test:
1. Checkout this PR and run `npm start` + `npm run wp-env start`
2. Create a new post
3. Check the "Props" block is available in the inserter
4. Enter some text in the block like "Props to @xy" 
5. You can add multiple props blocks and also nest them in others (like a group)
6. Save the post
7. ‼️ Atm this will trigger an error because on local environments `WordPressdotorg\Profiles\add_activity` does not exist
8. You can comment out this code, click update again and check that the meta field `_wporg_props` does exist and has the correct values in it

## Screenshots:
grafik

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,5 +64,7 @@
 			"selector-class-pattern": null
 		}
 	},
-	"dependencies": {}
+	"dependencies": {
+		"classnames": "^2.3.1"
+	}
 }
--- a/mu-plugins/blocks/props/index.php
+++ b/mu-plugins/blocks/props/index.php
@@ -1,0 +1,106 @@
+<?php
+/**
+ * Block Name: Notice Block
+ * Description: Add a color-coded notice to your post.
+ */
+
+namespace WordPressdotorg\MU_Plugins\Props;
+
+add_action( 'init', __NAMESPACE__ . '\init' );
+add_action( 'save_post', __NAMESPACE__ . '\handle_save_post', 10, 2 );
+
+/**
+ * Registers the block using the metadata loaded from the `block.json` file.
+ * Behind the scenes, it registers also all assets so they can be enqueued
+ * through the block editor in the corresponding context.
+ *
+ * @see https://developer.wordpress.org/reference/functions/register_block_type/
+ */
+function init() {
+	register_block_type( __DIR__ . '/build' );
+}
+
+/**
+ * Handles syncing the props to WordPress.org when publishing/updating a post with the props block in it
+ *
+ * @param int      $post_id The post id that is saved.
+ * @param \WP_Post $post The post object with the new data.
+ */
+function handle_save_post( $post_id, $post ) {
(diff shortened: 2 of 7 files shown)


gaambo:
I've used `npm run lint:js` locally but don't get the error messages that the CI shows me. 
Also I didn't commit my `package-lock.json` but `classnames` is in `package.json` so the CI build should get that?

StevenDufresne:
This PR is closed. Is that correct? I'm under the impression that it should still be open.

package.json
```diff
@@ -64,5 +64,7 @@
 			"selector-class-pattern": null
 		}
 	},
-	"dependencies": {}
+	"dependencies": {
+		"classnames": "^2.3.1"
```
StevenDufresne, line 68: I would avoid importing `classnames`. It don't think we need it for the block.

gaambo:
> This PR is closed. Is that correct? I'm under the impression that it should still be open.

@StevenDufresne 
I didn't get any feedback or information if that block is still needed, or if I took the right direction. Also since the last commit the block API has changed and probably a few changes to the block should be made to work perfectly with the newest core versions. Since I had no time (and currently have no time) to work on this, I closed it. 
Feel free to take it as-is and build on top of it.

StevenDufresne:
👍 Okay. I've added this to my list of things :)

PR #404, opened 2023-06-14
New Block: Local Navigation container

ryelle:
This introduces a new container block for the Local Navigation in the header. The purpose of this is mostly to keep the shared styles contained, while also giving us a way to handle any custom functionality (not needed right now, but it would have helped with the breadcrumbs issue WordPress/wporg-main-2022#325 — while that doesn't apply to the new design, something like that could come up again).

The proposed designs show a small local header on interior pages. This block would handle the black-background section.

Section Header L2

See https://github.com/WordPress/wporg-mu-plugins/issues/465

**Screenshots**

By default, this inserts a full-width blue banner, with breadcrumbs and a navigation block

default

The brush stroke style variation is also available (I'm not sure if this will carry over to the redesign, but it is currently used on the Downloads subpages).

after-download-nightly

And the colors can be fully customized, background, text, border, etc

after-about-reqs

The banner scrolls, sticking under the sticky-global-header.

after-about-reqs-scroll

Small/tablet sized screens 

mid-screen

On small screens, the menu collapses (in these screenshots I've had it use the text display so we don't get two stacked hamburgers 🍔 ). When open, it takes over the whole screen (native GB behavior). We could iterate on this some more for a less "in your face" menu, if needed.

| Small screen, closed | Small screen, open |
|---|---|
| small-screen | small-screen-open |

Menu with dropdowns are also supported, but not likely to be used on the redesign.

with-dropdown

**To Test**

- Try the PR for wporg-main-2022 using this block (TBD)
- Try using the site editor to add the block to a template
- Or mock up the block in the post editor, and copy it into a template file
- It should display as expected/nothing weird should happen 🙂


StevenDufresne (approved):
I ran this and it did work properly.

The title is "Try Local Navigation". Are there reasons to not merge this?

ryelle:
> The title is "Try Local Navigation". Are there reasons to not merge this?

Nope, was just to indicate that this was an idea I wanted feedback/review on.

StevenDufresne:
> > The title is "Try Local Navigation". Are there reasons to not merge this?
> 
> Nope, was just to indicate that this was an idea I wanted feedback/review on.

Got it. 

>while also giving us a way to handle any custom functionality

Does this block need to dynamically render right now? (excuse me if there is an obvious reason why and I'm missing it 😴 )

ryelle:
> Does this block need to dynamically render right now?

The main reason I defaulted to dynamic rendering was to prevent block validation errors if the markup does need to change — this way we could, for example, drop in a mobile-only toggle button or change the wrapper element without needing to write out deprecation transforms. It also simplifies the markup in the HTML templates when you don't need to write out both the block tag and an html tag.

PR #406, opened 2023-06-19
Add a GitHub App client abstraction

dd32:
We have several GitHub apps in use today, but each one has their own authorization client and details within them.

This introduces a more generic `GitHub_App_Authorization` class which can be used to make authenticated-as-app API requests against GitHub.

This builds off https://github.com/WordPress/wordcamp.org/commit/dd6035fbb18feaf9d1a68fb7b7d6edbe22401c0c and modified to suit.

See https://github.com/WordPress/wordpress.org/compare/trunk...dd32:try/make-gh-inviter for an implementation of using it as a client.

One major change here, other than adding the client, is that this adds non-development Composer dependencies. The build process is updated to skip installing dev deps (none of which appear to be needed for build at present) and ignore .gitignore as we want it included in the build.
The svn-sync process would need to move it from `./vendor/` to `./mu-plugins/vendor/`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
 
             - name: Install all dependencies
               run: |
-                  composer install
+                  composer install --no-dev
                   npm install
 
             - name: Build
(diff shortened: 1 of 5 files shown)


dd32:
Just noting, that given the composer dependency, I was tempted to replace `\Ahc\Jwt\JWT` with an included method, simply due to the sheer simplicity of JWT tokens, and that we're not needing to validate/etc.

That would simplify the PR changes.. but it seems like we should just use existing libraries where possible, and I've reviewed this library, I decided not to go down that route. I just wish the composer dependency didn't include all of it's tests and development files and was just the library by itself.

PR #410, opened 2023-06-22
Sidebar Container: Run the fixed sidebar check after expanding the table of contents

### LINKED ISSUE #409: Fixed sidebar creates loss of visible focus & unreachable items.
When the fixed sidebar is longer than the viewport, items in the sidebar that are below the scroll do not come into focus during keyboard navigation.  It's also not possible in that case to reach the lower part of the sidebar without scrolling to the end of the page.

There is a way to do this - if you toggle the sidebar closed, then re-open it, the `is-fixed-sidebar` class is removed, and you'll be able to use the sidebar normally. However, this is more "taking advantage of a bug to be able to use the site" than intent, I'm guessing. 

See [Roles and Capabilities](https://wordpress.org/documentation/article/roles-and-capabilities/) for an example of this problem; [see also this slack message in #docs](https://wordpress.slack.com/archives/C02RP4WU5/p1687386572350359) where this issue was noted.

ryelle:
Fixes #409 — When the page was loaded initially, the `isSidebarWithinViewport` check was run on the collapsed sidebar, so the height check in `isSidebarWithinViewport` passed and the sidebar is set as fixed. Then, a few lines later, the collapse/expand toggle logic expanded the sidebar on large screens, and now it could be too tall for the fixed sidebar behavior.

This fix just swaps the initial `isSidebarWithinViewport` to after the toggle button logic, so that the menu is in it's correct state when the height is measured.

To test:

- Use the documentation test site
- View a few articles with different ToC lengths, for example http://localhost:8888/article/roles-and-capabilities/, http://localhost:8888/article/before-you-create-a-network/
- Try at different screen sizes

### DIFF
--- a/mu-plugins/blocks/sidebar-container/src/view.js
+++ b/mu-plugins/blocks/sidebar-container/src/view.js
@@ -85,15 +85,6 @@ function isSidebarWithinViewport( container ) {
 
 function init() {
 	const container = document.querySelector( '.wp-block-wporg-sidebar-container' );
-
-	if ( container ) {
-		if ( isSidebarWithinViewport( container ) ) {
-			container.classList.add( 'is-fixed-sidebar' );
-			onScroll(); // Run once to avoid footer collisions on load (ex, when linked to #reply-title).
-			window.addEventListener( 'scroll', onScroll );
-		}
-	}
-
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
Nice work, LGTM :shipit:

PR #414, opened 2023-07-11
Add wporg/time block

### LINKED ISSUE #413: Add support for [time] shortcode
The [time] shortcode is included in the o2 and p2 plugins. Find a way to make it available when this theme is loaded.

adamwoodnz:
Closes #413 

This PR reimplements the existing [[time] shortcode](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-time-shortcode.php) as a Gutenberg format.

The block uses the Format API so that the time can be applied to a string within a parent block, eg. a paragraph.

#### Notable changes from the existing shortcode implementation:

1. The reformatting is achieved by filtering and modifying `the_content` rather than the shortcode content
2. The time parsing (`parse_time`) is separated from the dom manipulation (`transform_time_blocks`)
3. The time parsing is always relative to the post date, there is no [attribute support](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-time-shortcode.php#L53)
4. The frontend `time_converter_script` no longer has a jQuery dependency

### Screenshots

| Editor | Frontend |
|-|-|
| Screenshot 2023-07-11 at 1 22 27 PM | Screenshot 2023-07-11 at 5 44 40 PM | 

### Testing

1. Create a post, add various valid (eg. 'Tuesday, April 5th, at 15:00 UTC') and invalid datetime descriptions, and format as a `Time` using the option in the rich text toolbar 
2. Save and view on the frontend
3. The date displayed should be reformatted to your timezone
4. Test various styles
5. Change the post date and check that the displayed time changes too
6. Add multiple Times and check that the JS is only added to the page once
7. Test cross browser and [include Safari](https://github.com/WordPress/wporg-mu-plugins/pull/414/files#diff-56e893c6923a0a2c2b2277cfea55838a748dc980d25fb141597e82c8b4858b51R137)


mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,175 @@
...
+					}
+				)();
+
+				document.querySelectorAll('.wporg-time-date').forEach( ( dateElement ) => {
+					let localTime = '';
+					const date = parseDate( dateElement.getAttribute( 'title' ) );
+
+					if ( date ) {
+						if ( ! toLocaleTimeStringSupportsLocales ) {
+							console.log("🚀 ~ file: index.php:143 ~ document.querySelectorAll ~ toLocaleTimeStringSupportsLocales:", toLocaleTimeStringSupportsLocales)
```
adamwoodnz: Remove this after testing

mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,177 @@
...
+
+	return $time;
+}
+
+/**
+ * Prints script to convert time in the viewers local time zone.
+ */
+function time_converter_script() {
+	?>
+	<script type="text/javascript" id="time_converter_script">
```
ryelle: Could you add this on the frontend via a `viewScript`?

mu-plugins/blocks/time/src/index.js
```diff
@@ -0,0 +1,32 @@
...
+			);
+		} }
+		isActive={ isActive }
+	/>
+);
+
+registerFormatType( metadata.name, {
+	title: metadata.title,
+	tagName: 'a',
+	className: 'wporg-time',
```
ryelle: This creates a link that doesn't do anything — could you maybe use the `time` element instead? Use the PHP to add the `datetime` attribute, and then use that to build up the element in JS?

mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,177 @@
...
+		if ( $parsed_time === null ) {
+			continue;
+		}
+		
+		// Build the link and abbr microformat.
+		$time_element->setAttribute( 'href', 'https://www.timeanddate.com/worldclock/fixedtime.html?iso=' . gmdate( 'Ymd\THi', $parsed_time ) );
+
+		$new_time_content = $dom->createElement( 'abbr', $time_content );
+		$new_time_content->setAttribute( 'class', 'wporg-time-date' );
+		$new_time_content->setAttribute( 'title', gmdate( 'c', $parsed_time ) );
```
ryelle: Do you know if there has been any feedback on the accessibility of this `abbr` + `title` approach?

ryelle (changes requested):
This is registering a block on the server, but the code in `index.js` doesn't actually create a block. While `register_block_type` works, it would be more correct to register the script a different way. You could try [the method I used for showcase](https://github.com/WordPress/wporg-showcase-2022/blob/e6e247a49c6063c06d3414f6cd93ad9a14b402c4/source/wp-content/themes/wporg-showcase-2022/functions.php#L134-L147), which still uses the `block.json` and the generated index.asset.php.

Overall, the references to this being a block are not accurate - the parent folder & block.json are fine because those are necessary for the build process (much like the "style" block we use in the themes). For clarity though it would be best to update those references (like in the docs).

It looks like we never set up php linting on this project, but I see a few issues in`time/index.php`. You can run `composer run lint mu-plugins/blocks/time` to see them.

ryelle:
> The time parsing is always relative to the post date, there is no [attribute support](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-time-shortcode.php#L53)

A quick search on make.w.org/core shows that they use the `relative` attribute often, so it would be good to think about how we could add support (maybe in the future).

mu-plugins/blocks/time/index.php
adamwoodnz: I did try that with no success, I'll have another go

mu-plugins/blocks/time/index.php
adamwoodnz: I don't, I'll ask

mu-plugins/blocks/time/src/index.js
adamwoodnz: Done

mu-plugins/blocks/time/index.php
adamwoodnz: Added a `viewScript` entry but it isn't automatically enqueued. I've used something similar to [the method you used for showcase](https://github.com/WordPress/wporg-showcase-2022/blob/e6e247a49c6063c06d3414f6cd93ad9a14b402c4/source/wp-content/themes/wporg-showcase-2022/functions.php#L134-L147) for this also

mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,124 @@
...
+	$xpath = new \DOMXPath( $dom );
+	$time_elements = $xpath->query( "//*[contains(concat(' ', normalize-space(@class), ' '), ' wporg-time ')]" );
+
+	if ( empty( $time_elements ) ) {
+		return $content;
+	}
+
+	$metadata = get_metadata();
+	$time_converter_script_handle = register_block_script_handle( $metadata, 'viewScript', 0 );
+	wp_enqueue_script( $time_converter_script_handle, null, null, filemtime( __DIR__ . '/build/convert_times.js' ), true );
```
adamwoodnz, line 72: I've added all the parameters so that it loads in the footer. Not sure of there is a better way.

mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,124 @@
...
+	$dom->loadHTML( $content, LIBXML_NOERROR );
+	$xpath = new \DOMXPath( $dom );
+	$time_elements = $xpath->query( "//*[contains(concat(' ', normalize-space(@class), ' '), ' wporg-time ')]" );
+
+	if ( empty( $time_elements ) ) {
+		return $content;
+	}
+
+	$metadata = get_metadata();
+	$time_converter_script_handle = register_block_script_handle( $metadata, 'viewScript', 0 );
```
adamwoodnz, line 71: Couldn't find a nice way to register this script in `init` for better performance, and then only enqueue it here

adamwoodnz:
> It looks like we never set up php linting on this project, but I see a few issues in`time/index.php`

After my MacOS upgrade to Ventura my php version went back to 8 and I had phpcs errors which blocked me seeing these in VSCode. Fixed now.

mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,124 @@
...
+function transform_times( $content ) {
+	if ( empty( $content || is_admin() ) ) {
+		return $content;
+	}
+
+	// Find the time block elements by the classname "wporg-time"
+	$dom = new \DOMDocument();
+
+	// Ignore warnings about htlm5 tags.
+	$dom->loadHTML( $content, LIBXML_NOERROR );
```
adamwoodnz, line 62: Without this there can be **many** warnings displayed (nav, time, and many other html5 tags aren't supported).

Alternatively we could use regexp or a different dom parser...

mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * Builds the time block output.
+ *
+ * This implements replacing the raw time strings in the post content with formatted times able to be converted to local times with JS.
+ *
+ * @param string $content Post content.
+ * @return string Content with display times reformatted.
+ */
+function transform_times( $content ) {
+	if ( empty( $content || is_admin() ) ) {
```
pkevan: presumably this is trying to check for empty $content or is_admin? Should it not be:

```
if ( empty( $content ) || is_admin() ) {
```

adamwoodnz: Yes, nice catch :)

adamwoodnz:
Feedback received elsewhere: it would be best to avoid filtering `the_content` on page view, especially if it involves loading a DOM parser.

Some possible solutions for this:
- hook into the content save flow for php based DOM manipulation (probably a filter)
- use php regexp to find and replace instead of DOM parsing
- do all the parsing and DOM manipulation client side

adamwoodnz:
> You could try [the method I used for showcase](https://github.com/WordPress/wporg-showcase-2022/blob/e6e247a49c6063c06d3414f6cd93ad9a14b402c4/source/wp-content/themes/wporg-showcase-2022/functions.php#L134-L147), which still uses the `block.json` and the generated index.asset.php

@ryelle do you see any benefits with this approach, compared to [this example](https://github.com/WordPress/gutenberg-examples/blob/trunk/non-block-examples/format-api/index.php)?

adamwoodnz:
Alternative solution using client side parsing in #415

mu-plugins/blocks/time/index.php
adamwoodnz: I [asked](https://wordpress.slack.com/archives/C02RP4X03/p1689894967035259), feedback was that a `time` element would be better, and in particular the `title` on the `abbr` could be bad.

adamwoodnz:
Closing in favour of #415

PR #415, opened 2023-07-18
Add time format with client side date parsing

### LINKED ISSUE #413: Add support for [time] shortcode

adamwoodnz:
Closes #413 

This is an alternative approach to the server side implementation in #414 

Pros:
- date description parsing is done at the time of applying the format, rather than on page load on the server
- no filtering of `the_content`

Cons:
- if the date to be parsed by the format needs to be changed, or the post date is changed, the time format needs to be toggled off and on again to re-parse the changed date

UPDATE:
If the time description changes, the format will now be removed so that it can be reapplied and parsed.


mu-plugins/blocks/time/index.php
```diff
@@ -0,0 +1,42 @@
...
+			}
+		}
+	);
+
+	$view_script_handle = register_block_script_handle( $metadata, 'viewScript', 0 );
+	add_action(
+		'wp_enqueue_scripts',
+		function() use ( $view_script_handle ) {
+			if ( ! is_admin() ) {
+				wp_enqueue_script( $view_script_handle, null, null, filemtime( __DIR__ . '/build/convert_times.js' ), true );
```
tellyworth: It's generally best to use a literal version number instead of `filemtime()` in production, otherwise every page view requires a FS read. We're not exactly consistent about that across the codebase though. Not a blocker.

mu-plugins/blocks/time/src/convert_times.js
```diff
@@ -0,0 +1,80 @@
...
+
+			localTime += formatTime( date );
+
+			dateElement.innerText = localTime;
+			dateElement.removeAttribute( 'data-iso' );
+
+			const linkElement = document.createElement( 'a' );
+			linkElement.setAttribute(
+				'href',
+				`https://www.timeanddate.com/worldclock/fixedtime.html?iso=${ datetimeISO }`
```
tellyworth: Just noting for future possibilities: should we consider replacing timeanddate.com?

mu-plugins/blocks/time/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+				if ( ! dateDescription || isActive ) {
+					toggleWithoutEnhancing();
+
+					return;
+				}
+
+				// Remove the word "at" from the string, if present.
+				// Allows strings like "Monday, April 6 at 19:00 UTC" to work.
+				const dateCleaned = dateDescription.replace( 'at ', '' );
```
tellyworth, line 94: (I was curious about why this was here - looks like it's from the original shortcode)

tellyworth (approved):
Looks fine to me. I've only reviewed the code, not tested.

mu-plugins/blocks/time/index.php
ryelle: This script is already registered by the `register_block_script_handle` function with the generated dependencies & version string, so you can enqueue it by just using the handle. To also output it in the footer, you can change the `group` — that's all the `$in_footer` [parameter does](https://github.com/WordPress/wordpress-develop/blob/bea4307daa21a3f97d159898b0ac676332e0e7de/src/wp-includes/functions.wp-scripts.php#L173-L184).

```suggestion
				wp_enqueue_script( $view_script_handle );
				wp_script_add_data( $view_script_handle, 'group', 1 );
```

mu-plugins/blocks/time/src/convert_times.js
```diff
@@ -0,0 +1,80 @@
+function convertTimes() {
```
ryelle: [File names should be separated with a hyphen](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#naming-conventions), so this should be `convert-times.js`. Or you could use `view.js` if this is meant to be the `viewScript`— I don't have a strong opinion about following that convention though.

mu-plugins/blocks/time/src/convert_times.js
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		return false;
+	} )();
+
+	document.querySelectorAll( '.wporg-time' ).forEach( ( dateElement ) => {
+		let localTime = '';
+		const datetime = dateElement.getAttribute( 'datetime' );
+		const datetimeISO = dateElement.getAttribute( 'data-iso' );
+		const date = datetime && parseDate( datetime );
```
ryelle: This was coming up null for me with a valid time: 2023-07-12T19:00:00-04:00

Although, I think that timestamp is wrong, it should either be `2023-07-12T19:00:00-00:00` or `2023-07-12T15:00:00-04:00`

ryelle (commented):
The editor side looks great, but it's not working on the frontend, and I think something's off with how the time is converted. I entered "Wednesday July 12 2023 at 15:00 EDT", but the datetime attribute is using the UTC value plus the timezone (`2023-07-12T19:00:00-04:00`).

mu-plugins/blocks/time/src/convert_times.js
adamwoodnz: Renamed

adamwoodnz:
> I entered "Wednesday July 12 2023 at 15:00 EDT", but the datetime attribute is using the UTC value plus the timezone (`2023-07-12T19:00:00-04:00`).

Hmm I hadn't tested with anything other than strings with UTC, as that's what I'd seen in posts. I'll take a look, ideally we could support more than that.

mu-plugins/blocks/time/src/convert_times.js
adamwoodnz: Looks like the regex didn't allow negative offsets, which didn't affect me with `+12`. Can you please try again?

adamwoodnz: I've changed my mac to `America/New_York` and with the date string 'Wednesday July 12 2023 at 15:00 EDT' I now see

Screenshot 2023-07-24 at 9 25 44 PM

While with my local timezone I see

Screenshot 2023-07-25 at 11 52 22 AM

adamwoodnz: This seems consistent with the shortcode version on this draft https://make.wordpress.org/core/?p=106234&preview=1&_ppp=5aa5224312

mu-plugins/blocks/time/src/index.js
```diff
@@ -0,0 +1,86 @@
...
+				);
+			} }
+			isActive={ isActive }
+		/>
+	);
+};
+
+registerFormatType( name, {
+	title: title,
+	tagName: 'time',
```
adamwoodnz, line 133: FYI I got accessibility team feedback on the old `<abbr title"...">` format and there was general agreement that this `time` element would be better, see [comment on alternative PR](https://github.com/WordPress/wporg-mu-plugins/pull/414#discussion_r1272964500).

adamwoodnz:
For some reason this isn't working in my sandbox. The editor script loads and the format appears to be applied (inspecting the element in the editor shows the format), but when I save and load the frontend there is no format applied, and if I reload the editor the format is gone too. It's as if it's being stripped by some tag filter... @ryelle @tellyworth any ideas?

ryelle:
> It's as if it's being stripped by some tag filter...

Ahh, I don't see `time` in [the $allowedposttags list](https://github.com/WordPress/wordpress-develop/blob/bea4307daa21a3f97d159898b0ac676332e0e7de/src/wp-includes/kses.php#L68), so I think it _is_ being stripped out. It's not a problem on single sites because those allow unfiltered_html for admins/editors. I think you can add `time` with [the wp_kses_allowed_html filter](https://developer.wordpress.org/reference/functions/wp_kses_allowed_html/).

adamwoodnz:
> I think you can add `time` with [the wp_kses_allowed_html filter](https://developer.wordpress.org/reference/functions/wp_kses_allowed_html/)

Thanks @ryelle [this seems to work well](https://github.com/WordPress/wporg-mu-plugins/pull/415/commits/3046990443531eeeae98cf7d50e6a06b2968e9d5)

adamwoodnz:
> I think you can add `time` with [the wp_kses_allowed_html filter](https://developer.wordpress.org/reference/functions/wp_kses_allowed_html/)

Thanks @ryelle [this seems to work well](https://github.com/WordPress/wporg-mu-plugins/pull/415/commits/3046990443531eeeae98cf7d50e6a06b2968e9d5)

ryelle:
Front end view works now. I tried a few different formats, some copied from make/core, and it all worked well  👍🏻 

I know you mentioned you can't reparse an updated date, but is it possible to auto-clear the format when the content changes? I think that's going to be confusing. I tried editing the day in a date and it turned into this:

| Editor | Frontend |
|---|---|
|  | Screenshot 2023-07-26 at 5 49 23 PM |

adamwoodnz:
> Front end view works now. I tried a few different formats, some copied from make/core, and it all worked well 👍🏻

🎉 

> is it possible to auto-clear the format when the content changes?

Nice idea, I'll give it a try

mu-plugins/blocks/time/postcss/style.pcss
```diff
@@ -0,0 +1,3 @@
+.wporg-time {
+	text-decoration: underline;
```
adamwoodnz, line 2: I find it's useful to be able to see the content the format has been applied to, and this seems the most appropriate style given that it will end up as a link on the frontend.

mu-plugins/blocks/time/postcss/style.pcss
```diff
@@ -0,0 +1,3 @@
+.wporg-time {
+	text-decoration: underline dotted;
```
adamwoodnz: I find it's useful to be able to see the content the format has been applied to, and this seems the most appropriate style given that it will end up as a link on the frontend.

adamwoodnz:
> is it possible to auto-clear the format when the content changes?

@ryelle I've implemented this now, please try again.

mu-plugins/blocks/time/postcss/style.pcss
StevenDufresne: The block editor content is rendered in an iframe. I don't think this will work.

StevenDufresne:
I don't know if this is a valid use case, but I got some weird behavior:
1. Successfully create a date,
2. Change `UTC` to `GMT` to simulate a user error
3. Notice you get double dates.

Screen Capture on 2023-08-07 at 09-32-53

adamwoodnz:
> I got some weird behavior

Thanks for this testing. This is intermittent for me. I did manage to repro once, but after refreshing I can't repro again. I think we can merge and monitor.

mu-plugins/blocks/time/postcss/style.pcss
adamwoodnz: Does it not show for you?

adamwoodnz: I see it in the post editor and page editor, but not in the site editor. In fact there are issues with the Gutenberg selectors in the site editor too. I've opened a follow up [issue](https://github.com/WordPress/wporg-mu-plugins/issues/422).

adamwoodnz: I've added partial support in the Site Editor by using inline styles https://github.com/WordPress/wporg-mu-plugins/pull/415/commits/ea1dad6132df73bf94f494d31bf59dee379b5267

I'll do further work in the follow up issue once I have a solution.

composer.json
```diff
@@ -28,7 +28,7 @@
 		"wp-coding-standards/wpcs": "2.*",
 		"wp-phpunit/wp-phpunit": "~6.0",
 		"wporg/wporg-repo-tools": "dev-trunk",
-		"yoast/phpunit-polyfills": "^1.0"
+		"yoast/phpunit-polyfills": "^1.1"
```
adamwoodnz, line 31: The [PHP unit tests failed](https://github.com/WordPress/wporg-mu-plugins/actions/runs/5932617325) with `Error: Version mismatch detected for the PHPUnit Polyfills. Please ensure that PHPUnit Polyfills 1.1.0 or higher is loaded. Found version: 1.0.5` so I've bumped inline with that

adamwoodnz: Seems like a very new release https://github.com/Yoast/PHPUnit-Polyfills/releases

dd32: This seems like a safe change to me.  Seems weird that it'd detect the out-of-dateness...

PR #441, opened 2023-09-05
New block: Query filter

### LINKED ISSUE #419: Nav component: Filter
As part of the nav components used across the site, I have been testing ideas for the filter flow and ended up with a version that feels pretty solid and flexible to be used in all subsites. Before describing its parts, here are two prototype recordings showing a common used case based on Showcase designs ([wporg-showcase-2022#130](https://github.com/WordPress/wporg-showcase-2022/issues/130))

### Desktop
https://github.com/WordPress/wporg-main-2022/assets/895819/dbdf629e-5de6-4dff-a992-8b3ca1408d09

### Mobile

https://github.com/WordPress/wporg-main-2022/assets/895819/7f07f047-0b30-416d-ab0e-960cb322e71a

## Description

The core approach of this idea is browsing and applying filters per filter type. Other ideas I handled relied on a toggle displaying all filters and a single action for submitting all criteria, but this design was preferred due to its cleanest and transparency by showing up the available filters.

For pages where applying filters take you to a search results page, like in Showcase, the content area adds a result and sort bar above it. Otherwise, both bars are visible in the page. This idea aims to simplify the amount of pages and ways of searching and browsing content, so I envision that its feasibility will be revisited per subsite.

All subcomponents are 40px height and can be found in the `Filters 6.1` section of the [Filter component file](https://www.figma.com/file/82KoxsTqEYlcCzHoJCKPlt/Filter-component?type=design&node-id=196-7522&mode=design&t=s07eokbmWD9HVaAQ-4). They will be mark as ready for dev once we reach agreement.

UI components of Filter componet with all their states

ryelle:
See #419 — this adds a new block called "Query Filter", which renders a dropdown form using values passed in via filter. When "Apply" is clicked, the form is submitted, and the resulting URL is expected to handle the applied filters. In many cases, wp_query can handle combined filters without any customization (ex, taxonomies). For non-core filters, hooking into `pre_get_posts` (+ maybe `query_vars`?) in the child theme should be enough.

This block is using the [Interactivity API](https://github.com/WordPress/gutenberg/tree/trunk/packages/interactivity), which uses `data-wp-` attributes to attach events and conditional attributes. This API is still experimental, but I took a lot of inspiration from [the navigation block](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/navigation).

By itself, this block doesn't render anything. It expects a `key` attribute, which is used to create a specific filter for the individual sites to populate a settings object. For example, the following would be used to create an "example" filter.

```html
<!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"right"}} -->
<div class="wp-block-group">
    <!-- wp:wporg/query-filter {"key":"example"} /-->
</div>
<!-- /wp:group -->
```

This creates the WP filter hook `wporg_query_filter_options_example`, which expects an object used to render the options. There's [more documentation inline in render.php](https://github.com/WordPress/wporg-mu-plugins/blob/0f2c391b8d8ff47c04b5233dee8e76cd2d7bd810/mu-plugins/blocks/query-filter/render.php#L10-L24).

To create a simple dropdown with 3 options:

```php
add_filter(
	'wporg_query_filter_options_example',
	function() {
		return array(
			'label' => 'Example <span>1</span>',
			'key' => 'q',
			'action' => home_url(),
			'options' => array(
				'opt-1' => 'Option 1',
				'opt-2' => 'Option 2',
				'opt-3' => 'Option 3',
			),
			'selected' => array( 'opt-2' ),
		);
	}
);
```

The output would look like this:

Screenshot 2023-09-05 at 6 51 56 PM

When "Apply" is clicked, the page loads `/?q[]=opt-2`. It's up to each site to load the correct archive.

❓  I think this approach will work well for the child theme setup, but I definitely want feedback on it before committing us all to this for all the directories 🙂 

🖌️  I've also used block-specific custom properties so that other themes can easily change colors with `theme.json` changes (here's an example in showcase, though i removed it once everything was set up since the defaults work). This is in case the theme directory needs a dark scheme.

**Practical example: Showcase**

I've got this working with real filters in https://github.com/WordPress/wporg-showcase-2022/pull/147 — here are some screen recordings of everything in place:

A basic filtering, starting on the homepage, shows all 3 filters & options. When a filter is applied, the results load using the archive template.

https://github.com/WordPress/wporg-mu-plugins/assets/541093/46b6268d-23b8-429a-976f-99af1fe7637e

More filtering interactions — clearing filters using "Clear" and by clicking the option; applying multiple filters, clicking out before applying.

https://github.com/WordPress/wporg-mu-plugins/assets/541093/a231e149-e4b9-40df-b6b7-1d1578bec859

Search and filters can work together — once on a search results page, filtering will filter the search results. (the opposite is not true, if you search on a filtered page it will remove the filters, we talked through that and it was expected behavior)

https://github.com/WordPress/wporg-mu-plugins/assets/541093/277c686f-efb8-4d03-856a-48fdebfb30a9

When clicking on a tag archive page, the filter for that tag is correctly selected.

https://github.com/WordPress/wporg-mu-plugins/assets/541093/54106695-e564-4c32-82d1-6ba447993cba

And on mobile, the dropdown switches to a modal

Screen Shot 2023-09-05 at 18 37 24


mu-plugins/blocks/query-filter/render.php
```diff
@@ -0,0 +1,127 @@
...
+			<div class="wporg-query-filter__modal-actions">
+				<input
+					type="button"
+					class="wporg-query-filter__modal-action-clear"
+					value="<?php esc_attr_e( 'Clear', 'wporg' ); ?>"
+					data-wp-on--click="actions.wporg.queryFilter.clearSelection"
+				/>
+				<input
+					type="submit"
+					value="<?php esc_html_e( 'Apply', 'wporg' ); ?>"
```
StevenDufresne, line 122: I think it was agreed that this would be updated with the filter count:

[Context Link](https://github.com/WordPress/wporg-mu-plugins/issues/419#issuecomment-1679390461)
[Figma Link](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=2847-4389&mode=design&t=Y2B54ZHq9soTX1nE-0)

| Image |
| --- |
| Screenshot 2023-09-06 at 10 22 42 AM |

mu-plugins/blocks/query-filter/postcss/style.pcss
```diff
@@ -0,0 +1,314 @@
...
+		background-color: var(--wp--custom--button--active--color--background);
+	}
+
+	&.wporg-query-filter__modal-action-clear {
+		background-color: transparent;
+		color: #a13f2a; /* pomegrade-1 fails contrast against white. */
+		font-weight: 400;
+
+		&:hover,
+		&:active {
```
StevenDufresne, line 201: Should this have a `:focus` state?

StevenDufresne:
I tested it locally, and while I don't yet have the expertise to review the interactivity implementation, it does work well and updates the query string appropriately. 👍

mu-plugins/blocks/query-filter/render.php
```diff
@@ -0,0 +1,127 @@
...
+
+// Initial state to pass to Interactivity API.
+$init_state = [
+	'isOpen' => false,
+	'hasHover' => false,
+];
+$encoded_state = wp_json_encode( [ 'wporg' => [ 'queryFilter' => $init_state ] ] );
+
+// Set up a unique ID for this filter.
+$html_id = wp_unique_id( "filter-{$settings['key']}-" );
```
adamwoodnz, line 44: TIL

mu-plugins/blocks/query-filter/src/view.js
```diff
@@ -0,0 +1,136 @@
...
+
+function updateToggleLabel( store, count ) {
+	const { context } = store;
+	const toggle = context.wporg.queryFilter.toggleButton;
+	if ( ! toggle ) {
+		return;
+	}
+	toggle.querySelector( 'span' ).innerText = count;
+	if ( count ) {
+		toggle.classList.add( 'has-filter-applied' );
```
adamwoodnz: Doesn't look like `.has-filter-applied` is used for anything, maybe not required?

mu-plugins/blocks/query-filter/src/view.js
```diff
@@ -0,0 +1,136 @@
...
+							event.preventDefault();
+							context.wporg.queryFilter.firstFocusableElement.focus();
+						}
+					}
+				},
+				handleMouseEnter: ( { context } ) => {
+					context.wporg.queryFilter.hasHover = true;
+				},
+				handleMouseLeave: ( { context } ) => {
+					context.wporg.queryFilter.hasHover = null;
```
adamwoodnz: Is there a reason not to always keep this as a boolean?

mu-plugins/blocks/query-filter/render.php
ryelle: I've been only referencing the "filters 6.1" page in the Filter component file, thinking that was the final version 🙃 

The link doesn't work for me (opens the file but wrong page)… but I found it by searching through the other pages, so thanks for pointing it out.

@WordPress/meta-design — Can you confirm how this should behave? It sounds like the top number (on black, red highlight) would be the number of filters currently applied on the page, while the Apply button (on blue, yellow highlight) would be the number of filters currently selected— so it would stay "Popular tags 1" even if you select two more, but the button would be "Apply (3)"?

Screenshot 2023-09-06 at 10 50 06 AM

mu-plugins/blocks/query-filter/postcss/style.pcss
ryelle: It inherits the focus from the focus just above here, it doesn't need a custom focus style— but if design wants to create one I can change it.

Screenshot 2023-09-06 at 11 01 10 AM

mu-plugins/blocks/query-filter/postcss/style.pcss
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	&:active {
+		color: var(--wp--custom--button--active--color--text);
+		background-color: var(--wp--custom--button--active--color--background);
+	}
+
+	&.wporg-query-filter__modal-action-clear {
+		background-color: transparent;
+		color: #a13f2a; /* pomegrade-1 fails contrast against white. */
```
ryelle, line 197: @WordPress/meta-design I meant to flag this in the description— the mockup uses "Pomegrade" on white but that [fails to meet contrast requirements](https://webaim.org/resources/contrastchecker/?fcolor=E26F56&bcolor=FFFFFF). There's a note in the design library about darker versions of colors, but I didn't see any, so I've made this one up. If you'd rather something different, it just needs to have [a 4.5+ value against light-grey-2](https://webaim.org/resources/contrastchecker/?fcolor=A13F2A&bcolor=F6F6F6), since I'm using that as the hover color.

Screenshot 2023-09-06 at 11 02 36 AM

Hover:
Screenshot 2023-09-06 at 11 02 50 AM

mu-plugins/blocks/query-filter/src/view.js
ryelle: 👍🏻

mu-plugins/blocks/query-filter/src/view.js
ryelle: There wasn't, but I've removed this mouse behavior to fix some other glitches I had noticed https://github.com/WordPress/wporg-mu-plugins/pull/441/commits/4f610db003632cb489041ae76e07a4c585c238fa.

mu-plugins/blocks/query-filter/render.php
jasmussen: > It sounds like the top number (on black, red highlight) would be the number of filters currently applied on the page, while the Apply button (on blue, yellow highlight) would be the number of filters currently selected— so it would stay "Popular tags 1" even if you select two more, but the button would be "Apply (3)"?

Correct, but honestly I think we can remove the number on the Apply button for now, I'm not sure it adds clarity. If Francisco when he comes back from AFK has to correct me, I will buy him an apology milkshake at that time. How does this sound?

ryelle: I think it was an accessibility request, but maybe it would be easier to understand the UX once this is merged and live on the staging site. I'll leave this as-is (no number on button) for this initial pass, and ask for feedback once there's a live place to try it out.

fcoveram: @ryelle is correct. It was an accessibility request shared [here](https://github.com/WordPress/wporg-mu-plugins/issues/419#issuecomment-1679390461).

PR #444, opened 2023-09-06
Block: Query Filters: Enable single-select options in filter block

ryelle:
This is built on top of the filter block added in #441 — this adds a setting to enable single-select (radio button) behavior for the options, for filters that can't accept multiple values. For example, sorting — this can use the same toggle/dropdown behavior, but you can only sort by one value at a time (at least, that's all we'll allow).

Now, if a filter is set up with `multiple: false`, it will use this single-select version. For example:

```
<!-- wp:wporg/query-filter {"key":"sort","multiple":false} /-->
```

See https://github.com/WordPress/wporg-showcase-2022/pull/147

| | Screenshot |
|---|---|
| Closed, the filter looks like the other toggles, but doesn't flag that there's a selected element (page is currently sorted by title) | sort-closed |
| Opened, the dropdown looks pretty similar to the multi-select versions, but you can't clear an item by clicking it again (because it's a radio button behind the scenes, for a11y) — clear still works to clear the selection though. | sort-date |

**To test**

Try this with the showcase PR: https://github.com/WordPress/wporg-showcase-2022/pull/147

### DIFF
--- a/mu-plugins/blocks/query-filter/render.php
+++ b/mu-plugins/blocks/query-filter/render.php
@@ -40,8 +40,20 @@ $init_state = [
 ];
 $encoded_state = wp_json_encode( [ 'wporg' => [ 'queryFilter' => $init_state ] ] );
 
+$has_multiple = isset( $attributes['multiple'] ) && $attributes['multiple'];
+
 // Set up a unique ID for this filter.
 $html_id = wp_unique_id( "filter-{$settings['key']}-" );
+
+$button_classes = array_keys(
+	array_filter(
+		array(
+			'wporg-query-filter__toggle' => true,
+			'has-no-filter-applied' => ! count( $settings['selected'] ),
+			'is-single-select' => ! $has_multiple,
+		)
+	)
+);
 ?>
 <div
 	<?php echo get_block_wrapper_attributes(); // phpcs:ignore ?>
@@ -52,7 +64,7 @@ $html_id = wp_unique_id( "filter-{$settings['key']}-" );
 	data-wp-on--keydown="actions.wporg.queryFilter.handleKeydown"
 >
 	<button
-		class="wporg-query-filter__toggle <?php echo count( $settings['selected'] ) ? '' : 'has-no-filter-applied'; ?>"
+		class="<?php echo esc_attr( implode( ' ', $button_classes ) ); ?>"
 		data-wp-class--is-active="context.wporg.queryFilter.isOpen"
 		data-wp-on--click="actions.wporg.queryFilter.toggle"
 		data-wp-bind--aria-expanded="context.wporg.queryFilter.isOpen"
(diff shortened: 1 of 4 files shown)


mu-plugins/blocks/query-filter/render.php
```diff
@@ -38,8 +38,20 @@
 ];
 $encoded_state = wp_json_encode( [ 'wporg' => [ 'queryFilter' => $init_state ] ] );
 
+$has_multiple = isset( $attributes['multiple'] ) && $attributes['multiple'];
```
adamwoodnz, line 43: After this change the default will be single, is that likely to be the most common configuration?

The code ends up checking the inverse in several places, eg. `'is-single-select' => ! $has_multiple,` which makes me think maybe the attribute should be `single: boolean`, thereby simplifying the logic?

So we add the attribute based on the desired outcome rather than disabling the default.

adamwoodnz: Ah no I see the default is `multiple: true`, ignore that bit

ryelle: I was thinking about it like the `select` element's `multiple` attribute, but I could change it if this seems unclear?

adamwoodnz: Hmm hadn't considered that. Ok makes sense to me 👍

adamwoodnz:
Design/a11y question; should the current sort be output somewhere?

adamwoodnz:
This is working really nicely for me. There is one inconsistency I noticed; with multi select you move through the items with tab, versus single select it's the up and down arrows. It was a bit jarring for me moving from one to the other but maybe it's expected... is it intended?

ryelle:
> with multi select you move through the items with tab, versus single select it's the up and down arrows. 

Yeah, that's expected because of the checkbox vs radio implementation— you can tab through checkbox items, but radio options are a single tab-stop so that only one item is selected. I would prefer not to change that, since right now it's using default form behavior that we know works with a11y.

jasmussen:
>  Design/a11y question; should the current sort be output somewhere?

I'm still catching up after a meetup, but I think [this is the latest filter component](https://www.figma.com/file/82KoxsTqEYlcCzHoJCKPlt/Filter-component?type=design&node-id=196%3A7522&mode=design&t=BnbUH5bTLmKz6NLZ-1). And the design here shows the main filter category as the permanent fixture, but adds a number of selected filters from that taxonomy as a small number:

Screenshot 2023-09-11 at 10 47 10
Screenshot 2023-09-11 at 10 47 16

ryelle:
> And the design here shows the main filter category as the permanent fixture, but adds a number of selected filters from that taxonomy as a small number:

Yeah, that's how it's already working on the multiple-select PR (#441), this PR is for a single-selection dropdown like sorting— so if we used that it would always be `1`, which doesn't seem very useful.

My latest change in [the companion showcase PR](https://github.com/WordPress/wporg-showcase-2022/pull/147) just adds the current value to the label:

| Default | Title | Newest |
|---|---|---|
|  |  |  |

jasmussen:
This seems like a place where we land the simplest version, learn, and if need be, iterate.

PR #449, opened 2023-09-15
Block: Query Filter: Design & accessibility fixes

ryelle:
See #419 & https://github.com/WordPress/wporg-showcase-2022/issues/151. This PR addresses the design & accessibility feedback for the Query Filter block. 

- Fix the input position to fix scroll issues (tabbing to items off screen moves them into view, selecting items does not scroll the whole page).
- Add the currently selected count to the Apply button when it's a multiple-select dropdown.
- Updates the font size, line height, and padding across the buttons and options to match the design.
- Add a hover state to options, add border radius to all option states (hover, checked, focus).
- Move dropdown down 10px.
- Update the Clear button to be `aria-disabled` when no items are selected, use that to style it as grey.
- When items are selected, Clear is now blueberry.

Thanks for the feedback @jasmussen & @joedolson!

## Screenshots

| | Screenshot |
|---|---|
| Multiple select, "e-commerce" selected, "Design" hovered | qf-selected-hover |
| Single select, none selected so Clear is disabled | qf-clear |

**To test**

Use this with Showcase to have something set up already.

No specific test instructions, just try interacting with the filters and make sure it matches the feedback & design.

### DIFF
--- a/mu-plugins/blocks/query-filter/postcss/style.pcss
+++ b/mu-plugins/blocks/query-filter/postcss/style.pcss
@@ -117,6 +117,7 @@
 
 .wporg-query-filter__option {
 	--wporg-query-filter--icon-size: 24px;
+	position: relative;
 }
 
 .wporg-query-filter__option input[type] {
@@ -128,6 +129,8 @@
 	overflow: hidden;
 	padding: 0;
 	position: absolute;
+	top: 0;
+	left: 0;
 	width: 1px;
 	word-wrap: normal !important;
 
--- a/mu-plugins/blocks/query-filter/postcss/style.pcss
+++ b/mu-plugins/blocks/query-filter/postcss/style.pcss
@@ -21,8 +21,8 @@
 	border-radius: 2px;
 	color: var(--wp--custom--wporg-query-filters--toggle--color--text);
 	font-family: var(--wp--preset--font-family--inter);
-	font-size: var(--wp--custom--button--typography--font-size);
-	line-height: var(--wp--custom--button--typography--line-height);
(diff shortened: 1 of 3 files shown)


StevenDufresne:
I had trouble testing this. Something with my local environment. The code changes look fine though.

Would it be possible to contain scrolling on the window when we reach the end of the menu on mobile using `overscroll-behavior: contain`?

adamwoodnz:
Just a thought

> Update the Clear button to be aria-disabled when no items are selected

Should 'Apply' also be disabled when nothing is selected to apply?

jasmussen:
Looks good. 

Yes, if we can lock the scrolling below, especially on mobile, that would be great, I know this is a challenge.

ryelle:
> Should 'Apply' also be disabled when nothing is selected to apply?

Technically applying nothing is still valid — if you have a combination of filters, say, 2 tags and 1 flavor, you can clear the tags with Clear, then Apply to view just the selected flavor.

> Yes, if we can lock the scrolling below, especially on mobile, that would be great, I know this is a challenge.

I've prevented the scroll on mobile — scrolling on larger screens might still be useful to get the dropdown in a better position to read (I had disabled it everywhere, tested it, and immediately was annoyed I couldn't move the dropdown 😅 ).

ndiego:
> I've prevented the scroll on mobile — scrolling on larger screens might still be useful to get the dropdown in a better position to read (I had disabled it everywhere, tested it, and immediately was annoyed I couldn't move the dropdown 😅 ).

Yeah, locking on mobile is great, but allow scrolling on Desktop.

PR #450, opened 2023-09-19
Add external indicators and set .TV to lowercase in global footer

ndiego:
This PR adds external indicators to all links that take the user outside of .org and sets .TV to lowercase in WordPress.tv. Both changes are to ensure consistency with the global header navigation. 

**Before**

image

**After**

image


ndiego:
Closed in favor of https://github.com/WordPress/wporg-mu-plugins/pull/464

PR #454, opened 2023-09-25
Global Header/Footer: Hide the arrow emoji for screen reader users

### LINKED ISSUE #453: Prevent screen readers from reading the arrow emoji in menu items
See https://github.com/WordPress/wporg-parent-2021/pull/94, https://github.com/WordPress/wporg-main-2022/issues/311

It was requested that the arrow emojis be wrapped in `aria-hidden`, so that they're not announced to screen reader users. The code added there does not apply to the header & footer nav menus.

Since more of these will be introduced in #450, we should try to do the same for the nav items.

Example of unwrapped arrow:
Screenshot 2023-09-20 at 11 31 07 AM

ryelle:
Fixes #453 — The arrows are primarily decorative, and screen readers can be overly verbose when encountering special characters. This brings over the function from https://github.com/WordPress/wporg-parent-2021/pull/94 to wrap the arrows in `aria-hidden`.

Screenshot 2023-09-25 at 12 53 05 PM

**To test**

- View source/inspect a list item with an arrow
- It should be wrapped in a span

Or

- Try using a screen reader, navigate through to the menu items with arrows
- Only the text should be read, not the arrows


StevenDufresne (approved):
LGTM 👍

PR #455, opened 2023-09-25
Global Footer: Switch multiple nav blocks to lists inside a nav wrapper.

### LINKED ISSUE #402: Global footer navigation should not use five separate `nav` landmarks
The global footer navigation is currently structured as five separate navigation landmarks. This is only one navigation group, and should be in a single navigation menu.

Additionally, these navigation menus have no accessible name. As individual sections, there isn't really any reasonable name for these menus - they aren't distinct collections of information. But as a single navigation menu, 'Footer' would be a perfectly reasonable name for the menu.

This would significantly help in reducing aural clutter for screen readers, since there are currently 9 navigation landmarks in global scope.

ryelle:
Fixes #402 — This switches to the global footer block to use a `nav` group block around 5 lists, instead of 5 separate nav blocks. Additionally, the label "Footer" has been added to the `nav`, so it now appears as "Footer navigation" when browsing by landmark.

This also had to change the way menu items were localized, removing the filter on nav-link blocks in favor of directly calling the function on appropriate links. 

Props @joedolson for the report

**Screenshots**

| | Production | Branch |
|---|---|---|
| Home, wide screen |  |  |
| Home, small screen |  |  |
| Rosetta, wide screen |  |  |
| Rosetta, small screen |  |  |
| Classic theme, wide screen |  |  |
| Classic theme, small screen |  |  |

**To test**

1. If you can, apply this to a sandbox and test on various sites— new & old themes, rosetta, etc.
2. Otherwise, use with one of the redesign sites (main, showcase, etc) to test locally
3. The footer is now one single `nav` element, individual items are lists
4. There should be no visual changes in the footer

Optionally test with a screen reader to see the single Footer navigation landmark.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -810,7 +810,6 @@ function render_global_footer( $attributes, $content, $block ) {
 
 	if ( is_rosetta_site() ) {
 		$locale_title = get_rosetta_name();
-		add_filter( 'render_block_data', __NAMESPACE__ . '\localize_nav_links' );
 	} else {
 		$locale_title = '';
 	}
@@ -831,8 +830,6 @@ function render_global_footer( $attributes, $content, $block ) {
 		require_once __DIR__ . '/classic-footer.php';
 		$footer_markup = ob_get_clean();
 	}
-
-	remove_filter( 'render_block_data', __NAMESPACE__ . '\localize_nav_links' );
 
 	$wrapper_attributes = get_block_wrapper_attributes(
 		array( 'class' => 'global-footer wp-block-group' )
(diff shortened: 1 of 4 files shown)


StevenDufresne (approved):
Looks good. Noting that there's a non-consequential `padding-right`.

```
.wp-block-group.global-footer>*, .wp-block-group.global-header {
    padding-right: var(--wp--custom--alignment--scroll-bar-width);
}
```

Screenshot 2023-09-27 at 2 31 10 PM

ryelle:
Good catch, that doesn't seem to be necessary since the footer items are wrapped in a group with side-padding. I've removed it in https://github.com/WordPress/wporg-mu-plugins/pull/455/commits/3d3b06a36266f20ad38f48f1c543013b694e0db5.

PR #458, opened 2023-09-26
Global header: Update navigation menu items

### LINKED ISSUE #316: Add Showcase as a top-level navigation item

### LINKED ISSUE #360: Navigation: Rephrase a few items, reconsider sorting

> Certain instances are verbally obscure. Revise naming conventions (ie: Mobile versus Get Apps or WordPress Apps).

For reference, this is the navigation on desktop and mobile:

Screenshot 2023-03-28 at 10 19 06

Screenshot 2023-03-28 at 10 19 21

Outside of some separate navigation considerations such as WordPress/wporg-mu-plugins#370 and WordPress/wporg-mu-plugins#359, it's worth revisiting naming conventions. As a baseline, we can do the main suggested change:

[ ] Change "Mobile" to "Get Apps" or "WordPress Apps"

CC: @thetinyl in case you have suggestions.

## Download

From the same issue:

> There’s duplication of nav instances (ie: Download&Extend, Get WP, and Get WP), while the footer doesn’t have it. Worth revising through the IA exercise.

Extracting from that:

* we have in the navigation both the blue Get WordPress button, and the menu item called "Get WordPress"
* the top menu item is also called "Download & Extend"
* a download link is not present in the footer: should it be?

One basis suggestion could be to rename the Download & Extend section to just "Extend", remove the "Get WordPress" link inside, and default to taking you to themes. This would also solve [wporg-mu-plugins/issues/358](https://github.com/WordPress/wporg-mu-plugins/issues/358). 

extend

Additinoally, there are two community image sources: Openverse and the Photo Directory. The former is in Extend, the latter is in Community. It seems like these two should be in the same dropdown, either Openverse or Community. 

extend menu

Community

Any thoughts?

Which links to include in the footer is worth revisiting the heuristic for. It seems very likely that there's prior knowledge here, since the footer includes some links that are omitted from the header navigation, so some curation here that considers previous conversations might be good. 

CC: @fcoveram

### LINKED ISSUE #417: Add "Blocks" page to Extend menu
As work continues to add a Blocks page https://github.com/WordPress/wporg-main-2022/issues/245, I wanted to kick off an issue for adding it under the Extend menu once it's live. cc @jasmussen and @ndiego.

### LINKED ISSUE #440: Navigation: Make primary navigation submenus "Open on click"
This issue was extracted from the [larger discussion](https://github.com/WordPress/wporg-mu-plugins/issues/360) about updating the primary navigation on WordPress.org. 

The proposal is for the submenus to only open on click. Currently, the submenus open on hover, and the first item in the submenu is the same link as the top-level menu item. Switching to open on click will simplify the menu. 

| Current | Proposed (not real wp.org site) |
| --- | --- |
| submenu-hover | menu demo |

"Open on click" is a setting in the Navigation block, so no additional development work should be needed.  

image

ryelle:
Fixes #316, fixes #360, fixes #440, fixes #417. This updates the menu structure to match the layout in this comment: https://github.com/WordPress/wporg-mu-plugins/issues/360#issuecomment-1732619788

Additionally, this changes the menu behavior to "open on click", rather than hover, which avoids the duplicate menu items. Changing this required some more CSS and JS changes than expected, so it would be good to get another  👀  on that just in case.

There should be no major visual changes, the main visible difference is that now the focus goes around both the title and dropdown icon.

Screenshot 2023-09-26 at 6 25 27 PM

| | Screenshot |
|---|---|
| Showcase, no menus open | new-header-showcase |
| Showcase, small screens | new-header-mobile |
| Blue header, with a menu open | new-header-blue |

**To test**

- Try on a sandbox or with one of the redesign projects
- The menu items should match the issue comment
- Items with submenus should have chevron icons
- Hovering over items will highlight them but not open submenus
- Clicking an item with a submenu should open the menu, not change page navigation
- Non-submenu items should behave as expected too

Try on Rosetta sites if you can sandbox.

> [!WARNING]
> This should not be merged until Showcase is ready to be moved to the top-level, once the new theme launches.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -15,7 +15,6 @@ add_filter( 'wp_enqueue_scripts', __NAMESPACE__ . '\register_block_assets', 200 
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_compat_wp4_styles', 5 ); // Before any theme CSS.
 add_action( 'wp_head', __NAMESPACE__ . '\preload_google_fonts' );
 add_filter( 'style_loader_src', __NAMESPACE__ . '\update_google_fonts_url', 10, 2 );
-add_filter( 'render_block_core/navigation-link', __NAMESPACE__ . '\swap_submenu_arrow_svg' );
 add_filter( 'render_block_core/search', __NAMESPACE__ . '\swap_header_search_action', 10, 2 );
 add_filter( 'render_block_data', __NAMESPACE__ . '\update_block_style_colors' );
 
@@ -465,7 +464,7 @@ function get_global_menu_items() {
 		),
 		array(
 			'title'   => esc_html_x( 'Learn', 'Menu item title', 'wporg' ),
-			'url'     => 'https://learn.wordpress.org/',
+			'url'     => '#',
 			'type'    => 'custom',
 			'submenu' => array(
 				array(
@@ -497,7 +496,7 @@ function get_global_menu_items() {
 		),
 		array(
 			'title'   => esc_html_x( 'Community', 'Menu item title', 'wporg' ),
-			'url'     => 'https://make.wordpress.org/',
+			'url'     => '#',
 			'type'    => 'custom',
 			'submenu' => array(
 				array(
@@ -534,7 +533,7 @@ function get_global_menu_items() {
 		),
 		array(
 			'title'   => esc_html_x( 'About', 'Menu item title', 'wporg' ),
-			'url'     => 'https://wordpress.org/about/',
+			'url'     => '#',
 			'type'    => 'custom',
 			'submenu' => array(
 				array(
@@ -962,16 +961,6 @@ function set_current_item_class( $menu_items ) {
 	}
 
 	return $menu_items;
-}
-
-/**
- * Replace the current submenu down-arrow with a custom icon.
- *
- * @param string $block_content The block content about to be appended.
- * @return string The filtered block content.
- */
-function swap_submenu_arrow_svg( $block_content ) {
-	return str_replace( block_core_navigation_link_render_submenu_icon(), "<svg width='10' height='7' viewBox='0 0 10 7' stroke-width='1.2' xmlns='http://www.w3.org/2000/svg'><path d='M0.416667 1.33325L5 5.49992L9.58331 1.33325'></path></svg>", $block_content );
 }
 
 /**
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -422,23 +422,23 @@ function get_global_menu_items() {
 			'type'    => 'custom',
 		),
 		array(
-			'title'   => esc_html_x( 'Download & Extend', 'Menu item title', 'wporg' ),
-			'url'     => 'https://wordpress.org/download/',
+			'title' => esc_html_x( 'Showcase', 'Menu item title', 'wporg' ),
+			'url'   => 'https://wordpress.org/showcase/',
+			'type'  => 'custom',
+		),
+		array(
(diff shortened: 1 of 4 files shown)


adamwoodnz (commented):
👍 Working nicely for me on Developer.

I noticed that when tabbing with my keyboard at mobile size I can focus the submenu items, eg. Extend, but can't act upon them, maybe an accessibility concern. I know it's not usual user behaviour but maybe mobile assistive technologies will focus things in a similar manner.

ndiego:
From the screenshots, this is looking great! I will need to figure out how to set up a testing environment 😅

ndiego:
> I noticed that when tabbing with my keyboard at mobile size I can focus the submenu items, eg. Extend, but can't act upon them, maybe an accessibility concern. I know it's not usual user behaviour but maybe mobile assistive technologies will focus things in a similar manner.

In testing with the Navigation block, if you don't have a URL set for nav links with a submenu, tabbing does not focus on the nav links (in this case Extend, etc.). So I think if you remove the # and set no URL at all ([here for example](https://github.com/WordPress/wporg-mu-plugins/pull/458/files#diff-460e50b82495cab49b3ce3a702572bec919fae2cbec691937f8abbe7de6e2c33R435)), the tabbing should work as expected.

outdoor2kode (approved):
This looks like a great enhancement! 👍 

A few notes:

1. Tickets that were mentioned have been tested and appear to have no issues.
2. The [comment](https://github.com/WordPress/wporg-mu-plugins/issues/360#issuecomment-1708233178) in #360 doesn't seem to have reached a conclusion yet. The ticket might need to remain open or the discussion could be moved to a new ticket after this PR gets merged.
3. For this [commit](https://github.com/WordPress/wporg-mu-plugins/pull/458/commits/c40bf8211eb59ae15b4a8148dada1a2cb962b4e1), I've noticed the ellipsis on es.wp.org, and it works as expected.
4. Regarding Nick's [comment](https://github.com/WordPress/wporg-mu-plugins/pull/458#issuecomment-1736549616), whether I set the URL to an empty string or delete the entire line, both tabbing and screen readers still focus on the item. Maybe I've mistaken the context or suggestion. It also doesn't seem to work by removing the URL [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/global-header-footer/header.php#L50) in the `header.php` directly.

ndiego:
> Regarding Nick's https://github.com/WordPress/wporg-mu-plugins/pull/458#issuecomment-1736549616, whether I set the URL to an empty string or delete the entire line, both tabbing and screen readers still focus on the item. Maybe I've mistaken the context or suggestion. It also doesn't seem to work by removing the URL [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/global-header-footer/header.php#L50) in the header.php directly.

Looks like I was wrong. Sorry for the confusion.

ndiego (approved):

ryelle:
> I noticed that when tabbing with my keyboard at mobile size I can focus the submenu items, eg. Extend, but can't act upon them, maybe an accessibility concern.

Yeah, it looks like this is a gutenbug— the mobile menu is intentionally expanded, but the semantics on the submenu toggle buttons make it seem like it's still possible to expand/collapse them. `aria-expanded` is initially false, and toggled on click, despite the menu always being expanded.

> The [comment](https://github.com/WordPress/wporg-mu-plugins/issues/360#issuecomment-1708233178) in #360 doesn't seem to have reached a conclusion yet. The ticket might need to remain open or the discussion could be moved to a new ticket after this PR gets merged.

I think the resolution there was that Hosting should be a top-level menu item, so the discussion of a local menu on Download is not necessary.

ndiego:
> I think the resolution there was that Hosting should be a top-level menu item, so the discussion of a local menu on Download is not necessary.

Yeah, adding Hosting back as a top-level menu item makes that comment mute.

PR #473, opened 2023-10-12
Query filters: Fix scrollbar overlaying dropdown menu

### LINKED ISSUE #467: Query Filter: On mobile, horizontal scroll bar is visible on page load.
On my phone, I get the horizontal scrollbar for the list of filters when i load the page. If I interact with the menu and stop, it will fade away as expected. It isn't a transitional state since it will persist. It doesn't happen on all devices (tested using dev tool), but it does with my personal phone and I can reproduce using IPhone SE via dev tools.

## On Page Load
Screenshot 2023-10-03 at 12 35 24 PM

outdoor2kode:
Fixes #467 (To be specific, https://github.com/WordPress/wporg-mu-plugins/issues/467#issuecomment-1754349676)

Related: https://github.com/WordPress/wporg-showcase-2022/pull/248

This PR adds a toggle function for `overflow-x`. When a theme assigns `overflow-x: scroll` to the query filter, change it to `overflow-x: hidden` when the dropdown menu is opened. And if it's closed, change it back to `scroll`.

Approach here was employed due to the overlapping issue can't be easily resolved just by adjusting the `z-index` or `position` of the query filter and the dropdown menu for a few potential reasons:
1. Query Filter is the parent of Dropdown Menu.
2. Dropdown menu uses position: fixed.
3. Behavior of the Scrollbar is determined by browser or OS.

#### Screenshots
| Before | After |
| ------ | -----|
| before | after |

### DIFF
--- a/mu-plugins/blocks/query-filter/src/view.js
+++ b/mu-plugins/blocks/query-filter/src/view.js
@@ -31,9 +31,9 @@ function toggleOverflowX() {
 	if ( filtersElement ) {
 		const currentOverflowX = window.getComputedStyle( filtersElement ).overflowX;
 
-		if ( currentOverflowX === 'hidden' ) {
+		if ( 'hidden' === currentOverflowX ) {
 			filtersElement.style.overflowX = 'scroll';
-		} else if ( currentOverflowX === 'scroll' ) {
+		} else if ( 'scroll' === currentOverflowX || 'auto' === currentOverflowX ) {
 			filtersElement.style.overflowX = 'hidden';
 		}
 	}
--- a/mu-plugins/blocks/query-filter/src/view.js
+++ b/mu-plugins/blocks/query-filter/src/view.js
@@ -15,6 +15,30 @@ const focusableSelectors = [
 	'[tabindex]:not([tabindex^="-"])',
 ];
 
+/**
+ * Toggles the overflow-x style of the query filter between 'hidden' and 'scroll'.
+ *
+ * In certain themes (e.g., showcase), an 'overflow-x: scroll' is added on mobile screens to always display
+ * the horizontal scrollbar, indicating to users that there's more content to the right.
+ * However, this persistent display feature causes the dropdown menu to be overlaid by the scrollbar
+ * when opened (See issue https://github.com/WordPress/wporg-mu-plugins/issues/467#issuecomment-1754349676).
+ * This function serves to address that issue.
+ *
+ */
+function toggleOverflowX() {
+	const filtersElement = document.querySelector( '.wporg-query-filters' );
+
+	if ( filtersElement ) {
+		const currentOverflowX = window.getComputedStyle( filtersElement ).overflowX;
+
+		if ( currentOverflowX === 'hidden' ) {
+			filtersElement.style.overflowX = 'scroll';
+		} else if ( currentOverflowX === 'scroll' ) {
+			filtersElement.style.overflowX = 'hidden';
+		}
+	}
+}
+
 function closeDropdown( store ) {
 	const { context } = store;
 	context.wporg.queryFilter.isOpen = false;
@@ -23,6 +47,8 @@ function closeDropdown( store ) {
 	const count = context.wporg.queryFilter.form?.querySelectorAll( 'input:checked' ).length;
 	updateButtons( store, count );
 	document.documentElement.classList.remove( 'is-query-filter-open' );
+
+	toggleOverflowX();
 }
 
 function updateButtons( store, count ) {
@@ -58,6 +84,7 @@ wpStore( {
 					} else {
 						context.wporg.queryFilter.isOpen = true;
 						document.documentElement.classList.add( 'is-query-filter-open' );
+						toggleOverflowX();
 					}
 				},
 				handleKeydown: ( store ) => {


outdoor2kode:
Context:
When on a mobile screen, due to the horizontal overflow, a persistent scrollbar has been set up via CSS to inform users there's more content to the right. This configuration led to the issue mentioned in this PR: the scrollbar overlays the dropdown menu.

Current Situation:
This PR addresses the overlapping scrollbar issue. However,

1. Steve mentioned [here](https://github.com/WordPress/wporg-mu-plugins/issues/467#issuecomment-1744194955) that always displaying the scrollbar looks a bit odd visually. I think the length of the scrollbar might also contribute to this awkward appearance.
image

2. On my Mac using Chrome and resizing the window, I can always see the scrollbar. But when I switch to my phone (iPhone 12), the scrollbar disappears completely. As a result, on the Archive page, one would be completely unaware of the existence of the Sort filter. This can be tested on showcase-v2.

I guess the simplest solution here is to let the filters wrap, as shown in the video below. This method also addresses the issues raised in this PR. 
A more complex solution might be rethinking the arrangement of the filters or thinking about ways to hint to users that there's more content to the right.

Thoughts on this? @jasmussen @marko-srb 

https://github.com/WordPress/wporg-mu-plugins/assets/18050944/c484931d-0aff-4241-9483-277752020f2a

mu-plugins/blocks/query-filter/src/view.js
```diff
@@ -23,6 +47,8 @@ function closeDropdown( store ) {
 	const count = context.wporg.queryFilter.form?.querySelectorAll( 'input:checked' ).length;
 	updateButtons( store, count );
 	document.documentElement.classList.remove( 'is-query-filter-open' );
+
+	toggleOverflowX();
```
adamwoodnz, line 51: I think we can remove this and just call the function outside of the `if` conditions on `toggle`

mu-plugins/blocks/query-filter/src/view.js
```diff
@@ -58,6 +84,7 @@ wpStore( {
 					} else {
 						context.wporg.queryFilter.isOpen = true;
 						document.documentElement.classList.add( 'is-query-filter-open' );
+						toggleOverflowX();
 					}
```
adamwoodnz, line 88: It seems like we can just call this, rather than doing it for each of open and close.

```suggestion
					}
					
					toggleOverflowX();
```

mu-plugins/blocks/query-filter/src/view.js
```diff
@@ -15,6 +15,30 @@ const focusableSelectors = [
...
+ */
+function toggleOverflowX() {
+	const filtersElement = document.querySelector( '.wporg-query-filters' );
+
+	if ( filtersElement ) {
+		const currentOverflowX = window.getComputedStyle( filtersElement ).overflowX;
+
+		if ( currentOverflowX === 'hidden' ) {
+			filtersElement.style.overflowX = 'scroll';
+		} else if ( currentOverflowX === 'scroll' ) {
```
adamwoodnz: For me the default state isn't `scroll`, but `auto`. Maybe we can do it for either?
```suggestion
		} else if ( 'scroll' === currentOverflowX || 'auto' ===  currentOverflowX ) {
```

adamwoodnz (changes requested):
👍 Works for me, with a small change 🙂

mu-plugins/blocks/query-filter/src/view.js
adamwoodnz: Ah maybe this PR depends on `overflow-x` being set in this [PR](https://github.com/WordPress/wporg-showcase-2022/pull/248)

adamwoodnz:
> Thoughts on this? @jasmussen @marko-srb

Or @WordPress/meta-design

jasmussen:
I _think_ @fcoveram's intent with the filter bar design he worked on was for them to scroll horizontally, and if that is the case, this z-index fix seems good. But I'd love if he had time to take a look. Depending on that feedback, letting the dropdowns stack seems acceptable to me as well.

StevenDufresne:
I think the design works fine, but we could revisit the type of affordance. 

Youtube comes to mind as an example:
Screenshot 2023-10-16 at 12 05 27 PM

fcoveram:
@jasmussen is right about the horizontal intention.

To give more context, the idea of showing [search and filter actions in two bars](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=2847-3418&mode=design&t=X9vGtKNRTMLYeS3l-4) began with the need for displaying the number of results by a11y reasons and placing apart the sort action, as the latter is not a filter. In most cases, especially in the _index family_'s landing page (Showcase, Themes, Patterns, and Plugins), the page is a curated list of items, whereas the search result/archive allows users to play with filters and sorting. Therefore, outlining these two use cases reinforced the context difference.

Since the layout changed in [wporg-showcase-2022#167](https://github.com/WordPress/wporg-showcase-2022/issues/167), I'm drawn to follow @StevenDufresne's reference. I thought about it during the design, yet the sort action will be placed at the end of the list in a filter-actions context. 

If we consider the two bars idea valid, I'm happy to create a ticket to address it.

mu-plugins/blocks/query-filter/src/view.js
outdoor2kode: I think we might need to keep this as it is, otherwise the [closeDropdown](https://github.com/WordPress/wporg-mu-plugins/blob/262f0816fc6a4b7f271afcf3dd0e66fe2743447c/mu-plugins/blocks/query-filter/src/view.js#L94) within `handleKeydown` won't work, resulting in the filter list being unscrollable after pressing Esc.

mu-plugins/blocks/query-filter/src/view.js
outdoor2kode: Yeah, you're right, that's because it's set to scroll over there. However, I think we can still add `auto` here. This way, even if someone changes it to `auto` later on that side, things won't break.

outdoor2kode:
@fcoveram thanks for the details. The two bars idea makes sense to me, I reckon we can follow Steve's example and this is supposed to be shipped before launching?

If not, the simpler solution (let the filters wrap) should be used first. This will only bring benefits (users with smaller screens can see the Sort option), with no downsides (the experience for users with larger screens remains unchanged).

mu-plugins/blocks/query-filter/src/view.js
adamwoodnz: Ah yes I see what you mean

adamwoodnz (approved):
LGTM in it's current state

outdoor2kode:
I'm going to merge this PR first as the original issue has been resolved. 
@fcoveram feel free to continue leaving your feedback here or open a new ticket for [this](https://github.com/WordPress/wporg-mu-plugins/pull/473#issuecomment-1765093515) if you prefer. 🙏

PR #484, opened 2023-10-19
Make top margin of sidebar container configurable via theme.json

### LINKED ISSUE #483: Sidebar Container: Make top margin configurable
For the redesigned Developer site the sidebar container needs to move down the page, to allow for the search bar moving out and into the content. The block is also used on the Documentation site where it doesn't need to shift (yet), so the margin needs to be configurable for different sites.

The margin is used in calculations for switching between sticky and static modes, so adding extra margin or padding to internal blocks isn't an option.

### Screenshots

| Developer | Documentation |
|-|-|
| Screenshot 2023-10-19 at 3 18 09 PM | Screenshot 2023-10-19 at 3 17 11 PM |

adamwoodnz:
Closes #483 

Makes the top margin of the sidebar container configurable, so that it can be changed for the redesigned Developer site.

Keeps the existing value as the default so that the Documentation site isn't affected.

### Testing

Easiest to use the [Developer PR](https://github.com/WordPress/wporg-developer/pull/299), see testing instructions there.

### DIFF
--- a/mu-plugins/blocks/sidebar-container/postcss/style.pcss
+++ b/mu-plugins/blocks/sidebar-container/postcss/style.pcss
@@ -21,7 +21,7 @@
 		/* Right offset should be "edge spacing" at minimum, otherwise calculate it to be centered. */
 		right: max(var(--wp--preset--spacing--edge-space), calc((100% - var(--wp--style--global--wide-size)) / 2));
 		width: var(--local--block-end-sidebar--width);
-		margin-top: var(--wp--preset--spacing--edge-space) !important;
+		margin-top: var(--wp--custom--wporg-sidebar-container--spacing--margin--top) !important;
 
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/sidebar-container/postcss/style.pcss
```diff
@@ -44,3 +44,8 @@
 		scroll-margin-top: var(--wp-local-header-offset, 0);
 	}
 }
+
+/* Set up the custom properties. These can be overridden by settings in theme.json. */
+:where(body) {
+	--wp--custom--wporg-sidebar-container--spacing--margin--top: var(--wp--preset--spacing--edge-space);
```
adamwoodnz, line 50: Default value set to the existing value so that the Documentation site doesn't require any changes

PR #489, opened 2023-10-25
Latest News block: Add support for cards style and popular categories

### LINKED ISSUE #448: Update Latest News block to support cards style and popular categories list
Required for the Developer redesign:

1. 'Cards' style variation
2. Optional 'Popular categories' list below post list

[Design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=2934%3A6794&mode=dev):

Screenshot 2023-09-14 at 12 40 27 PM

adamwoodnz:
Closes #448

Adds a Cards style variation and 'showCategories' attribute to the block to support the [design for Developer](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=2934%3A6794&mode=dev).

#### Editor

Screenshot 2023-10-25 at 2 41 29 PM

#### Frontend

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_latest-news-cards_(Desktop) (1) | localhost_8888_latest-news-cards_(iPad) | localhost_8888_latest-news-cards_(Samsung Galaxy S20 Ultra) |

#### Developer

localhost_8888_(Desktop) (9)

### Testing
1. Ensure you have posts and categories saved
2. Add a latest news block, set style to 'Cards' and enable 'Show Categories'
3. Block should display like design in editor and frontend
4. Regression test the default styles
5. Ensure most used categories are displayed in the popular categories list

### DIFF
--- a/mu-plugins/blocks/latest-news/latest-news.php
+++ b/mu-plugins/blocks/latest-news/latest-news.php
@@ -75,7 +75,7 @@ function render_block( $attributes ) {
 		if ( ! empty( $category ) ) {
 			if ( isset( $category[0] ) ) {
 				$category_element = sprintf(
-					'<a href="%1$s" class="wporg-latest-news__category">%2$s</a>',
+					'<a href="%1$s" class="wp-block-wporg-latest-news__category">%2$s</a>',
 					esc_html( get_category_link( $category[0]->term_id ) ),
 					esc_html( $category[0]->name )
 				);
@@ -90,7 +90,7 @@ function render_block( $attributes ) {
 		);
 
 		$list_items .= sprintf(
-			'<li>%1$s <div class="wporg-latest-news__details">%2$s %3$s %4$s</div></li>',
+			'<li>%1$s <div class="wp-block-wporg-latest-news__details">%2$s %3$s %4$s</div></li>',
 			$title_element,
 			$category_element,
 			! empty( $category_element ) ? '<span>·</span>' : '',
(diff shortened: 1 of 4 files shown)


jasmussen:
Looks good!

* Just to confirm, the text inside the boxes is 16px and the text below, "Popular categories" is 14 correct? (That's what they should be)
* Desktop looks good, and the way items wrap, especially the title, is correct, i.e. top-aligned vs. center aligned. The way the date stacks on mobile is also fine. But in your block editor screenshot, and in the tablet screenshot, the date (`September 14, 2023`) wraps on two lines, which is not ideal. There's probably a site-wide separate issue we can look into, in terms of changing the date format. I.e. maybe `Sep. 14, 2023` might help a lot, but unless we can do that in this PR it would be good to think about other solutions.
 * Can we add white-space: nowrap; to the date, so that it never wraps and instead expands the column? 
 * Or can we increase the breakpoint so it stacks sooner? 
 * Ideally we can make a consistent rule for displaying date/time-stamps to document for every section across the site. But in lieu of that, perhaps we can make an exception to do it just for this one site, then establish the rule here, and apply it elsewhere as we go?

Let me know your thoughts.

adamwoodnz:
> * the text inside the boxes is 16px and the text below, "Popular categories" is 14 correct?

Correct!

> * But in your block editor screenshot, and in the tablet screenshot, the date (`September 14, 2023`) wraps on two lines, which is not ideal

I'll experiment with no-wrap and column size, I think the full text is probably fine unless we like the shorter format.

> * perhaps we can make an exception to do it just for this one site, then establish the rule here, and apply it elsewhere as we go?

👍 Works for me

adamwoodnz:
Updated to use `white-space: nowrap` and also increased the breakpoint where we switch to a 2 row layout (599px -> 767px), works well I think:

### Screenshots

| Tablet (768px) | < Tablet (767px shown) |
|-|-|
| localhost_8888_latest-news-cards_(iPad) | localhost_8888_latest-news-cards_ |

adamwoodnz:
Merging this, but we can further refine the responsive behaviour if necessary.

PR #513, opened 2023-11-20
Rest-Api: Add a remembers REST api endpoint.

StevenDufresne:
This adds a rest api endpoint to retrieve memorialized users to display on `/remembers`.

**Todo**
- This currently returns the `user_id`, it should also return a display name
  - What's the best way to retrieve the `display_name`?
- Decide whether we need better permissions check

### DIFF
--- a/mu-plugins/rest-api/endpoints/class-wporg-rest-remembers-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-rest-remembers-controller.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace WordPressdotorg\MU_Plugins\REST_API\Remembers;
+
+use WP_Error;
+use WP_REST_Controller, WP_REST_Server, WP_REST_Response;
+
+defined( 'WPINC' ) || die();
+
+/**
+ *
+ * This controller is used to provide user data for wp.org/remembers.
+ *
+ * @see WP_REST_Controller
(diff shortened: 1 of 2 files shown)


pkevan:
Ideally, the endpoint would return all data required to build the page, so:

- user display name
- profile link (url)

user_id itself isn't important from looking at the designs.

There doesn't seem a need to have a permissions check, since the data is public, unless there is some other reason to do so?

StevenDufresne:
> profile link (url)

Ah yes, correct. That too.

> There doesn't seem a need to have a permissions check, since the data is public, unless there is some other reason to do so?

I don't think so, I just wanted to confirm.

pkevan:
I've added the fields into the response in `67c8d2c` - took me a while to figure out, since there is a bug with grabbing users across the network, when they aren't a member of the querying site, hence the `blog_id => 0` part.

StevenDufresne:
This works. I may have been thinking too much with my front-end hat when I suggested this REST endpoint. It appears like the block will need to be dynamic so calling a REST endpoint is probably not the most efficient way to retrieve this list. A mu-plugin could also work. 🤔

pkevan:
> thinking too much with my front-end hat when I suggested this REST endpoint

🤷 can always refactor it - the data can be grabbed separately, it's not hugely complex, as you can see!

PR #515, opened 2023-11-20
Global Header: Add top-level Events link to primary navigation and remove WordCamp and Meetups links

ndiego:
This PR adds Events ([events.wordpress.org](https://events.wordpress.org/)) as a top-level item in the header navigation on WordPress.org. It also removes the WordCamp and Meetups links. 

This corresponds to the new landing page [proposed here](https://github.com/WordPress/wordcamp.org/issues/1009), and **should only be merged once the new page is live**. 

| Before | After |
|-|-|
|image|image| 

~~@dorsvenabili do you think the Events link should be placed between the Community and About items?~~

**Edit:** I have updated the After preview to reflect the [comment](https://github.com/WordPress/wporg-mu-plugins/pull/515#issuecomment-1822638728) below

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -544,16 +544,6 @@ function get_global_menu_items() {
 					'type'  => 'custom',
 				),
 				array(
-					'title' => esc_html_x( 'WordCamp ↗︎', 'Menu item title', 'wporg' ),
-					'url'   => 'https://central.wordcamp.org/',
-					'type'  => 'custom',
-				),
-				array(
-					'title' => esc_html_x( 'Meetups ↗︎', 'Menu item title', 'wporg' ),
-					'url'   => 'https://www.meetup.com/pro/wordpress/',
-					'type'  => 'custom',
-				),
-				array(
 					'title' => esc_html_x( 'Job Board ↗︎', 'Menu item title', 'wporg' ),
 					'url'   => 'https://jobs.wordpress.net/',
 					'type'  => 'custom',
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -586,6 +586,11 @@ function get_global_menu_items() {
 					'type'  => 'custom',
 				),
 			),
+			array(
+				'title' => esc_html_x( 'Events', 'Menu item title', 'wporg' ),
+				'url'   => 'https://events.wordpress.org/',
+				'type'  => 'custom',
+			),
 		),
 	);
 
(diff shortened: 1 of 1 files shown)


dorsvenabili:
I'd include the "Events" menu between "Learn" and "Community" because "Events" has a lower barrier to get involved, it only implies to check next events and to confirm attendance. The Community submenues include a list of pages that have a higher barrier to understand and get involved in my opinion.

ndiego:
> I'd include the "Events" menu between "Learn" and "Community" because "Events" has a lower barrier to get involved, it only implies to check next events and to confirm attendance. The Community submenues include a list of pages that have a higher barrier to understand and get involved in my opinion.

Good call, I have updated the Events position.

StevenDufresne:
I agree that events need to be surfaced better. I do want to point out that the menu was trimmed down because it had grown and become hard to navigate. I feel like this is moving us in that direction again so we should be critical of adding items by default.

How often do we think people will click on the events page from say `/themes` or `/documentation`? 

Have we considered other options highlighting events with a more temporary treatment to gauge how many clicks this will get?

## Example

wordpress org_ (2)

dorsvenabili:
+1 to experiment with different places as long as it's permanently visible and it's not in a sub-menu.

For context: The Events Landing Page is very important for the sustainability of the project, for more people to connect with the community, learn, use WP and get involved, so I still think that Events should be in the top menu or in any prominent place of the header in all sites of wp.org :)

angelasjin:
Events are a primary, lower-barrier way that we bring new WordPressers in, so I agree that it should be immediately visible. You don't need to be a current community member to join an event! Further, it's also a place to connect with the existing community AND to learn about WordPress, so I agree with @dorsvenabili that we add it as a top level item alongside Learn and Community. 

I do also want to acknowledge @StevenDufresne's point around too many items in the top level menu and needing to be critical about them. And, I think this is one that should be added to top level :)

pkevan:
Screenshot 2023-12-12 at 15 29 26

^^ By adding another top level menu item, the menus gets condensed at certain screen width, roughly 890-950px (and before the menu becomes a hamburger style) - we should probably address that, if not immediately then soon.

ndiego:
> ^^ By adding another top level menu item, the menus gets condensed at certain screen width, roughly 890-950px (and before the menu becomes a hamburger style) - we should probably address that, if not immediately then soon.

Yeah good point. Clicking on the "..." does open a menu with all the hidden menu items in it, but not ideal.

angelasjin:
I would like to avoid needing to click on the "..." If that's the case, then I would recommend we put Events under Community, replacing the WordCamp and Meetups links.

angelasjin:
It sounds like adding another item to the top level menu would result in items being pushed into "..." on some screens, so my recommendation is to put "Events" under "Community" in the following order:
- Events
- Make WordPress
- Five for the Future
- Job Board

ndiego:
> Events
> Make WordPress
> Five for the Future
> Job Board

This ordering looks good to me. Increases the priority of Events in the list while not creating any issues with how the navigation breaks down on mobile devices.

PR #519, opened 2023-11-22
Local Navigation: Add custom height for local-navigation-bar

### LINKED ISSUE #373: Broken code block sticky header
sticky

outdoor2kode:
Fixes https://github.com/WordPress/wporg-developer/issues/373

This PR adds customizability to local-navigation-bar block so that themes that use the block can adjust the height according to different needs.

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -16,6 +16,7 @@
 }
 
 .wp-block-wporg-local-navigation-bar {
+	height: var(--wp--custom--local-navigation-bar--spacing--height, auto);
 
 	@media (min-width: 890px) {
 		& .global-header__wporg-logo-mark {
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -14,7 +14,7 @@
 }
 
 .wp-block-wporg-local-navigation-bar {
-	height: var(--wp--custom--local-navigation-bar--spacing--height, auto);
+	height: var(--wp--custom--local-navigation-bar--spacing--height, 60px);
 
 	@media (min-width: 890px) {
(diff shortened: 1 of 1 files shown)


mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
```diff
@@ -16,6 +16,7 @@
 }
 
 .wp-block-wporg-local-navigation-bar {
+	height: var(--wp--custom--local-navigation-bar--spacing--height, auto);
```
adamwoodnz: At the moment we control this height by adding padding. I think if we're going to set a fixed height we should remove that padding and set a default height here. If I do that on Developer, the items fill the height nicely due to the flex layout.
 
```suggestion
	height: var(--wp--custom--local-navigation-bar--spacing--height, 60px);
```

Screenshot 2023-11-23 at 3 12 51 PM

adamwoodnz (commented):
Nice work, this will obviously need testing on Showcase and Documentation sites :)

mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
outdoor2kode: I am thinking about keeping 'auto' but we can remove the padding. 

Because I'm looking to the theme itself to decide how much height to use, rather than uniformly fixing it in mu-plugins. Like the height [here](https://github.com/WordPress/wporg-developer/pull/404/files#diff-352bae2bc3c7b2b54e339a2f67279c9ae3fc544c474c8916082c95b3bba4d914R157) for the Developer is 60px, but other themes can define their own. 

That being said, I guess all our themes' heights should be 60px, right? 
In that case, we could just set it to 60px here.

outdoor2kode: Padding has been removed.

adamwoodnz: Nice, yeah setting the default here at 60px means no changes in any of the themes currently using it.

PR #525, opened 2023-11-29
Remove extra space above ToC when scrolling down the page

### LINKED ISSUE #521: Sidebar container: hide 'Back to top' link until scrolled
Can we show it only if you've scrolled at least a little bit?

image

This one is actually kind of important, to the point that if we can't hide until scroll for launch we should hide the entire link until we can get that to work. Mostly we want to avoid the above case where you are scrolled to the top, but the link is still showing, quite prominently in some cases.

### LINKED ISSUE #524: Extra space above TOC on scroll
When scrolling down the page, there is some additional white space above the TOC. 

image

This is separate, but related to https://github.com/WordPress/wporg-mu-plugins/issues/491

adamwoodnz:
Closes #524 
Closes #521 

The ToC now scrolls up with the page until it reaches 60px below the fixed local nav (gap is proportional to the local nav height), where it becomes fixed. When scrolling to the bottom it will become unfixed when it reaches 80px above the footer (matches edge space).

The changes include fixing the logic for the bottom positioning, which likely broke when we recently made changes to the global header positioning.

A side effect of delaying the fixed position is that we also delay the 'Back to Top' link being displayed (#521).

### Before

Gap at top, and jump at bottom

Developer toc before

### After

Scrolling, then fixed at top, smooth stop at bottom

Developer toc after

### Testing

Test with the Documentation and Developer sites, in sandbox or local. Behaviours to test:

- When ToC fits in viewport the scrolling behaviour should match the screenshot above
- When ToC doesn't fit it should just scroll with the content
- Check initial position aligns with H1 on Developer. On Documentation the start position is higher due to the search box, but it should match prod.
- Below desktop size the ToC should be collapsed below the H1 as before, margins should be unaffected.
- The ToC is used on many parts of the site, eg. single code reference, all the handbooks, CLI commands. Test a range.

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -14,7 +14,7 @@
 }
 
 .wp-block-wporg-local-navigation-bar {
-	height: var(--wp--custom--local-navigation-bar--spacing--height, 60px);
+	height: var(--wp--custom--local-navigation-bar--spacing--height);
 
 	@media (min-width: 890px) {
 		& .global-header__wporg-logo-mark {
@@ -121,4 +121,9 @@
 			margin-left: 8px;
 		}
 	}
+}
+
+/* Set up the custom properties. These can be overridden by settings in theme.json. */
+:where(body) {
+	--wp--custom--local-navigation-bar--spacing--height: 60px;
 }
--- a/mu-plugins/blocks/sidebar-container/src/view.js
+++ b/mu-plugins/blocks/sidebar-container/src/view.js
@@ -1,8 +1,9 @@
 /**
  * This is the calculated value of the admin bar + header height + local nav bar.
+ * LOCAL_NAV_HEIGHT fallback value matches that of the CSS variable.
  */
-const localNavHeight = getCustomPropValue( '--wp--custom--local-navigation-bar--spacing--height' );
-const FIXED_HEADER_HEIGHT = 32 + 90 + localNavHeight;
+const LOCAL_NAV_HEIGHT = getCustomPropValue( '--wp--custom--local-navigation-bar--spacing--height' ) || 60;
+const FIXED_HEADER_HEIGHT = 32 + 90 + LOCAL_NAV_HEIGHT;
 
 /**
  * Get the value of a CSS custom property.
--- a/mu-plugins/blocks/sidebar-container/src/view.js
+++ b/mu-plugins/blocks/sidebar-container/src/view.js
@@ -104,42 +104,11 @@ function init() {
 				toggleButton.setAttribute( 'aria-expanded', true );
 				list.setAttribute( 'style', 'display:block;' );
 			}
-
-			// Use the same media query that determines whether it's 2 columns,
-			// because we don't need to manage scroll when one column.
-			if ( ! window.matchMedia( '(min-width: 1200px)' ).matches ) {
-				return;
-			}
-
-			// After toggle, see if we need to update the sidebar classes.
-			if ( isSidebarWithinViewport( container ) ) {
-				container.classList.add( 'is-fixed-sidebar' );
-			} else {
-				container.classList.remove( 'is-fixed-sidebar' );
-				window.scrollTo( { top: 0, left: 0, behavior: 'instant' } );
-			}
-			// Remove the bottom sidebar class and re-run the check to re-add
-			// it if the newly-expanded sidebar crashes into the footer.
-			container.classList.remove( 'is-bottom-sidebar' );
-			const isBottom = onScroll();
-			// If the sidebar is at the bottom, opening it might push it
-			// upwards off the screen, so scroll to it (take into account
-			// the fixed headers, plus a little extra space).
-			if ( isBottom ) {
-				window.scrollTo( {
-					top:
-						container.offsetTop -
-						FIXED_HEADER_HEIGHT -
-						getCustomPropValue( '--wp--preset--spacing--20' ),
-					left: 0,
-					behavior: 'instant',
-				} );
-			}
 		} );
 	}
 
 	if ( container ) {
-		if ( isSidebarWithinViewport( container ) ) {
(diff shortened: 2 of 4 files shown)


mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
```diff
@@ -122,3 +122,8 @@
 		}
 	}
 }
+
+/* Set up the custom properties. These can be overridden by settings in theme.json. */
+:where(body) {
+	--wp--custom--local-navigation-bar--spacing--height: 60px;
```
adamwoodnz, line 128: Leave some space between the fixed header and the heading

mu-plugins/blocks/table-of-contents/postcss/style.pcss
```diff
@@ -139,6 +139,6 @@
 
 @media (min-width: 890px) {
 	.is-toc-heading {
-		scroll-margin-top: var(--wp-local-header-offset, 0);
+		scroll-margin-top: calc(var(--wp--custom--local-navigation-bar--spacing--height) + var(--wp--preset--spacing--20));
```
adamwoodnz, line 142: Leave some space between the fixed header and the heading

mu-plugins/blocks/sidebar-container/src/view.js
```diff
@@ -97,68 +110,33 @@ function init() {
...
-			if ( isBottom ) {
-				window.scrollTo( {
-					top:
-						container.offsetTop -
-						FIXED_HEADER_HEIGHT -
-						getCustomPropValue( '--wp--preset--spacing--20' ),
-					left: 0,
-					behavior: 'instant',
-				} );
-			}
```
adamwoodnz, line 130: The toggle behaviour on desktop is disabled as of https://github.com/WordPress/wporg-mu-plugins/pull/499 so this is dead code

mu-plugins/blocks/sidebar-container/src/view.js
```diff
@@ -27,64 +40,60 @@ function getCustomPropValue( name, element = document.body ) {
...
-	// Only run the scroll code if the sidebar is fixed.
-	const sidebarContainer = document.querySelector( '.wp-block-wporg-sidebar-container' );
-	if ( ! sidebarContainer || ! sidebarContainer.classList.contains( 'is-fixed-sidebar' ) ) {
+	// Only run the scroll code if the sidebar is floating on a wide screen.
+	if ( ! mainEl || ! container || ! window.matchMedia( '(min-width: 1200px)' ).matches ) {
 		return;
 	}
 
-	const mainEl = document.getElementById( 'wp--skip-link--target' );
-	const footerStart = mainEl.offsetTop + mainEl.offsetHeight;
```
adamwoodnz, line 37: Optimization: these don't need to be looked up on every scroll event

mu-plugins/blocks/sidebar-container/src/view.js
```diff
@@ -27,64 +40,60 @@ function getCustomPropValue( name, element = document.body ) {
...
-
-	const gap = getCustomPropValue( '--wp--custom--wporg-sidebar-container--spacing--margin--top' );
-	const viewportYOffset = window
-		.getComputedStyle( document.documentElement )
-		.getPropertyValue( 'margin-top' )
-		.replace( 'px', '' );
```
adamwoodnz, line 43: Optimization: these don't need to be set on every scroll event

mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
```diff
@@ -14,7 +14,7 @@
 }
 
 .wp-block-wporg-local-navigation-bar {
-	height: var(--wp--custom--local-navigation-bar--spacing--height, 60px);
```
outdoor2kode, line 17: I originally thought this already meant that it could be overridden by settings in the theme.json, but it turns out it can't? Asking this because I saw you explicitly set up 60px below.
Or setting explicitly here is primarily for easier usage elsewhere in the code, but I only saw [one](https://github.com/WordPress/wporg-mu-plugins/pull/525/files#diff-2a5c2f89a05985b1cb2923379048eaf316a396d96566e495b29794cfd8c4b765R142) that doesn't have fallback in place though.

outdoor2kode (commented):
🚀 Works well. Just two minor questions.

After clicking the "Show Less" button and the sidebar is repositioned, we probably can adjust the scroll position [here](https://github.com/WordPress/wporg-mu-plugins/pull/525/files#diff-59a30dc60b14bd54fe2d39307bbe43f25f12812bbbf65a7d34251fd81cc9dd0aR128) to a point where the sidebar is visible.
| Before | After |
|-|-|
| before | after |

mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
adamwoodnz: Good question; I've been following Kelly's example [here](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/query-filter/postcss/style.pcss#L325).

I can't remember exactly why I had to make this change, but I think I found that JS wasn't reading the custom property default value correctly, probably because this var is scoped to the component and not on body like they are in other components.

adamwoodnz:
> After clicking the "Show Less" button and the sidebar is repositioned, we probably can adjust the scroll position [here](https://github.com/WordPress/wporg-mu-plugins/pull/525/files#diff-59a30dc60b14bd54fe2d39307bbe43f25f12812bbbf65a7d34251fd81cc9dd0aR128) to a point where the sidebar is visible.

Yes that code isn't having the desired effect anymore, but also the page isn't breaking like it was before we added this fix, as the scroll code is now running even if the sidebar isn't `is-sidebar-fixed`.

I think what's expected is not to have the sidebar in view, but to maintain the scroll position when the button was clicked, in which case it should be the responsibility of the `wp-block-wporg-code-table`. What do you think?

I'll merge this as is anyway, and we can follow up with the right solution.

PR #534, opened 2023-12-05
Time Block: Optional link and format.

StevenDufresne:
In building https://github.com/WordPress/wordcamp.org/pull/1124, we need a front-end time block to display a more user-friendly block. There are 2 things that don't work with the current time block:
- It links to the WorldClock
- It doesn't allow you to update formatting.

This PR makes the WordClock link optional based on the presence of the `data-iso` property and allows you to pass in the data format in via `data-date-format`.

### DIFF
--- a/mu-plugins/blocks/time/src/view.js
+++ b/mu-plugins/blocks/time/src/view.js
@@ -16,7 +16,7 @@ function convertTimes() {
 
 	const formatTime = function (
 		date,
-		timeFormat = {
+		dateFormat = {
 			weekday: 'long',
 			month: 'long',
 			day: 'numeric',
@@ -26,7 +26,7 @@ function convertTimes() {
 			timeZoneName: 'short',
 		}
 	) {
-		return date.toLocaleTimeString( window.navigator.language, timeFormat );
+		return date.toLocaleTimeString( window.navigator.language, dateFormat );
 	};
 
 	const formatDate = function ( date ) {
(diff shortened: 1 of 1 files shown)


adamwoodnz:
So how are we intending to use this, just by adding html to be picked up by view.js, as if it was produced by the editor, eg. `<time class="wporg-time" datetime="..." data-iso="..." data-date-format="...">{ utcDateDescription }</time>`?

This markup is generated on change in the editor, as it's a rich text format, not a standard block.

StevenDufresne:
> So how are we intending to use this, just by adding html to be picked up by view.js, as if it was produced by the editor, eg. `<time class="wporg-time" datetime="..." data-iso="..." data-date-format="...">{ utcDateDescription }</time>`?
> 
> This markup is generated on change in the editor, as it's a rich text format, not a standard block.

Yes.

adamwoodnz:
Is there a PR with it being used yet?

adamwoodnz (approved):
Regression tested this in sandbox and looks ok to me.

It would be nice to test with the new usage though.

StevenDufresne:
Not yet. I can get one up.

StevenDufresne:
In the end, I don't think this will be usable for the events page since it replaces everything in the time block, removing important layout-related html. We should standardize dates across wp.org but since this is used specifically for in-content dates, we can leave it as is for now. We don't have good enough context for abstraction. I'll close this until a later date.

PR #538, opened 2023-12-07
Google Map: Optimize dependencies

### LINKED ISSUE #537: Google Map: Remove dependency on WordPress components
The Google Map block currently depends on Spinner and Button from the WordPress Components npm package. This package is not optimised for frontend use and adds a significant blocking request to the page, resulting in poor page load performance.

We should replace these simple UI elements with local components.

adamwoodnz:
Fixes #537 
Related: https://github.com/WordPress/wordcamp.org/issues/1128

Moves lighthouse perf score from ~68 to ~82 on localhost for the Events landing page.

Removes dependency on WordPress Components package and shifts lodash dependencies to individual packages (throttle and debounce).

#### Before

Screenshot 2023-12-08 at 10 08 19 AM

#### After

events spinner |

#### Before (All (3.6 MiB), Unused Bytes (1.8 MiB))

Screenshot 2023-12-08 at 9 56 10 AM

#### After (All (1.9 MiB), Unused Bytes (810.6 KiB))

Screenshot 2023-12-08 at 9 59 14 AM

### Testing

Should be tested with a site using the Screenshot Preview block, as it has dependency changes (Showcase?), and Events for the Google Map block.

### DIFF
--- a/mu-plugins/blocks/google-map/postcss/style.pcss
+++ b/mu-plugins/blocks/google-map/postcss/style.pcss
@@ -6,7 +6,6 @@
 	& .wporg-marker-search__container {
 		display: flex;
 
-		/* Enqueuing the `wp-components` stylesheet add too much to the page load just for these styles, so they're duplicated here. */
 		& button.is-tertiary {
 			background: none;
 			border: none;
--- a/mu-plugins/blocks/google-map/postcss/style.pcss
+++ b/mu-plugins/blocks/google-map/postcss/style.pcss
@@ -83,7 +83,7 @@
 		position: absolute;
 		left: var(--local-position);
 		top: var(--local-position);
-		color: var(--wp-components-color-accent, var(--wp-admin-theme-color, #3858e9));
+		color: var(--wp--custom--google-map--spinner--color);
 		overflow: visible;
 		opacity: 1;
 		background-color: transparent;
@@ -125,4 +125,9 @@
 			color: #ff2b2b;
 		}
 	}
+}
+
+/* Set up the custom properties. These can be overridden by settings in theme.json. */
+:where(body) {
+	--wp--custom--google-map--spinner--color: var(--wp--preset--color--blueberry-1, var(--wp-admin-theme-color, #3858e9));
 }
--- a/mu-plugins/blocks/google-map/postcss/style.pcss
+++ b/mu-plugins/blocks/google-map/postcss/style.pcss
@@ -74,8 +74,47 @@
 		background-image: url(../images/separator-dot.svg);
 		background-repeat: no-repeat;
 	}
+
+	& .wporg-google-map__spinner {
+		--local-size: 16px;
+		--local-position: calc(50% - (var(--local-size) / 2));
+
+		width: var(--local-size);
+		height: var(--local-size);
+		position: absolute;
+		left: var(--local-position);
+		top: var(--local-position);
+		color: var(--wp-components-color-accent, var(--wp-admin-theme-color, #3858e9));
+		overflow: visible;
+		opacity: 1;
+		background-color: transparent;
+
+		& > circle {
+			fill: transparent;
+			stroke-width: 1.5px;
+			stroke: rgb(221, 221, 221);
+		}
+
+		& > path {
+			fill: transparent;
+			stroke-width: 1.5px;
+			stroke: currentcolor;
+			stroke-linecap: round;
+			transform-origin: 50% 50%;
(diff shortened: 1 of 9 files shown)


mu-plugins/blocks/google-map/postcss/style.pcss
```diff
@@ -74,8 +74,47 @@
...
 
+@keyframes wp-block-wporg-google-map-spinner-animation {
+	0% {
+		transform: rotate(0deg);
+	}
+
+	100% {
+		transform: rotate(360deg);
+	}
+}
```
iandunn, line 116: There's a lot of code we have to fork for this, it may be simpler to use the older spinner instead:

`" width="20" height="20" />`

We wouldn't need to create a JS component for it or anything, we could just put it in the PHP render:

https://github.com/WordPress/wporg-mu-plugins/blob/8c447146a557982cd78ce8695aa8ec04be9319db/mu-plugins/blocks/google-map/index.php#L91

The CSS would be smaller, and may be able to be pared down even further:

https://github.com/WordPress/WordPress/blob/46406afcc4cc938f44115c49ae775bad52039710/wp-admin/css/common.css#L2315-L2327

I don't feel strongly, though. Forking the Gutenberg spinner is fine if you prefer that.

iandunn (approved):
This WFM 👍🏻 

It's disappointing that Gutenberg makes us replicate all of this. What's the point of having a component library if we can't use it on the front end? That's not something we can fix in this PR though.

mu-plugins/blocks/google-map/postcss/style.pcss
adamwoodnz: Let me see what that looks like

adamwoodnz: Tried this out and I think we should stick with the svg. It's one less request, and a little more CSS, but this won't make much difference really. There were a couple of problems with the gif:

1. If we put it directly in the php we can't center it, as the map container provides the height, and that hasn't loaded yet
2. If we render it in the react component we can center it, but it doesn't look good on blue as it's been rasterized on white

Screenshot 2023-12-08 at 11 52 45 AM

iandunn: 👍🏻

mu-plugins/blocks/google-map/postcss/style.pcss
```diff
@@ -74,8 +73,47 @@
...
+	& .wporg-google-map__spinner {
+		--local-size: 16px;
+		--local-position: calc(50% - (var(--local-size) / 2));
+
+		width: var(--local-size);
+		height: var(--local-size);
+		position: absolute;
+		left: var(--local-position);
+		top: var(--local-position);
+		color: var(--wp-components-color-accent, var(--wp-admin-theme-color, #3858e9));
```
StevenDufresne: Where does `--wp-components-color-accent` come from? Is that reliable?

StevenDufresne:
I haven't personally tested this, but the code looks good. 👍

mu-plugins/blocks/google-map/postcss/style.pcss
adamwoodnz: Hmm good spotting, I'll check

adamwoodnz: Turns out no. Set it to blueberry and made it configurable in https://github.com/WordPress/wporg-mu-plugins/pull/538/commits/6f063a3ee3fce3011750f640bfea591f02f12487

package.json
```diff
@@ -67,6 +66,8 @@
 	"dependencies": {
 		"@googlemaps/markerclusterer": "^2.4.0",
 		"google-map-react": "^2.2.1",
-		"locutus": "^2.0.16"
+		"locutus": "^2.0.16",
+		"lodash.debounce": "^4.0.8",
+		"lodash.throttle": "^4.1.1"
```
iandunn, line 71: Does this cause it to bundle that in the `front.js` build, rather than using the version that ships w/ Core?

If so, I wonder if that's actually worse for holistic performance, since the Core script would already be cached locally in a lot of places. I'm guessing that wouldn't show up in Lighthouse scores b/c it's not realistic enough to detect it.

adamwoodnz: Yeah it does cause it to be bundled. Sorry, I wasn't aware of the core dependency, and it wasn't clear to me looking at our imports, where the comments list it as an External dependency, and not a Wordpress one:

iandunn: I opened #543 to test it out, but it seems like the approach here is probably better 👍🏻

PR #543, opened 2023-12-11
Experiment: Revert to bundled Lodash dependency

iandunn:
Core already ships with lodash, and it's automatically used as an external when imported. Bundling it in individual block builds means that it can get downloaded two or more times.

This reverts a part of https://github.com/WordPress/wporg-mu-plugins/commit/e05dc6e60a6361feb5172ac42709d704068fe6e1. See https://github.com/WordPress/wporg-mu-plugins/pull/538#discussion_r1420797450

If lodash isn't already being enqueued by something else, though, then this approach will be worse. On events.w.org for instance, this removes 25k from `front.js`, but it adds `544k` by causing lodash to be enqueued.

We could maybe try to load lodash from the s.w.org CDN on all sites so that visitors can just download it once and have it reused across all w.org sites, but that could cause problems when w.org is running a different version than a site that's enqueing lodash from it. The `ver` param may protect against that, but I'm not sure it's worth the extra effort on the front end.

Overall I think it's probably better to leave things like they are, but I'm gonna go ahead and open this so we have a record of the attempt and the results.

### DIFF
--- a/mu-plugins/blocks/google-map/src/components/main.js
+++ b/mu-plugins/blocks/google-map/src/components/main.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import throttle from 'lodash.throttle';
+import { throttle } from 'lodash';
 
 /**
  * WordPress dependencies
--- a/mu-plugins/blocks/google-map/src/utilities/content.js
+++ b/mu-plugins/blocks/google-map/src/utilities/content.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import debounce from 'lodash.debounce';
+import { debounce } from 'lodash';
 const debouncedSpeak = debounce( speak, 1000 );
 
 /**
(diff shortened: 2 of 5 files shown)


iandunn:
I'm leaning towards closing this since it seems better overall to leave things like they are. What do you think @adamwoodnz ?

adamwoodnz (commented):
Yeah I'm finding it hard to reason about this because I'm not sure how often full lodash is being enqueued on the frontend. For now I think we're not loading a Google Map and Screenshot Preview on the same page, so we're not incurring that double penalty. That could change though, and would be an invisible cost.

In terms of the CDN; we already get this for each individual domain correct? But not one path for the whole of *.wordpress.org

I guess that's pretty good for performance. I'm on the fence 🙂

iandunn:
Yeah, I think we can close it for now, and readdress if we notice the perf scores drop anywhere because of lodash.

PR #546, opened 2023-12-13
Add Events to submenu

pkevan:
Removes WordCamp and Meetups link, and replaces with Events site.


pkevan:
Pending discussion around top level items, see: https://github.com/WordPress/wporg-mu-plugins/pull/515.

PR #548, opened 2023-12-14
Swap the info and tip icons in the Notice block

ndiego:
This PR follows the [corresponding PR](https://github.com/WordPress/wporg-developer/pull/435) in `wporg-developer` to update the icons in the callout markup.

The info notice should use the `i` icon, which means we can then use the `?` icon for the default (tip). The original Figma had the wrong icons, but it has now [been updated](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=101-2079&mode=design&t=B6tpdaZWB2yX7Tu5-0). 

| Before | After |
|-|-|
|image|image|


outdoor2kode (approved):
LGTM 👍

PR #551, opened 2023-12-20
Blocks: Notice: The `$content` parameter is a required parameter.

dd32:
This removes a PHP Deprecated notice, as a optional parameter was declared before a required parameter.

```
Deprecated: Optional parameter $content declared before required parameter $tag is implicitly treated as a required parameter in mu-plugins/blocks/notice/index.php on line 37
```

Not pushing this (yet) because I'm not sure if this function is called directly anywhere, which I don't think it should, but I haven't checked.

### DIFF
--- a/mu-plugins/blocks/notice/index.php
+++ b/mu-plugins/blocks/notice/index.php
@@ -34,7 +34,7 @@ function init() {
  *
  * @return string Shortcode output as HTML markup.
  */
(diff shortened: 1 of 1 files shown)


ryelle (approved):
It looks like this snuck in with #337. It wasn't called from anywhere else when it was added, so it should be fine to fix like this.

PR #554, opened 2024-01-10
Make sidebar container scroll internally

adamwoodnz:
See https://github.com/WordPress/wporg-developer/issues/453

This PR modifies the sidebar container so that once it becomes fixed, it also scrolls internally. When the footer collides with the bottom of the sidebar, the sidebar reduces in height rather than starting to scroll with the page.

It has been decoupled from the Table of Contents so that it can be used for the Developer Chapter List as well. JS changes are required to enable handling multiple sidebar containers on the page.

The scrolling areas have custom scrollbars so that they appear on hover. This has been inspired by the React docs ([example](https://react.dev/reference/react/legacy)), and should help users discover that these are scrollable more easily. With the typical Mac OS setting of hidden scrollbars, it is not immediately obvious. Note that this does not work in Firefox.

Overall this is a much simpler implementation, with no top positioning required, and no listener for `main` height changes.

### Testing

The Sidebar Container is currently used on [Developer Resources](https://developer.wordpress.org/wp-admin/) and [Documentation](https://wordpress.org/documentation).

#### Things to test

- Fixing and scrolling on pages with long and short ToCs and Chapter Lists
- Expand and collapse behaviour of ToC on small screens
- Scrollbars being shown on hover
- Show more/less Related items on Code Reference pages like http://localhost:8888/reference/functions/is_admin/

Use https://github.com/WordPress/wporg-developer/pull/447 to test Developer Resources

Documentation requires no changes, so you can use [that local environment](https://github.com/WordPress/wporg-documentation-2022) as is, and switch the mu-plugins install to this branch.

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -123,6 +123,13 @@
 	}
 }
 
+@media (min-width: 890px) {
+	/* stylelint-disable selector-id-pattern */
+	#wp--skip-link--target {
+		scroll-margin-top: var(--wp--custom--local-navigation-bar--spacing--height, 0);
+	}
+}
+
 /* Set up the custom properties. These can be overridden by settings in theme.json. */
 :where(body) {
 	--wp--custom--local-navigation-bar--spacing--height: 60px;
--- a/mu-plugins/blocks/sidebar-container/index.php
+++ b/mu-plugins/blocks/sidebar-container/index.php
@@ -45,11 +45,13 @@ function render( $attributes, $content, $block ) {
 			esc_html__( '↑ Back to top', 'wporg' )
 		)
 		: '';
+	$inlineBreakpoint = $attributes['inlineBreakpoint'];
 
 	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
-		'<div %1$s>%2$s%3$s</div>',
+		'<div %1$s data-breakpoint="%2$s">%3$s%4$s</div>',
 		$wrapper_attributes,
+		esc_attr( $inlineBreakpoint ),
 		$content,
 		$back_to_top
 	);
--- a/mu-plugins/blocks/sidebar-container/index.php
+++ b/mu-plugins/blocks/sidebar-container/index.php
@@ -39,10 +39,12 @@ function init() {
  * @return string Returns the block markup.
  */
 function render( $attributes, $content, $block ) {
-	$back_to_top = sprintf(
-		'<p class="has-small-font-size is-link-to-top"><a href="#wp--skip-link--target">%s</a></p>',
-		esc_html__( '↑ Back to top', 'wporg' )
-	);
+	$back_to_top = $attributes['hasBackToTop']
+		? sprintf(
+			'<p class="has-small-font-size is-link-to-top"><a href="#wp--skip-link--target">%s</a></p>',
+			esc_html__( '↑ Back to top', 'wporg' )
+		)
+		: '';
 
 	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
--- a/mu-plugins/blocks/sidebar-container/index.php
+++ b/mu-plugins/blocks/sidebar-container/index.php
@@ -39,19 +39,19 @@ function init() {
  * @return string Returns the block markup.
  */
 function render( $attributes, $content, $block ) {
+	$wrapper_attributes = get_block_wrapper_attributes();
+	$inline_breakpoint = $attributes['inlineBreakpoint'];
 	$back_to_top = $attributes['hasBackToTop']
 		? sprintf(
 			'<p class="has-small-font-size is-link-to-top"><a href="#wp--skip-link--target">%s</a></p>',
 			esc_html__( '↑ Back to top', 'wporg' )
 		)
 		: '';
-	$inlineBreakpoint = $attributes['inlineBreakpoint'];
 
-	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
 		'<div %1$s data-breakpoint="%2$s">%3$s%4$s</div>',
 		$wrapper_attributes,
-		esc_attr( $inlineBreakpoint ),
+		esc_attr( $inline_breakpoint ),
 		$content,
 		$back_to_top
 	);
--- a/mu-plugins/blocks/sidebar-container/postcss/style.pcss
+++ b/mu-plugins/blocks/sidebar-container/postcss/style.pcss
@@ -20,15 +20,6 @@
 		margin-bottom: 0 !important;
 		padding: var(--wp--preset--spacing--20) 0;
 
-		main & {
-			position: absolute;
-			top: calc(var(--wp-global-header-offset, 90px) + var(--wp--custom--local-navigation-bar--spacing--height, 60px));
-			margin-top: var(--wp--custom--wporg-sidebar-container--spacing--margin--top);
-
-			/* Right offset should be "edge spacing" at minimum, otherwise calculate it to be centered. */
-			right: max(var(--wp--preset--spacing--edge-space), calc((100% - var(--wp--style--global--wide-size)) / 2));
-		}
-
 		&:not(.is-fixed-sidebar) {
 
 			/* Match width of custom scrollbar when fixed, stops content width changing */
@@ -68,6 +59,15 @@
 		& * + .is-link-to-top {
 			border-top: 1px solid var(--wp--preset--color--light-grey-1);
 		}
+	}
+
+	main .wp-block-wporg-sidebar-container {
+		position: absolute;
+		top: calc(var(--wp-global-header-offset, 90px) + var(--wp--custom--local-navigation-bar--spacing--height, 60px));
+		margin-top: var(--wp--custom--wporg-sidebar-container--spacing--margin--top);
+
+		/* Right offset should be "edge spacing" at minimum, otherwise calculate it to be centered. */
+		right: max(var(--wp--preset--spacing--edge-space), calc((100% - var(--wp--style--global--wide-size)) / 2));
 	}
 }
 
--- a/mu-plugins/blocks/sidebar-container/postcss/style.pcss
+++ b/mu-plugins/blocks/sidebar-container/postcss/style.pcss
@@ -48,7 +48,6 @@
 
 			& .is-link-to-top {
 				display: block;
-				padding-top: var(--wp--preset--spacing--20);
 
 				& a {
 					color: var(--wp--preset--color--charcoal-4);
@@ -57,6 +56,7 @@
 		}
 
 		& * + .is-link-to-top {
+			padding-top: var(--wp--preset--spacing--20);
 			border-top: 1px solid var(--wp--preset--color--light-grey-1);
 		}
 	}
--- a/mu-plugins/blocks/sidebar-container/postcss/style.pcss
+++ b/mu-plugins/blocks/sidebar-container/postcss/style.pcss
@@ -18,7 +18,7 @@
 
 		width: var(--local--block-end-sidebar--width);
 		margin-bottom: 0 !important;
-		padding-top: var(--wp--preset--spacing--20);
+		padding: var(--wp--preset--spacing--20) 0;
 
 		main & {
 			position: absolute;
@@ -67,7 +67,6 @@
 
 		* + .is-link-to-top {
 			border-top: 1px solid var(--wp--preset--color--light-grey-1);
-			padding-bottom: var(--wp--preset--spacing--20);
 		}
 	}
 }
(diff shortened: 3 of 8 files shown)


mu-plugins/blocks/sidebar-container/src/view.js
```diff
@@ -36,103 +33,43 @@ function getCustomPropValue( name, element = document.body ) {
...
-		} );
-	}
 
-	if ( isSidebarWithinViewport() ) {
+	if ( mainEl && containers.length ) {
 		onScroll(); // Run once to avoid footer collisions on load (ex, when linked to #reply-title).
 		window.addEventListener( 'scroll', onScroll );
-
-		const observer = new window.ResizeObserver( () => {
-			// If the sidebar is positioned at the bottom and mainEl resizes,
```
adamwoodnz, line 117: Need to test this scenario; added to handle `Show less` actions

mu-plugins/blocks/sidebar-container/src/view.js
```diff
@@ -1,19 +1,16 @@
 /**
  * Fallback values for custom properties match CSS defaults.
  */
-const globalNavHeight = 90;
 
-const LOCAL_NAV_HEIGHT = getCustomPropValue( '--wp--custom--local-navigation-bar--spacing--height' ) || 60;
+const GLOBAL_NAV_HEIGHT = getCustomPropValue( '--wp-global-header-height' ) || 90;
```
adamwoodnz: Making this dynamic fixes a bug where top and bottom positions are incorrect, caused by the global header being shorter on short screens.

mu-plugins/blocks/sidebar-container/src/view.js
adamwoodnz: Tested, works fine as there is no top positioning to get out of sync

outdoor2kode:
I'm not sure if this has been discussed, but currently, the scrollbar appears only when the sidebar is fixed as described. What about the situation shown in the image below? At this point, the sidebar isn't fixed yet, but the chapter list is already overflowing.

image

mu-plugins/blocks/sidebar-container/postcss/style.pcss
```diff
@@ -14,41 +14,59 @@
...
+				border-radius: 10px;
+			}
 
-		&.is-fixed-sidebar .is-link-to-top,
-		&.is-bottom-sidebar .is-link-to-top {
-			display: block;
-			margin-top: 0;
+			.is-link-to-top {
+				display: block;
+				padding-top: var(--wp--preset--spacing--20);
```
outdoor2kode: We can also consider removing the extra padding when there's no TOC content.

image

image

mu-plugins/blocks/sidebar-container/postcss/style.pcss
```diff
@@ -14,41 +14,59 @@
...
-			& a {
-				color: var(--wp--preset--color--charcoal-4);
+				& a {
+					color: var(--wp--preset--color--charcoal-4);
+				}
 			}
 		}
 
-		.wp-block-wporg-table-of-contents + .is-link-to-top {
+		* + .is-link-to-top {
```
outdoor2kode: Seems we can move the style here to [above](https://github.com/WordPress/wporg-mu-plugins/pull/554/files#diff-54f3aa2f071554a0bd0ada800098cfd6cccf927f4076065991876124b149e222R58) as it would only show up when the sidebar is fixed.

outdoor2kode (approved):
Tested all the cases suggested in the description. Works as expected 👍 
Left some comments, but none of them are blockers.

mu-plugins/blocks/sidebar-container/postcss/style.pcss
```diff
@@ -14,41 +14,59 @@
...
-		top: calc(var(--wp-global-header-offset, 90px) + var(--wp--custom--local-navigation-bar--spacing--height, 60px));
-
-		/* Right offset should be "edge spacing" at minimum, otherwise calculate it to be centered. */
-		right: max(var(--wp--preset--spacing--edge-space), calc((100% - var(--wp--style--global--wide-size)) / 2));
 		width: var(--local--block-end-sidebar--width);
-		margin-top: var(--wp--custom--wporg-sidebar-container--spacing--margin--top);
 		margin-bottom: 0 !important;
+		padding: var(--wp--preset--spacing--20) 0;
+
+		main & {
```
ryelle: This seems to be the cause of [this issue for me](https://github.com/WordPress/wporg-developer/pull/469#issuecomment-1894672967)— this isn't sass, so the nesting rules aren't the same. This seems to be ignored by postcss, and then stripped out by cssnano, so the rules below are applied to all `.wp-block-wporg-sidebar-container`s. I don't know why it works for you & @renintw though.

If I move this code to a new section `main .wp-block-wporg-sidebar-container {…` it works as expected.

mu-plugins/blocks/sidebar-container/postcss/style.pcss
```diff
@@ -14,41 +14,59 @@
...
+				border: 4.5px solid transparent;
+				background-clip: content-box;
+				border-radius: 10px;
+			}
 
-		&.is-fixed-sidebar .is-link-to-top,
-		&.is-bottom-sidebar .is-link-to-top {
-			display: block;
-			margin-top: 0;
+			.is-link-to-top {
```
ryelle:
```suggestion
			& .is-link-to-top {
```

mu-plugins/blocks/sidebar-container/postcss/style.pcss
ryelle:
```suggestion
		& * + .is-link-to-top {
```

The postcss-nesting plugin requires the `&`.

mu-plugins/blocks/sidebar-container/postcss/style.pcss
adamwoodnz: > I don't know why it works for you & @renintw though.

Nice catch! All this seems to work with a dev build, but it all breaks with a prod build. Starting to think moving to one CSS style would be a really good idea as switching feature sets contributes to this sort of issue. I think that was suggested to be postcss.

ryelle: > I think that was suggested to be postcss.

I think that was before wp-scripts natively supported sass… and maybe if we switched to sass here, we could remove the custom build script here.

adamwoodnz:
> I'm not sure if this has been discussed, but currently, the scrollbar appears only when the sidebar is fixed as described. What about the situation shown in the image below? At this point, the sidebar isn't fixed yet, but the chapter list is already overflowing.

My expectation is that the user would do the default and start scrolling the page, as they wouldn't expect the Chapter List to scroll until that behaviour is observed.

https://github.com/WordPress/wporg-mu-plugins/assets/1017872/ece8f2b5-aaa0-41ec-9198-fdbe8a208359

Make sense?

outdoor2kode:
> My expectation is that the user would do the default and start scrolling the page, as they wouldn't expect the Chapter List to scroll until that behaviour is observed.

I see your idea. Thanks for the explanation with the screencast👍 
Was originally thought that users who are already familiar with this page might expect to be able to scroll directly through the chapter lists, similar to the React Docs. But the current setup also makes sense, as more list items can actually be viewed by scrolling as well.

adamwoodnz:
[In order for the Chapter List container to be fixed and scrolling while the ToC container is inline](https://github.com/WordPress/wporg-developer/pull/469#issuecomment-1894672967), the fixed styles have been decoupled from a specific breakpoint and managed via a class instead. The breakpoint for each container to switch between inline and fixed/floating can now be configured via a block attribute.

In making this solution work on iPads in particular, I found that the change in the heights of the admin bar and global nav, and the sticky behaviour of the local nav bar and admin bar make for very complex calculations. A resize handler has been added to set some these changing variables on load, and update on resize to keep everything in sync. A such you should also be able to test switching orientation, not only page load.

These changes can be viewed in [b24b6f2](https://github.com/WordPress/wporg-mu-plugins/pull/554/commits/b24b6f24d90ea198968299fdf0711879cb8482d8).

mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
```diff
@@ -123,6 +123,13 @@
 	}
 }
 
+@media (min-width: 890px) {
+	/* stylelint-disable selector-id-pattern */
+	#wp--skip-link--target {
+		scroll-margin-top: var(--wp--custom--local-navigation-bar--spacing--height, 0);
+	}
+}
+
```
adamwoodnz, line 132: Moved from the sidebar container styles, as it relates directly to the change in positioning of the local nav bar

mu-plugins/blocks/sidebar-container/src/view.js
```diff
@@ -26,120 +19,91 @@ let mainEl;
...
 		} );
-
-		observer.observe( mainEl );
 	}
 
-	// If there is no table of contents, hide the heading.
-	if ( ! document.querySelector( '.wp-block-wporg-table-of-contents' ) ) {
-		const heading = document.querySelector( '.wp-block-wporg-sidebar-container h2' );
-		heading?.style.setProperty( 'display', 'none' );
-	}
```
adamwoodnz, line 142: Moved to the [render](https://github.com/WordPress/wporg-mu-plugins/pull/554/files#diff-df75ded5b5b15eac75d036587faef1fec21f65511f97a9fdb27c8bf87e6a7b9aR60)

mu-plugins/blocks/sidebar-container/index.php
```diff
@@ -39,15 +39,19 @@ function init() {
...
-		'<p class="has-small-font-size is-link-to-top"><a href="#wp--skip-link--target">%s</a></p>',
-		esc_html__( '↑ Back to top', 'wporg' )
-	);
+	$back_to_top = $attributes['hasBackToTop']
+		? sprintf(
+			'<p class="has-small-font-size is-link-to-top"><a href="#wp--skip-link--target">%s</a></p>',
+			esc_html__( '↑ Back to top', 'wporg' )
+		)
+		: '';
+	$inlineBreakpoint = $attributes['inlineBreakpoint'];
```
ryelle:
```suggestion
	$inline_breakpoint = $attributes['inlineBreakpoint'];
```

PHP variables should be snake case. It looks like we're not running phpcs in the github actions, but you can run it locally with `composer lint mu-plugins/blocks/sidebar-container mu-plugins/blocks/table-of-contents`

mu-plugins/blocks/table-of-contents/index.php
```diff
@@ -56,11 +56,14 @@ function render( $attributes, $content, $block ) {
...
-		<!-- /wp:heading -->'
-	);
+	// If there is no content, don't render the heading.
+	$content .= empty( trim( wp_strip_all_tags( $post_content ) ) )
+		? '' 
+		: do_blocks(
+			'<!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"400"},"spacing":{"margin":{"top":"0","bottom":"0"}}},"fontSize":"normal","fontFamily":"inter"} -->
+			<h2 class="wp-block-heading has-inter-font-family has-normal-font-size" style="margin-top:0;margin-bottom:0;font-style:normal;font-weight:400">' . esc_html( $title ) . '</h2>
+			<!-- /wp:heading -->'
+		);
```
ryelle: Shouldn't the `if ( ! $items )` check prevent the entire block (incl heading) from being output? I know this is coming from the previous JS, but I'm not sure how we'd end up in a situation where ` empty( trim( wp_strip_all_tags( $post_content ) ) )` is true but `! $items` is false.

ryelle (approved):
Two pretty minor notes, but overall this is working well in my testing.

I did notice a bit of glitchy behavior when the content is shorter than the Chapters list, and the screen size is just tall enough to render chapters without scrolling, in this case ~1000-1100ish. The sidebar bounces back and forth between scroll/static.

https://github.com/WordPress/wporg-mu-plugins/assets/541093/da750cd8-43ca-4779-80e8-fff5d864f2a8

Marking this as approved, if you'd rather fix in a follow-up PR.

mu-plugins/blocks/table-of-contents/index.php
adamwoodnz: Nice catch, I moved the logic without checking what would happen just removing the JS. Works just as you described, reverted the render change in [07b512c](https://github.com/WordPress/wporg-mu-plugins/pull/554/commits/07b512cd587f1ec909ebe15f87d9919ae45091d2)

mu-plugins/blocks/sidebar-container/index.php
adamwoodnz: Fixed in [2aed927](https://github.com/WordPress/wporg-mu-plugins/pull/554/commits/2aed92730dd4ddb3e3e0864f0a09bc5d31da6453)

PR #556, opened 2024-01-11
Table of Contents: Make whole header clickable on small and medium screens

### LINKED ISSUE #547: ToC interaction target should be larger
In testing the new developer changes on mobile, I noticed the click area for the chapters on mobile is small. We should be able to click the whole container as it looks like a button. The same argument should be made for the table of contents.

## Screenshot

| Current | Proposed |
|--------|--------|
| Screenshot 2023-12-14 at 1 32 08 PM | Screenshot 2023-12-14 at 1 31 49 PM |

adamwoodnz:
Fixes #547 

I've decided to use absolute positioning to make the toggle button fill the header, so that the H2 doesn't need to be hidden and removed from the markup, thus maintaining the page structure. No visual changes.

If this approach is acceptable I'll follow on with the same for the Chapter List, in the Developer Resources repo. For now that remains a separate UI component, as there is further discussed functionality for it, like the ability to collapse left.

### Testing

The ToC block is used on Documentation and Developer Resources. Using one of those local environments, or sandbox, switch mu-plugins to this branch.

1. Open pages using the ToC, eg. [Code Reference](https://developer.wordpress.org/reference/classes/wp_query/), [Block Editor ](https://developer.wordpress.org/block-editor/) or [Optimization](https://wordpress.org/documentation/article/optimization/)
2. On screens < 1200px wide with the ToC inline, the toggle button should fill the header.
3. Expand and contract the ToC
4. Check for regressions on large screens

### DIFF
--- a/mu-plugins/blocks/table-of-contents/postcss/style.pcss
+++ b/mu-plugins/blocks/table-of-contents/postcss/style.pcss
@@ -7,32 +7,38 @@
 	@media (max-width: 1199px) {
 		border: 1px solid var(--wp--preset--color--light-grey-1);
 		border-radius: 2px;
-		padding: 15px var(--wp--preset--spacing--20);
 	}
 }
 
 .wporg-table-of-contents__header {
-	display: flex;
-	justify-content: space-between;
-	align-items: center;
+	position: relative;
+
+	@media (max-width: 1199px) {
+		padding: 15px var(--wp--preset--spacing--20);
+	}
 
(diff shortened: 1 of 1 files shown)


StevenDufresne (approved):
Works for me.

outdoor2kode:
Followed the testing steps and it worked for me 👍

ryelle (approved):
Looks good 👍🏻

PR #558, opened 2024-01-15
Add handbook meta link block

StevenDufresne:
This PR is a follow-up to https://github.com/WordPress/wporg-developer/pull/454.

This block is showing up in multiple different themes. [This one for example](https://github.com/WordPress/wporg-main-2022/pull/375).

## Example usage

Using
```
<!-- wp:wporg/handbook-meta-link {"heading":"Edit article","linkURL":"//wp.org","linkText":"Improve it on GitHub"} /-->
```

## IMPORTANT

This needs to be merged & deployed together with https://github.com/WordPress/wporg-developer/pull/470.


ryelle:
Could this be called something more generic? Or is it really necessary as a block? It's just a label + link text, not directly related to site or post meta values, or github. `wporg/label-with-link`?

Oh, I see the screen reader text now, and the history in https://github.com/WordPress/wporg-developer/pull/388. It's a bummer we need to create a whole block too have that kind of workaround.

Maybe `wporg/handbook-link`? Though if you do that, you'll want to update `filter_article_meta_block` in developer too.

Aside from that feedback, the PRs do work as-is.

I am seeing the changelog/improve this links on code reference pages (like https://developer.wordpress.org/reference/functions/block_has_support/), but I think that's true on production to. Is that intentional?
Screenshot 2024-01-15 at 5 22 58 PM

adamwoodnz:
> I am seeing the changelog/improve this links on code reference pages (like https://developer.wordpress.org/reference/functions/block_has_support/), but I think that's true on production to. Is that intentional?

No, looks like a regression 😞 

The designs show no meta info at all on code reference.

outdoor2kode:
> I am seeing the changelog/improve this links on code reference pages (like https://developer.wordpress.org/reference/functions/block_has_support/), but I think that's true on production to. Is that intentional?

This issue has already been fixed by https://github.com/WordPress/wporg-developer/pull/472.

For this PR, it works as described together with https://github.com/WordPress/wporg-developer/pull/470. 👍 

> Maybe wporg/handbook-link

It looks like a good idea. Like in the case of the developer, there is [this logic](https://github.com/WordPress/wporg-developer/blob/528e016bc7e7533347ed62a6c1638cfadee12bf3/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php#L85-L95) determining if a link is a GitHub link. So we might be able to remove the specific mention of GitHub to make it more generic here.

StevenDufresne:
Alright, I renamed this to `wporg/handbook-meta-link` and updated the developer PR to match: https://github.com/WordPress/wporg-developer/pull/470

outdoor2kode (approved):
Tested again and looks good to me.

PR #560, opened 2024-01-18
Header: Shrink inline padding to match top-level text alignment

### LINKED ISSUE #559: Main navigation dropdowns: Padding too wide
Click a dropdown, say "Community", and you get this:

Screenshot 2023-12-22 at 13 41 19

The padding of menu items is 24px, it should be 12px so that text lines up.

ryelle:
Fixes #559. This updates the padding to be the same on the header items as in the submenus, so that the text alignment is the same.

| Before | After |
|---|---|
|  |  |
|  |  |


jasmussen:
Looks good to me, thank you!

PR #562, opened 2024-01-26
Tools: Update packages, use wp-scripts for build process

ryelle:
The Interactivity API is now only available as a JS module, not a plain script, so we need to build and register our `view` code differently. As of v27, `@wordpress/scripts` supports module builds, but that also breaks compat with Node <18. So this PR updates the [node version to match core](https://make.wordpress.org/core/2023/12/20/updating-wordpress-to-use-more-modern-versions-of-node-js-npm-2/), and updates the rest of the JS packages.

I've also changed the build process to call wp-scripts directly for each block, rather than using webpack + importing the wp-scripts webpack config. The block build step is now equivalent to running:

```bash
npx wp-scripts build --experimental-modules --webpack-src-dir=./mu-plugins/blocks/[block]/src --output-path=./mu-plugins/blocks/[block]/build
```

Unfortunately due to the folder structure, we still need the custom script (wp-scripts doesn't understand how to find `[block]/src/block.json`, it only understands `src/[block]/block.json`).

See https://github.com/WordPress/wporg-parent-2021/issues/122, https://github.com/WordPress/gutenberg/pull/56143, https://github.com/WordPress/gutenberg/pull/57461

**To test**

- Run `nvm install` to switch the node version (or however you manage node versions), it should be 20.
- Run `yarn` to install the package updates
    - You might need to remove node_modules first, I had a strange issue with wp-scripts not updating correctly
- Run `yarn build`
    - All builds should succeed, both CSS and JS
    - View any file, it should be minified, ex: `cat mu-plugins/blocks/navigation/build/index.js`
- Run the new `yarn build:dev`
    - All builds should succeed
    - View any file, it should not be minified, ex: `cat mu-plugins/blocks/navigation/build/index.js`
- Run the build steps with a block, ex `yarn build:dev query-filter`
- Run the watch script
    - The initial build of all modules should succeed
    - Changing any files in blocks, JS, CSS, etc, should trigger a rebuild of the files


bin/build.js
```diff
@@ -109,7 +112,10 @@ projects.forEach( async ( file ) => {
 
 		// We `await` because JS needs to be built first— the first webpack step deletes the build
 		// directory, and could remove the built CSS if it was truely async.
-		await maybeBuildJavaScript( path.resolve( path.join( basePath, 'src' ) ), outputDir );
+		await maybeBuildBlock( path.resolve( path.join( basePath, 'src' ) ), outputDir );
+
+		// `maybeBuildBlock` supports SCSS, but current blocks still have postCSS
+		// files. Until those are converted, we still need a separate PostCSS step.
```
outdoor2kode: So will we eventually phase out all PostCSS in favor of SCSS? I'm not sure if the discussion [here](https://github.com/WordPress/wordcamp.org/pull/1172#discussion_r1425590556) is relevant because it seems to be discussing the opposite idea.

outdoor2kode (commented):
Followed the test steps, and all worked as expected. Thanks for the update👍

outdoor2kode (approved):

package.json
```diff
@@ -12,28 +12,27 @@
...
+		"postcss-cli": "11.0.0",
+		"postcss-import": "16.0.0",
+		"postcss-preset-env": "9.3.0",
 		"resolve-bin": "1.0.1",
-		"rtlcss": "3.5.0",
-		"url-loader": "^3.0.0",
-		"webpack": "5.47.1",
-		"webpack-cli": "4.9.1"
+		"rtlcss": "4.1.1",
+		"url-loader": "4.1.1"
```
adamwoodnz, line 28: How did you arrive at this set of dependencies? The changes seem more extensive than expected.

adamwoodnz (approved):
👍 Works for me too, thanks!

package.json
ryelle: I used `yarn outdated` to update everything available. I wanted to make sure any peer dependencies were up to date with the new wordpress/ packages & supported the latest node version.

bin/build.js
ryelle: Oh, thanks for pointing me to that conversation, it happened while I was away :)

Honestly I was remembering [this recent bug](https://github.com/WordPress/wporg-mu-plugins/pull/554#discussion_r1464064815) where the CSS was not building properly because it was using Sass formatting. I didn't realize there had been a more recent discussion than [the 2022 issue](https://github.com/WordPress/wporg-parent-2021/issues/1).

… turns out though, [wp-scripts does support `*.pcss`](https://github.com/WordPress/gutenberg/blob/2569121b896ec18db4ce4262d6bbfbac72b32045/packages/scripts/config/webpack.config.js#L211C1-L214C6), so we could remove this without forcing a Sass decision. I'll update the comment here and try that in a new PR.

ryelle:
I'm going to merge this, since it doesn't require a deploy & isn't affected by a site's Gutenberg version.

PR #569, opened 2024-02-02
Don't overwrite position 0 plugins when running in a switched context

dd32:
As a followup to #561 some fatal were encountered in a cron task.

It turns out that the flow is something like this:
1. Cron Load Pattern Directory
2. This kicks in, realises that it should load Gutenberg-16.8
3. The Pattern Directory code switches to the Translate.wordpress.org site, and includes the translate plugins
4. Translate.wordpress.org doesn't have the Gutenberg plugin active
   - `active_plugins[ array_search() = false ]` translates to `active_plugins[0] = gutenberg-16.8/...`, which was previously glotpress, not gutenberg.
   - It then loads the other GlotPress plugins that require GlotPress to be active first, and fatals.

Supporting loading plugins for a switched site is not something that WordPress really allows for; but is something that can be done "good enough" if the two sites are running an almost different set of plugins. For example; the Pattern Directory does it like this:
https://github.com/WordPress/pattern-directory/blob/43e0eedfba131cd1fdfc24ac2939a48629acff2e/public_html/wp-content/plugins/pattern-translations/includes/makepot.php#L166-L189

This PR just verifies that the `$from` plugin is actually in the array before mangling it.

I considered using `blog_id` checks or checking that `$check` exists in the array at filter time; but this isn't enough.
Consider the following, SiteA should be filtered to `gutenberg-16.8` and if the plugins for SiteB get loaded, it should also load that same version of Gutenberg to avoid conflicts.
 - SiteA: `gutenberg`, `plugin-that-causes-gutenberg-16.8-to-load`.
 - SiteB: `gutenberg`, `translation-plugin`

### DIFF
--- a/mu-plugins/plugin-tweaks/incompatible-plugins.php
+++ b/mu-plugins/plugin-tweaks/incompatible-plugins.php
@@ -36,6 +36,12 @@ const PLUGINS = [
  * Check the above list of plugins, and filter the appropriate option.
  *
  * This needs to be done on plugin inclusion, as network-wide plugins are included immediately after mu-plugins.
+ *
+ * NOTE: This doesn't support blog switching well at all, on a switched blog the filters will be applied to the wrong blog,
+ *       this isn't as bad as it sounds, as loading plugins for other sites from the context of a different site rarely
+ *       works in the first place. Instead, this code simply double-checks that the `$from` plugin is active.
+ *       The `$check` plugin is not checked for in the switched context, as including two different versions of the same
+ *       plugin on the same request is not going to work, so it still needs to attempt to load the versioned version.
  */
 function filter_the_filters() {
 	$active_plugins          = (array) get_option( 'active_plugins', [] );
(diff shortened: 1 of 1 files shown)


dd32:
> Consider the following, SiteA should be filtered to gutenberg-16.8 and if the plugins for SiteB get loaded, it should also load that same version of Gutenberg to avoid conflicts.
>
> SiteA: gutenberg, plugin-that-causes-gutenberg-16.8-to-load.
> SiteB: gutenberg, translation-plugin

This still doesn't "fix" that if the problematic plugin is **network** activated on SiteA, and **per-site** activated on SiteB.. 
To fix that, we could always do the `add_filter()` for both, if the plugin is detected in either..

The 2nd commit resolves that. Just in case `gutenberg` ever gets activated on the Translate site (It's network activated, so this code doesn't try to include it at present)

But then to take it a step further, should `filter_the_filters` be run on switch, to pick up cases like.. loading siteB and then switching to SiteA.. but if the incompatible plugin is already loaded, there's not much we can do... 

I think I'm happy with this latest change, as it supports things best we currently can, and should avoid fataling except in the most edge-casey situation.

Issue #535, opened 2023-12-06
A polished local nav component


fcoveram:
Now that the [new Showcase](https://wordpress.org/showcase/) has been released, Developers section [is in progress](https://developer.wordpress.org/redesign-test/), and [other designs](https://www.figma.com/files/project/1339429) are in a solid position, I noticed there is room for improvement to make navigation components more flexible and appealing when displaying location across the site.

I have been exploring different ideas, and this week I landed on a design that addresses key areas of navigation. Before dive into detail, here is a prototype involving four sections of the site.

https://github.com/WordPress/wporg-mu-plugins/assets/895819/84b608d7-1f21-4793-8317-72fcf85ee29f

https://github.com/WordPress/wporg-mu-plugins/assets/895819/9e6b70ac-bf55-425d-a687-cbfb2f4a6a26

---

In a previous comment, I mixed up the site depth taxonomy and the nested levels. That ended up confusing more and revisiting the breadcrumb location a few times. So to avoid new misunderstandings, I will refer to the page structure from the following logic.

| Level | Example 1 | Example 2 | Example 3 |
|:---|:---|:---|:---|
| Level 1 | News | About WordPress | Showcase |
| Level 2 | Community | Etiquette | [The Line Hotels](https://wordpress.org/showcase/the-line-hotels/) |
| Level 3 | … | … | … |

## Components height

One minor spacing tweak across components is making global header, local nav, and breadcrumb components 70px height on big screens and 60px on small ones. This equivalency makes the layout reachable across breakpoints and eases the interaction shown in the video.

![Components height](https://github.com/WordPress/wporg-mu-plugins/assets/895819/f0cccb74-3cca-463c-bb59-f157e7b44551)

## Current page visibility

When users are in level 2 pages not displayed in the local nav, the component shows the page name next to the section label. For level 3 and deeper pages, page name is hidden as breadcrumb takes that role. But once you scroll down, all pages, except for the section’s landing pages, show the page name. This will ease page and section recognition as the component remains fixed when you scroll down.

This change makes the site details page in Showcase simpler by removing breadcrumb above the cover image, whereas, deep pages in Documentation keep you in context as you scroll down in text-heavy posts.

![Current page visibility](https://github.com/WordPress/wporg-mu-plugins/assets/895819/7eba9e96-4eef-4c51-9559-37985b195c5e)

## Handy breadcrumb

[...]

PR #570, opened 2024-02-07
Navigation: Add support for submenu items in the hardcoded menus

ryelle:
This updates the Navigation block filter to enable a `submenu` config to allow nested navigation, and adds style to the Local Navigation Bar to better style nested menus on small screens.

Currently, this only supports submenus that open on click (matching the behavior of the global nav). If we need linked parent elements, we can add `url` support in the future.

See #535.

**To test**

Try with the About PR https://github.com/WordPress/wporg-main-2022/pull/387, or add a menu with a `submenu` to the config for any site:

```php
add_filter(
	'wporg_block_navigation_menus',
	function( $menus ) {
		return array(
			'test' => array(
				array(
					'label' => 'News',
					'url' => '#',
				),
				array(
					'label' => 'About',
					'submenu' => array(
						array(
							'label' => 'Item A',
							'url' => '#',
						),
						array(
							'label' => 'Item B',
							'url' => '#',
						),
					),
				),
			),
		);
	}
);
```

Use this menu with the following code:

```html
<!-- wp:navigation {"menuSlug":"test","textColor":"blueberry-1","icon":"menu","layout":{"type":"flex","orientation":"horizontal"},"fontSize":"small","openSubmenusOnClick":true} /-->
```

The navigation should render correctly - using the hierarchy - and there should be no PHP errors.

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -114,6 +114,7 @@
 	/* Navigation. */
 
 	/* Remove padding from menu items with background color, which is used to color the modal background. */
+	& .wp-block-navigation ul.has-background,
 	& .wp-block-navigation:where(.has-background) .wp-block-navigation-item a:not(.wp-element-button),
 	& .wp-block-navigation:where(.has-background) .wp-block-navigation-submenu a:not(.wp-element-button) {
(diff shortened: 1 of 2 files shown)


adamwoodnz (approved):
👍 Works for me in https://github.com/WordPress/wporg-main-2022/pull/387, left a couple of comments there.

PR #573, opened 2024-02-12
Local Navigation: Improve style and behavior across screen sizes

ryelle:
This PR implements the style & behavior changes described in #535. The biggest change is that the navigation will now switch to the collapsed (mobile) view if it does not have enough space to stay on one single line (fixes https://github.com/WordPress/wporg-documentation-2022/issues/80). 

I did that by dropping in a second menu automatically - a filter is added to any nav blocks with `wporg/local-navigation-bar` as a parent block. That filter (`add_extra_navigation`) runs over the render_block HTML, so it can use all the same block attributes to create a second `wp:navigation` which is always in the collapsed (mobile) state. It's also given a special class name, so that it can be hidden/made visible by CSS. This ensures only one menu is visible/accessible at a time. In the javascript, a function is added to resize to check the available width against the section title & navigation items, and toggles the appropriate class based on what fits better.

Using a second menu like this also avoids trying to work around the current navigation menu toggle code, which has changed dramatically in the past.

Since the "mobile" view now shows on non-mobile screen sizes, the design was also updated and the styles here switch to a dropdown style when this version is shown at >600px wide.

There are also style changes to update focus styles & minor spacing fixes.

**Screenshots**

Documentation, using the page from https://github.com/WordPress/wporg-documentation-2022/issues/80:

| Default | Open menu | Mobile menu |
|---|---|---|
|  |  |  |

The bottom border should only show when scrolling:

https://github.com/WordPress/wporg-mu-plugins/assets/541093/c9225443-b58c-4be2-b07b-eafd621ca259

Here's an example of the menu collapsing when it would have wrapped (plus the "L3 header" variant updates):
https://github.com/WordPress/wporg-mu-plugins/assets/541093/71c069fd-9a7c-4c28-9e27-50c385faa286

Some long page titles still overflow the space available even with the collapsed nav menu. I've just set them to overflow:hidden, and at 550px wide, the page title is hidden.

| Full title visible | Overflow kicks in | Small screen, title hidden |
|---|---|---|
|  |  |  |

**To test**

Check out this PR and use it with any site, it doesn't _require_ any local changes.

I'll have a separate PR for Developer to implement the different page level header templates, which was used to capture some of these screenshots.

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/src/view.js
+++ b/mu-plugins/blocks/local-navigation-bar/src/view.js
@@ -53,7 +53,7 @@ function init() {
 			let navWidth = container.dataset.navWidth;
 			if ( ! navWidth ) {
 				const navElement = container.querySelector( 'nav:not(.wporg-is-collapsed-nav)' );
-				const navGap = ( window.getComputedStyle( navElement ).gap || '20px' ).replace( 'px', '' ) * 1;
+				const navGap = parseInt( window.getComputedStyle( navElement ).gap, 10 ) || 20;
 				// Get the nav width based on items, so that it stays
 				// consistent even if the menu wraps to a new line.
 				const menuItems = navElement.querySelectorAll( '.wp-block-navigation__container > li' );
@@ -76,9 +76,9 @@ function init() {
 
 			const availableWidth =
 				window.innerWidth -
-				paddingInlineStart.replace( 'px', '' ) * 1 -
-				paddingInlineEnd.replace( 'px', '' ) * 1 -
-				gap.replace( 'px', '' ) * 1;
+				parseInt( paddingInlineStart, 10 ) -
+				parseInt( paddingInlineEnd, 10 ) -
+				parseInt( gap, 10 );
 
 			const titleElement = container.querySelector( '.wp-block-site-title, div.wp-block-group' );
 			if ( ! titleElement ) {
--- a/mu-plugins/blocks/local-navigation-bar/src/view.js
+++ b/mu-plugins/blocks/local-navigation-bar/src/view.js
@@ -52,17 +52,17 @@ function init() {
 			// so that the uncollapsed visible menu's width is used.
 			let navWidth = container.dataset.navWidth;
 			if ( ! navWidth ) {
+				const navElement = container.querySelector( 'nav:not(.wporg-is-collapsed-nav)' );
+				const navGap = ( window.getComputedStyle( navElement ).gap || '20px' ).replace( 'px', '' ) * 1;
 				// Get the nav width based on items, so that it stays
 				// consistent even if the menu wraps to a new line.
-				const menuItems = container.querySelectorAll(
-					'nav:not(.wporg-is-collapsed-nav) .wp-block-navigation__container > li'
-				);
+				const menuItems = navElement.querySelectorAll( '.wp-block-navigation__container > li' );
 				navWidth =
 					[ ...menuItems ].reduce(
 						( acc, current ) => ( acc += current.getBoundingClientRect().width ),
 						0
 					) +
-					20 * ( menuItems.length - 1 ); // 20px gap between items.
+					navGap * ( menuItems.length - 1 ); // 20px gap between items.
 
 				// Save the value for future resize callbacks.
 				container.dataset.navWidth = Math.ceil( navWidth );
--- a/mu-plugins/blocks/local-navigation-bar/src/view.js
+++ b/mu-plugins/blocks/local-navigation-bar/src/view.js
@@ -36,6 +36,64 @@ function init() {
 
 		document.addEventListener( 'scroll', debounce( onScroll ), { passive: true } );
 		onScroll();
+
+		// Check the size of child elements to determine if the local navigation
+		// menu should be collapsed in mobile-view by default. If so, toggle a
+		// CSS class to show the nav block with {"overlayMenu":"always"}
+		// added by `add_extra_navigation`.
+		const onResize = () => {
+			// Bail early on small screens, the visible nav block is already mobile.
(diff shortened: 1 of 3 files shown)


mu-plugins/blocks/local-navigation-bar/src/view.js
```diff
@@ -36,6 +36,64 @@ function init() {
...
+				gap = '0px',
+			} = window.getComputedStyle( container );
+
+			const availableWidth =
+				window.innerWidth -
+				paddingInlineStart.replace( 'px', '' ) * 1 -
+				paddingInlineEnd.replace( 'px', '' ) * 1 -
+				gap.replace( 'px', '' ) * 1;
+
+			const titleElement = container.querySelector( 'p.wp-block-site-title, div.wp-block-group' );
```
StevenDufresne: Noting that `p.wp-block-site-title` is specific and we would be then checking `getBoundingClientRect` on a `null` object.

ryelle: Ah, because we're using that as the heading for showcase mobile… I assumed this would always be a `p` because the page's real `h1` appears in the content (this is true for Developer, Documentation, …everything but Showcase). I've asked about whether Showcase is an exception or another variation https://github.com/WordPress/wporg-mu-plugins/issues/535#issuecomment-1942110550

adamwoodnz:
Looking great, still testing, but found a broken width on Developer, looks like padding causing the wrapping

adamwoodnz:
Seems unnecessary to me to show the title on the homepage on tablet and mobile, especially as the local nav isn't sticky on those screen sizes, so it isn't visible once scrolled

| Tablet | Mobile | Scrolled |
|-|-|-|
| developer wordpress org_(iPad) | developer wordpress org_(Samsung Galaxy S20 Ultra) | developer wordpress org_(Samsung Galaxy S20 Ultra) (1) |

adamwoodnz:
From #535 

> For level 3 and deeper pages, page name is hidden as breadcrumb takes that role. But once you scroll down, all pages, except for the section’s landing pages, show the page name.

I think we're missing this behaviour, for me the page name is visible in the local nav before scrolling

https://github.com/WordPress/wporg-mu-plugins/assets/1017872/c897c639-1582-4854-b41c-35db80bf3de3

mu-plugins/blocks/local-navigation-bar/src/view.js
```diff
@@ -36,6 +36,64 @@ function init() {
...
+
+			const {
+				paddingInlineStart = '0px',
+				paddingInlineEnd = '0px',
+				gap = '0px',
+			} = window.getComputedStyle( container );
+
+			const availableWidth =
+				window.innerWidth -
+				paddingInlineStart.replace( 'px', '' ) * 1 -
```
adamwoodnz: I have always used [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) for this type of conversion, not sure what's better, but maybe it's easier to understand the intention. Not a suggestion, just an illustration:
```suggestion
				parseInt(paddingInlineStart, 10) -
```

adamwoodnz (approved):
This is great 👍 

Fixes the Docs bug for me and nails most of the new behaviour. I've left comments for things I've found but none are blockers imo.

ryelle:
> I think we're missing this behaviour, for me the page name is visible in the local nav before scrolling

I thought the animation of both the W and the page title was a little strange, so I'm waiting for some feedback on the main issue (see [the comment here](https://github.com/WordPress/wporg-mu-plugins/issues/535#issuecomment-1939906083)).

> Seems unnecessary to me to show the title on the homepage on tablet and mobile, especially as the local nav isn't sticky on those screen sizes, so it isn't visible once scrolled

Could you bring that up on the main issue? https://github.com/WordPress/wporg-mu-plugins/issues/535

mu-plugins/blocks/local-navigation-bar/src/view.js
ryelle: Personally I prefer `Number` over `parseInt` (the 10 always throws me off 🤷🏻), but this is a good point, I'll update it.

ryelle: Ah, `parseInt` parses up to a non-number value, so I wouldn't need to remove the `'px'`… Okay, that's probably a good reason to use that here :)

mu-plugins/blocks/local-navigation-bar/src/view.js
ryelle: Addressed in https://github.com/WordPress/wporg-mu-plugins/pull/573/commits/9c8edd9ec319bfae2bc8a014cec6e56befbe6928

mu-plugins/blocks/local-navigation-bar/src/view.js
ryelle: Done in https://github.com/WordPress/wporg-mu-plugins/pull/573/commits/8e5458bd30923943a44636fefc5d1daa145090ed

ryelle:
> Looking great, still testing, but found a broken width on Developer, looks like padding causing the wrapping

Fixed with https://github.com/WordPress/wporg-mu-plugins/pull/573/commits/bba37134b2e28db842aa9a00976b5b170485a6fb

> I think we're missing this behaviour, for me the page name is visible in the local nav before scrolling

I've added a class to fade-in the page title, and updated developer to use it. It'll look like this: 

https://github.com/WordPress/wporg-mu-plugins/assets/541093/84e87956-4a7b-48c6-b84d-e87506b38605

I think that addresses all the feedback here, but let me know if I missed something.

outdoor2kode (approved):
Tested and looks good 👍

Issue #579, opened 2024-02-15
Blocks: Set `version` property in block.json


dd32:
Currently we don't define the `version` field in the block.json file, which causes WordPress to use `$wp_version` as the version/cache-buster for CSS files.

This causes some less than ideal scenario's when testing, as the old CSS gets loaded until a new core commit occurs.

It appears that setting the `version` block.json field will result in the CSS being appropriately cache busted.

I would suggest we use the hash of all files in `build/` once built.

Note: JS files are busted using the value from `build/view.asset.php`.

PR #580, opened 2024-02-15
Build: Append the hash of the build to block.json for CSS cache busting purposes.

dd32:
See #579

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -5,13 +5,14 @@
  */
 const chalk = require( 'chalk' );
 const fs = require( 'fs' ); // eslint-disable-line id-length
+const { sync: glob } = require( 'fast-glob' );
+const { hashElement } = require( 'folder-hash' );
 const path = require( 'path' );
 const postcss = require( 'postcss' );
 const rtlcss = require( 'rtlcss' );
 const { sync: resolveBin } = require( 'resolve-bin' );
 const { sync: spawn } = require( 'cross-spawn' );
 const postCssConfig = require( '../postcss.config.js' );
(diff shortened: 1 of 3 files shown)


dd32:
Something I'm not sure about here, It sets it in block.json, whether it's in `build/`, `src/` or `.`, as It turns out there isn't any consistency.

That does mean running the script on a development machine will dirty the git checkout.

I guess this could be limited to running when it's in a production target or in a github action?

ryelle:
I noticed this wasn't catching every instance of `block.json` — the `global-header-footer` project has two blocks in it, so there are two files to find. I updated the PR to use `glob`, which is what wp-scripts uses to find the block.jsons to build.

> Something I'm not sure about here, It sets it in block.json, whether it's in build/, src/ or ., as It turns out there isn't any consistency.

I think we can safely set this in just the build block.jsons (https://github.com/WordPress/wporg-mu-plugins/pull/580/commits/18384883665427c94dacbda6c306306fa1d67046). The exception currently is `screenshot-preview`, but that doesn't register the styles via block.json anyway, it uses the filemtime for style.css version. I'm more inclined to "fix" screenshot-preview than have the `build` change the tracked files.

It looks like `global-header-footer`, `screenshot-preview`, and `horizontal-slider` all register the CSS separately from block registration, and use filemtime — for the latter two I think that's because they're not being used as real blocks? They're both on the current theme directory. In any case, when testing those versions will not update with the build hash (I don't think we need to change this ATM, since there is a reason, just writing it out).

ryelle:
Merging to get a `version` into the blocks on production.

PR #584, opened 2024-03-07
Add check to avoid fatals on direct file access

pkevan:
Avoids fatal errors when these files are accessed directly e.g.

[ja.wordpress.org](http://ja.wordpress.org/) 198.143.164.10 - [07/Mar/2024:09:04:13 +0000] "GET /cba94a84/wp-content/mu-plugins/pub-sync/blocks/time/ HTTP/1.1"


StevenDufresne:
@pkevan Need a review for this?

pkevan:
sure - can't remember doing this!

StevenDufresne:
Looks good to me, but added @dd32 and @ryelle seeing that they have more WP experience than myself :)

dd32:
Originally I suggested renaming these, or disabling direct PHP access at the server level, but that ran into issues (mostly, that the allow list is outdated).

I have no concerns about merging this, other than me not feeling it's warranted to have a check at the top of every `.php` file.

dd32 (approved):

Issue #589, opened 2024-03-13
Styling bug in the Site Breadcrumbs block


ndiego:
I noticed this in the Showcase: https://wordpress.org/showcase/submit-a-wordpress-site/thanks/

There is a strange gradient effect on the `:after` pseudo element

<img width="1045" alt="image" src="https://github.com/WordPress/wporg-mu-plugins/assets/4832319/6e02f092-f0fe-4923-a349-03f367cd7d47">

PR #593, opened 2024-03-28
resolve-breadcrump-issue

NidhiDhandhukiya74:
Resolve issue of :- #589

### DIFF
--- a/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
+++ b/mu-plugins/blocks/site-breadcrumbs/postcss/style.pcss
@@ -7,7 +7,6 @@
 		position: absolute;
 		inset: 0 0 0 auto;
 		width: 60px;
-		background-image: linear-gradient(to left, #ffff, #fff0);
 	}
 
 	& .wporg-site-breadcrumbs__wrapper {


ryelle:
Thanks for the PR, @NidhiDhandhukiya. We do need to keep that gradient though (in some form), it's used to show that the breadcrumbs overflows on smaller screens.

For example, in [this developer documentation article](https://developer.wordpress.org/themes/global-settings-and-styles/settings/use-root-padding-aware-alignments/):

Screenshot 2024-03-28 at 1 35 19 PM

ryelle (changes requested):
Thanks for the PR, @NidhiDhandhukiya. We do need to keep that gradient though (in some form), it's used to show that the breadcrumbs overflows on smaller screens.

For example, in [this developer documentation article](https://developer.wordpress.org/themes/global-settings-and-styles/settings/use-root-padding-aware-alignments/):

Screenshot 2024-03-28 at 1 35 19 PM

ryelle:
This was fixed in https://github.com/WordPress/wporg-mu-plugins/commit/288f29cb6d0b57ea3972174d2b9ecca5b6c2d651.

PR #595, opened 2024-04-02
Override the visited link color for affected blocks

adamwoodnz:
See https://github.com/WordPress/wporg-parent-2021/issues/128

Visited link color is being added to the parent theme. This PR overrides that color for blocks which have specific styles.


adamwoodnz:
Not required now that we're only targeting links in `.entry-content`, see [f92fe1a](https://github.com/WordPress/wporg-parent-2021/pull/129/commits/f92fe1ab0de0dfaa9e3eca0cc2324bcbce96afce)

PR #596, opened 2024-04-03
Query Filter: Update style and behavior of single-select filters

### LINKED ISSUE #594: Query Filter block: Update "single select" variation
See https://github.com/WordPress/pattern-directory/issues/648

Single selection flow

Design: ["Single selection" section](https://www.figma.com/file/SQwX90pGxS0XlNTEyNhDRE/WordPress.org-Navigation?type=design&node-id=1006-756&mode=design&t=vqRm8SmEnAJmUkCc-11) in Figma file

ryelle:
Fixes #594 — This implements the new design for single-select filters, which removes "Clear" and shows a heading in the dropdown at all screen sizes. This also changes the selected style from an ✖️  to a ✔️  on multiple-choice filters, since the same applies to single.

This has a corresponding patterns PR which updates the labels: https://github.com/WordPress/pattern-directory/pull/656

**Screenshots**

|  | Screenshot | 
|---|---|
| Home | home-default |
| "Filter" open | home-filter-open |
| "Sort" open | home-sort-open |
| All patterns | all-patterns-default |
| All patterns, with filters applied | all-patterns-filtered |
| Multiple select, with some selected | multiple-open |
| Multiple select, none selected | multiple-open-empty |
| Mobile (modal view) | modal-view |

Part of the design had "Apply" disabled until new items were checked, but we can't really do that. There's no way to differentiate between a 3 items already selected, vs 2 already selected + 1 new — they're all just selected or not.

**To test**

- Apply this PR in a project that uses query filters— showcase, patterns, events
- Open a single-select filter
- There should be no clear button
- The selected item should have a ✔️ 
- Open a multiple-select filter
- There should be a clear button
- The selected items should have a ✔️ 

The functionality should not change, applying (and clearing multi-select) filters should work.

### DIFF
--- a/mu-plugins/blocks/query-filter/render.php
+++ b/mu-plugins/blocks/query-filter/render.php
@@ -57,6 +57,15 @@ $button_classes = array_keys(
 	)
 );
 
+$modal_classes = array_keys(
+	array_filter(
+		array(
+			'wporg-query-filter__modal' => true,
+			'is-single-select' => ! $has_multiple,
+		)
+	)
+);
+
 if ( $selected_count && $has_multiple ) {
 	/* translators: %s is count of currently selected filters. */
 	$apply_label = sprintf( __( 'Apply (%s)', 'wporg' ), $selected_count );
@@ -87,7 +96,7 @@ if ( $selected_count && $has_multiple ) {
 	></div>
 
 	<div
-		class="wporg-query-filter__modal"
+		class="<?php echo esc_attr( implode( ' ', $modal_classes ) ); ?>"
 		id="<?php echo esc_attr( $html_id ); ?>"
 		data-wp-bind--hidden="!context.isOpen"
 		data-wp-effect--focus="effects.focusFirstElement"
@@ -146,6 +155,7 @@ if ( $selected_count && $has_multiple ) {
 			?>
 
 			<div class="wporg-query-filter__modal-actions">
+				<?php if ( $has_multiple ) : ?>
 				<input
 					type="button"
 					class="wporg-query-filter__modal-action-clear"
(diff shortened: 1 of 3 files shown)


mu-plugins/blocks/query-filter/render.php
```diff
@@ -146,15 +155,18 @@ class="wporg-query-filter__modal-close"
 			?>
 
 			<div class="wporg-query-filter__modal-actions">
+				<?php if ( $has_multiple ) : ?>
```
StevenDufresne, line 158: This throws a JS exception since the clear button is conditional:
https://github.com/WordPress/wporg-mu-plugins/blob/21c8d937b1c6003d5fb5a2598278fd1d7905a547/mu-plugins/blocks/query-filter/src/view.js#L73

mu-plugins/blocks/query-filter/src/style.pcss
```diff
@@ -112,9 +112,26 @@
 	color: var(--wp--custom--wporg-query-filters--color--text);
 	display: none;
 
-	& .wporg-query-filter__modal-header {
+	&:where(:not(.is-single-select)) .wporg-query-filter__modal-header {
 		display: none;
 	}
+
+	& .wporg-query-filter__modal-header {
+		padding-block-start: var(--wp--preset--spacing--10);
```
StevenDufresne: Can we make the padding match the left side:

| Current | Proposal |
|--------|--------|
| Screenshot 2024-04-04 at 9 56 02 AM | Screenshot 2024-04-04 at 9 54 38 AM  |

ryelle: Looks like that actually should be 18px by the design, so I'll get that updated 👍🏻

mu-plugins/blocks/query-filter/render.php
ryelle: Checking for `clearButton` now, in https://github.com/WordPress/wporg-mu-plugins/pull/596/commits/dcd4af908577d1250487c8bcd93c2115d651e78b

mu-plugins/blocks/query-filter/src/style.pcss
ryelle: Updated the padding in https://github.com/WordPress/wporg-mu-plugins/pull/596/commits/eaa32e0e5ba3b9e54c4a35c25a2d0d6b6d1640b1

StevenDufresne (approved):
👍

PR #597, opened 2024-04-04
Local Navigation Bar: Update scroll offset to prevent local nav bar overlapping content

### LINKED ISSUE #586: Local Nav: Should adjust scroll positioning
The global header adjusts `html { --wp-global-header-offset: ... }` such that anchored links do not fall under the sticky header.

The local nav doesn't do this however, resulting in anchor links that hide behind a sticky header.

For example, this will expand the first jetpack FAQ item, and the heading of it will not be visible.

https://test.wordpress.org/plugins/jetpack/#is%20jetpack%20free%3F

ryelle:
Fixes #586. The local nav bar overlaps content when using anchor links. There is [some code in the parent theme](https://github.com/WordPress/wporg-parent-2021/blob/624ac76935bb9c072e72ecde9b465218852c92e5/source/wp-content/themes/wporg-parent-2021/sass/block-styles.scss#L524-L529) that used to work for the sticky header, but since we unstuck the header in favor of the local nav bar, this doesn't apply anymore.

I've taken the same approach here, but put it in the Local Navigation Bar's CSS, so that it will load whenever that block is used. This adds `scroll-padding-top` to `html`, to account for the local nav bar + admin bar (if it exists).

Some child themes have added `scroll-margin-top` to individual elements, this will add extra space until it's removed, but I don't think that needs to block rolling this fix out.

**Screenshots**

Developer

| With `scroll-margin-top` on TOC headings | Without `scroll-margin-top` |
|---|---|
| Screen Shot 2024-04-04 at 13 00 34 | Screen Shot 2024-04-04 at 13 00 49 |

Forums, with the local `scroll-margin-top` disabled:

forum

Plugin FAQs

plugins

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -1,5 +1,5 @@
 /* Set up the custom properties. These can be overridden by settings in theme.json. */
-:where(body) {
+:where(html) {
 	--wp--custom--local-navigation-bar--spacing--height: 60px;
 }
 
@@ -318,8 +318,7 @@
 }
 
 @media (min-width: 890px) {
-	/* stylelint-disable selector-id-pattern */
-	#wp--skip-link--target {
-		scroll-margin-top: var(--wp--custom--local-navigation-bar--spacing--height, 0);
+	html {
+		scroll-padding-top: calc(var(--wp--custom--local-navigation-bar--spacing--height, 0px) + var(--wp-admin--admin-bar--height, 0px));
 	}
 }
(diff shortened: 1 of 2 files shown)


adamwoodnz (approved):
LGTM

Visually I think it's nice to have a little more space above the linked element, but maybe it's up to individual themes to set that. For example the Dev Resources title looks better with 20px space to the local nav so that the content is vertically aligned. Modifying the local `scroll-margin-top` achieves this.

developer wordpress org_block-editor_(2 1366x768)

ryelle:
Yeah, I think that should be up to the individual themes. I've updated the ToC heading offset to 20px, since that's part of the mu-block, and left the regular offset to exactly the heading height. I'm going to merge this so that the plugins FAQs are offset correctly, and then the margin can be removed from wporg-support-2024.

PR #599, opened 2024-04-08
New block: Query "Has results" to add extra content around query loops

ryelle:
See https://github.com/WordPress/pattern-directory/issues/636 – We have a query loop with a heading, but the heading does not make sense if there are no results in the query. This new block displays its contents conditionally, if the parent query loop has results.

**Screenshots**

In the editor, this behaves exactly like the "No results" block. You can add it to a Query Loop, and any content inside it will only show up if the query has results.

editor

If there are posts, the heading and paragraph appear:

has-results

If there are no posts, the "no results" content appears:

no-results

**To test**

- Add a Query Loop to a page
- Inside the query loop, add the "Has results" block
- Add content inside "Has results"
- The content should only show up on the frontend when there are posts in the query loop

Alternately, test with https://github.com/WordPress/pattern-directory/pull/657

### DIFF
--- a/mu-plugins/blocks/query-has-results/index.php
+++ b/mu-plugins/blocks/query-has-results/index.php
@@ -1,0 +1,74 @@
+<?php
+/**
+ * Block Name: Has results
+ * Description: Contains the block elements to display when there are query results.
+ *
+ * @package wporg
+ */
+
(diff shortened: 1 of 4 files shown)


StevenDufresne (approved):
LGTM

PR #605, opened 2024-04-23
Screenshot Preview Block: Split React code and block, use Interactivty API for block

ryelle:
This PR pulls out the block code from the existing `blocks/screenshot-preview` folder, and brings in the screenshot block from the pattern directory (using the Interactivity API). The new block lives in `blocks/screenshot-preview-block`. Spitting the folder like this lets us keep the current React code & `__wporg_screenshot_preview_render` in place, while using the newer block for the new themes (to avoid changes to the current Theme Directory).

The new block is used like this. See [the readme for all attributes](https://github.com/WordPress/wporg-mu-plugins/blob/51e1c2129843bdd2bca77ab875e0f89f4c7c2c4e/mu-plugins/blocks/screenshot-preview-block/README.md#attributes)

<table role="table">
<thead><tr><th>Code</th><th>Screenshot</th></tr></thead>
<tbody><tr>
<td width="50%"><code>&lt;!-- wp:wporg/screenshot-preview {"src":"https://developer.wordpress.org/","href":"https://developer.wordpress.org/","alt":"WordPress Developer Resources","align":"wide"} /--&gt;</code></td>
<td width="50%">Screenshot 2024-04-23 at 4 14 11 PM</td>
</tr>
</tbody>
</table>

After the Theme Directory refresh is launched, we can remove `blocks/screenshot-preview`. That should be the only place this is used.

We could also simplify the pattern-thumbnail block by using this one, but that still needs to be a separate block as it uses a dynamic URL from the current pattern.

**Screenshots**

This is in use on the in-progress theme directory (See https://github.com/WordPress/wordpress.org/pull/279 about the overly tall images)

Screenshot 2024-04-23 at 3 28 16 PM

Screenshot 2024-04-23 at 3 28 05 PM

**To test**

1. Manually add the block to a page or template (there is no UI yet, maybe in the future)
2. Try using different attributes when loading the image
3. It should show a loading indicator, then show the screenshot
4. If the src URL is missing or invalid (ex, `localhost`), it should not output anything

### DIFF
--- a/mu-plugins/blocks/screenshot-preview-block/README.md
+++ b/mu-plugins/blocks/screenshot-preview-block/README.md
@@ -28,8 +28,10 @@ echo do_blocks( '<!-- wp:wporg/screenshot-preview {"src":"https://wordpress.org/
 
 | Name          | Type    | Description                                | Default |
 |---------------|---------|--------------------------------------------|---------|
-| alt           | string  | Alt text for image.                        | ""      |
-| href          | string  | Destination for link wrapper, if provided. | ""      |
-| src           | string  | Source (website) to capture for display    | ""      |
-| width         | integer | Image width                                | 800     |
-| viewportWidth | integer | Viewport width                             | 1200    |
+| alt            | string  | Alt text for image.                         | ""      |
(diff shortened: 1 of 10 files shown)


adamwoodnz (commented):
Sorry for late review, LGTM ✨

PR #606, opened 2024-05-01
Fix alignment of dropdown caret in the Local Navigation Bar block.

### LINKED ISSUE #604: Local Navigation: Fix dropdown caret alignment on mobile
This is a minor thing, but the dropdown caret in the local navigation should probably be aligned with the double-line "hamburger" icon in the primary navigation. 

| Current | Proposed | 
|-|-|
|image|image|

cc @WordPress/meta-design

ndiego:
Fixes https://github.com/WordPress/wporg-mu-plugins/issues/604

The dropdown "caret" was misaligned in the Local Navigation Bar block on mobile. It should line up with the = icon in the global header. 

| Current | Proposed |
|-|-|
|image|image|

And here's a video of the fix in action. 

https://github.com/WordPress/wporg-mu-plugins/assets/4832319/f6517c9f-e3da-4692-8ba7-2b206eac82f8

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -113,6 +113,15 @@
 		/* Needed for aligning the dropdown caret correctly. */
 		padding-right: 0;
 
+		/* Add right padding if the dropdown caret is not being displayed. */
+		&:not(.wporg-show-collapsed-nav) {
+			& .wp-block-navigation {
+				&:not(.wporg-is-mobile-nav) {
+					padding-right: calc(16px + var(--wp--custom--alignment--scroll-bar-width));
+				}
+			}
+		}
+
 		& .global-header__wporg-logo-mark {
 			display: none;
 		}
@@ -184,11 +193,6 @@
 
 	/* Navigation. */
 	& .wp-block-navigation {
-
-		/* Add right padding if the dropdown caret is not being displayed. */
-		&:not(.wporg-is-mobile-nav):not(.wporg-is-collapsed-nav) {
-			padding-right: calc(16px + var(--wp--custom--alignment--scroll-bar-width));
-		}
 
 		& .wp-block-navigation-submenu__toggle {
 			padding-block: calc(var(--wp--preset--spacing--10) / 2);
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -114,12 +114,8 @@
 		padding-right: 0;
 
 		/* Add right padding if the dropdown caret is not being displayed. */
-		&:not(.wporg-show-collapsed-nav) {
-			& .wp-block-navigation {
-				&:not(.wporg-is-mobile-nav) {
-					padding-right: calc(16px + var(--wp--custom--alignment--scroll-bar-width));
-				}
-			}
+		& nav:not(.wporg-is-mobile-nav):not(.wporg-is-collapsed-nav) {
+			padding-right: var(--wp--preset--spacing--edge-space);
 		}
 
 		& .global-header__wporg-logo-mark {
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -110,8 +110,8 @@
 		position: relative !important;
 		top: 0 !important;
 
-		/* Matches the padding of the global header button. */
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
```diff
@@ -184,6 +184,12 @@
 
 	/* Navigation. */
 	& .wp-block-navigation {
+
+		/* Add right padding if the dropdown caret is not being displayed. */
+		&:not(.wporg-is-mobile-nav):not(.wporg-is-collapsed-nav) {
+			padding-right: calc(16px + var(--wp--custom--alignment--scroll-bar-width));
```
ryelle: For some reason, the `.wp-block-navigation` class is on both the `nav` and the top `ul`, so this padding is being applied twice, and there shouldn't be space here- the li (white box) should line up with the dotted line.

Screenshot 2024-05-01 at 5 20 30 PM

ryelle: I think this is also affecting the code that checks if the menu will collide with the page title.

Screenshot 2024-05-01 at 5 28 05 PM

ryelle: This seems to work as expected (using the `edge-space` because this should line up with the content).

```
@media (max-width: 889px) {

	/* Add right padding if the dropdown caret is not being displayed. */
	&nav:not(.wporg-is-mobile-nav):not(.wporg-is-collapsed-nav) {
		padding-right: var(--wp--preset--spacing--edge-space);
	}
}
```

ryelle: The `view.js` will also need to be updated so that the navWidth calculation accounts for this new padding.

ryelle (changes requested):
I forgot about the double-nav implementation here, you didn't pick an easy task to jump in on 😅 

Looking good on small screens:

Screenshot 2024-05-01 at 5 28 38 PM

Except for About — this has some padding set on that block so that pattern should be updated once this is merged.
Screenshot 2024-05-01 at 5 22 42 PM

The mid-and-larger sizes need some updates still, though, left a thread of comments.

ndiego:
> I forgot about the double-nav implementation here, you didn't pick an easy task to jump in on 😅

I knew it was too easy. I'll dig into this tomorrow. Thanks for reviewing.

ndiego:
@ryelle, thanks again for testing. I think I have solved the issues in the recent commit.

mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
```diff
@@ -110,8 +110,17 @@
...
+		padding-right: 0;
+
+		/* Add right padding if the dropdown caret is not being displayed. */
+		&:not(.wporg-show-collapsed-nav) {
+			& .wp-block-navigation {
+				&:not(.wporg-is-mobile-nav) {
+					padding-right: calc(16px + var(--wp--custom--alignment--scroll-bar-width));
+				}
+			}
+		}
```
ryelle: This still has the double-padding issue:

Screenshot 2024-05-02 at 10 24 16 AM

If you add the `nav` element, it should apply to just the top-level class. Also, previously this matched the content alignment (`edge-space`) not the icon, and I don't know if the change to match the icon when it's text was intentional or not. I don't have a strong opinion either way, it's only 4px :)

```suggestion
		&nav:not(.wporg-is-mobile-nav):not(.wporg-is-collapsed-nav) {
			padding-right: var(--wp--preset--spacing--edge-space); // or the calculated one
		}
```

ryelle (changes requested):

mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
ndiego:
```
&nav:not(.wporg-is-mobile-nav):not(.wporg-is-collapsed-nav) {
        padding-right: var(--wp--preset--spacing--edge-space); // or the calculated one
}
```
This did not work for me, but `& nav:not....` did. Just updated the PR with this change.

ryelle (approved):
Looks good now! Thanks for keeping at it 🙂

ndiego:
> Looks good now! Thanks for keeping at it 🙂

😅 And thanks for all your help!

StevenDufresne:
It looks like this happened again?

https://meta.trac.wordpress.org/ticket/7656

ryelle:
Looks like we still need to do [this step](https://github.com/WordPress/wporg-mu-plugins/pull/606#pullrequestreview-2034481832):

> Except for About — this has some padding set on that block so that pattern should be updated once this is merged.

PR #609, opened 2024-05-13
REST API: Add locale suggestion endpoint & helpers for plugins and themes

ryelle:
See https://github.com/WordPress/wporg-theme-directory/pull/52 — I've moved `rest-api-locale.php` to `mu-plugins/helpers/locale.php`, and the endpoint to `rest-api/endpoints/`. In fact, I've updated the code to handle both theme and plugin requests, so that all this locale-banner-related code can stay in one place (this isn't exactly what was requested on the other PR, but once I started building it out as two similar endpoints, it made sense to share a base class & live here).

The endpoints:

- https://wordpress.org/themes/wp-json/wporg-themes/v1/locale-banner/ — Main theme directory banner
- https://wordpress.org/themes/wp-json/wporg-themes/v1/locale-banner/pulitzer — A specific theme request
- https://wordpress.org/plugins/wp-json/wporg-plugins/v1/locale-banner/ — Main plugin directory banner
- https://wordpress.org/plugins/wp-json/wporg-plugins/v1/locale-banner/recipe-block/ — A specific plugin request

These can also be used from Rosetta URLs

- https://eo.wordpress.org/themes/wp-json/wporg-themes/v1/locale-banner/
- https://eo.wordpress.org/plugins/wp-json/wporg-plugins/v1/locale-banner/recipe-block/

The routes should only appear on the relevant sites, so you can't request `wordpress.org/themes/wp-json/wporg-plugins/v1/locale-banner/`.

This leaves open the possibility to add a `patterns` variation later, if we wanted to.

**To test**

Same instructions as https://github.com/WordPress/wporg-theme-directory/pull/52

> Try out various combinations of header locale (Accept-Language header), theme, rosetta origin site, etc. This needs a sandbox to test for the locale database.

You can also try with the `try/mu-plugins-locale-api` branch on [wporg-theme-directory](https://github.com/WordPress/wporg-theme-directory/compare/try/mu-plugins-locale-api), which should show the banner if your browser is set up to send alternate language suggestions.

There is also a `debug` parameter you can add to the endpoints to show more information about the request. For example,

```sh
$ curl 'https://es.wordpress.org/themes/wp-json/wporg-themes/v1/locale-banner/?debug=1' -H 'Accept-Language: en,ko-KR,es-ES,fr,eo,it,ru'

{
	"currentLocale": "es_ES",
	"suggestions": [
		"ko_KR",
		"es_ES",
		"fr_FR",
		"eo",
		"it_IT",
		"ru_RU"
	],
	"message": "The theme directory is also available in <a href=\"https:\/\/ko.wordpress.org\/themes\/\">한국어<\/a>, <a href=\"https:\/\/fr.wordpress.org\/themes\/\">Français<\/a>, <a href=\"https:\/\/eo.wordpress.org\/themes\/\">Esperanto<\/a>, <a href=\"https:\/\/it.wordpress.org\/themes\/\">Italiano<\/a>, 그리고 <a href=\"https:\/\/ru.wordpress.org\/themes\/\">Русский<\/a>."
}
```

### DIFF
--- a/mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
@@ -20,9 +20,7 @@ abstract class Base_Locale_Banner_Controller extends \WP_REST_Controller {
 				'callback' => array( $this, 'get_response' ),
 				'args' => array(
 					'debug' => array(
-						'sanitize_callback' => function( $param ) {
-							return (bool) $param;
-						},
+						'type' => 'boolean',
 					),
 				),
 				'permission_callback' => '__return_true',
@@ -36,9 +34,7 @@ abstract class Base_Locale_Banner_Controller extends \WP_REST_Controller {
 				'callback' => array( $this, 'get_response_for_item' ),
 				'args' => array(
 					'debug' => array(
-						'sanitize_callback' => function( $param ) {
-							return (bool) $param;
-						},
+						'type' => 'boolean',
 					),
 					'slug' => array(
 						'validate_callback' => array( $this, 'check_slug' ),
--- a/mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace WordPressdotorg\MU_Plugins\REST_API;
+
+/**
+ * Base_Locale_Banner_Controller
+ */
+class Base_Locale_Banner_Controller extends \WP_REST_Controller {
+	/**
+	 * Register the endpoint routes used across both themes and plugins.
+	 *
+	 * @see register_rest_route()
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			array(
+				'methods' => \WP_REST_Server::READABLE,
+				'callback' => array( $this, 'get_response' ),
+				'args' => array(
+					'debug' => array(
+						'sanitize_callback' => function( $param ) {
+							return (bool) $param;
+						},
+					),
+				),
+				'permission_callback' => '__return_true',
+			)
+		);
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<slug>[^/]+)/',
+			array(
+				'methods' => \WP_REST_Server::READABLE,
+				'callback' => array( $this, 'get_response_for_item' ),
+				'args' => array(
+					'debug' => array(
+						'sanitize_callback' => function( $param ) {
+							return (bool) $param;
+						},
+					),
+					'slug' => array(
+						'validate_callback' => array( $this, 'check_slug' ),
+					),
+				),
+				'permission_callback' => '__return_true',
+			)
+		);
+	}
+
+	/**
+	 * Check if the given slug is a valid item.
+	 *
+	 * Must be defined in the child class.
+	 */
+	public function check_slug( $param ) {
+		_doing_it_wrong(
+			'Base_Locale_Banner_Controller::check_slug',
+			/* translators: %s: register_routes() */
+			sprintf( "Method '%s' must be overridden.", __METHOD__ ),
+			'4.7.0'
+		);
+		return false;
+	}
+
+	/**
+	 * Send the response as plain text so it can be used as-is.
+	 */
+	public function send_plain_text( $result ) {
+		header( 'Content-Type: text/text' );
+		if ( $result ) {
+			echo '<div>' . $result . '</div>'; // phpcs:ignore
+		}
+
+		return null;
+	}
+}
--- a/mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
@@ -5,7 +5,7 @@ namespace WordPressdotorg\MU_Plugins\REST_API;
 /**
  * Base_Locale_Banner_Controller
  */
-class Base_Locale_Banner_Controller extends \WP_REST_Controller {
+abstract class Base_Locale_Banner_Controller extends \WP_REST_Controller {
 	/**
 	 * Register the endpoint routes used across both themes and plugins.
 	 *
@@ -54,15 +54,7 @@ class Base_Locale_Banner_Controller extends \WP_REST_Controller {
 	 *
 	 * Must be defined in the child class.
 	 */
-	public function check_slug( $param ) {
-		_doing_it_wrong(
-			'Base_Locale_Banner_Controller::check_slug',
-			/* translators: %s: register_routes() */
-			sprintf( "Method '%s' must be overridden.", __METHOD__ ),
-			'4.7.0'
-		);
-		return false;
-	}
+	abstract public function check_slug( $param );
 
 	/**
 	 * Send the response as plain text so it can be used as-is.
--- a/mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
+++ b/mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
@@ -122,22 +122,21 @@ class Plugins_Locale_Banner_Controller extends Base_Locale_Banner_Controller {
 		if ( 'en_US' !== $current_locale && $current_gp_locale && ! in_array( $current_locale, $translated_locales ) ) {
 			$output_locale = $current_locale;
 			switch_to_locale( $output_locale );
-			$suggest_string = sprintf(
-				// translators: %s: Locale name.
-				__( 'This plugin is not translated into %s yet.', 'wporg' ),
-				$current_gp_locale->native_name
-			);
 
-			// Append some other suggestions if they exist.
 			if ( ! empty( $suggestion_links ) ) {
-				$suggest_string .= ' ' . sprintf(
-					// translators: %s: List of links to plugin in other locales.
-					__( 'This plugin is available in %s.', 'wporg' ),
+				$suggest_string = sprintf(
+					// translators: %1$s: Locale name, %2$s: List of links to plugin in other locales.
+					__( 'This plugin is not translated into %1$s yet, but it is available in %2$s.', 'wporg' ),
+					$current_gp_locale->native_name,
 					wp_sprintf_l( '%l', $suggestion_links )
 				);
+			} else {
+				$suggest_string = sprintf(
+					// translators: %s: Locale name.
+					__( 'This plugin is not translated into %s yet.', 'wporg' ),
+					$current_gp_locale->native_name
+				);
 			}
-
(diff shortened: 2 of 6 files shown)


mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
```diff
@@ -0,0 +1,191 @@
...
+				$language
+			);
+		}
+
+		$suggest_string = '';
+
+		unset( $suggestion_links[ $current_locale ] );
+
+		// If we're on a rosetta site, and the plugin is not translated, the message should ask for help.
+		if ( 'en_US' !== $current_locale && $current_gp_locale && ! in_array( $current_locale, $translated_locales ) ) {
```
StevenDufresne, line 122: I'm curious why `$current_gp_locale` needs to be checked in this if statement. That's not to question its need, but as a future developer looking at this code, it isn't clear to me why this would be falsey and what should happen in that case. Minor.

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
```diff
@@ -0,0 +1,191 @@
...
+	public function __construct() {
+		$this->namespace = 'wporg-plugins/v1';
+		$this->rest_base = 'locale-banner';
+	}
+
+	/**
+	 * Validate the plugin slug.
+	 */
+	public function check_slug( $param ) {
+		return is_string( $param ) && $param && Plugin_Directory::get_plugin_post( $param );
```
StevenDufresne: Is the second`$param` check necessary?

mu-plugins/rest-api/endpoints/class-wporg-themes-locale-banner-controller.php
```diff
@@ -0,0 +1,190 @@
...
+		$locale_subdomain_assoc = get_all_locales_with_subdomain();
+		$current_locale = get_locale();
+		$current_gp_locale = \GP_Locales::by_field( 'wp_locale', $current_locale );
+
+		// Build a list of WordPress locales which we'll suggest to the user.
+		$suggest_locales = array_values( array_intersect( get_locale_from_header(), get_all_valid_locales() ) );
+
+		$suggestion_links = [];
+		foreach ( $suggest_locales as $locale ) {
+			$language = \GP_Locales::by_field( 'wp_locale', $locale )->native_name;
```
StevenDufresne, line 46: In [previous code](https://github.com/WordPress/wporg-mu-plugins/pull/609/files#diff-06764635da25e9390a050f0d6d05b3566f751bbdc8fa8f9e4afae04d17a107c5R122), we checked to make sure `$current_gp_locale` was not false which is a [possible return value](https://github.com/GlotPress/gp-locales/blob/cc5d93389d0d8ed473699900b29d31ddab27830e/locales.php#L2181). I assume we probably want to do it here as well.

mu-plugins/rest-api/endpoints/class-wporg-themes-locale-banner-controller.php
```diff
@@ -0,0 +1,190 @@
...
+		$current_locale = get_locale();
+		$current_gp_locale = \GP_Locales::by_field( 'wp_locale', $current_locale );
+		$translated_locales = get_translated_locales( 'theme', $theme_slug );
+
+		// Build a list of WordPress locales which we'll suggest to the user.
+		$suggest_locales = array_values( array_intersect( get_locale_from_header(), $translated_locales ) );
+
+		$suggestion_links = [];
+		foreach ( $suggest_locales as $locale ) {
+			$language = \GP_Locales::by_field( 'wp_locale', $locale )->native_name;
```
StevenDufresne, line 108: Same comment as https://github.com/WordPress/wporg-mu-plugins/pull/609/files#r1599220197

StevenDufresne:
I've taken a look at the code and left some minor suggestions. I don't know how much is moved code versus new so I'll let you determine what needs to be updated in regards to that. Otherwise, I feel comfortable merging this.

mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	public function check_slug( $param ) {
+		_doing_it_wrong(
+			'Base_Locale_Banner_Controller::check_slug',
+			/* translators: %s: register_routes() */
+			sprintf( "Method '%s' must be overridden.", __METHOD__ ),
+			'4.7.0'
+		);
+		return false;
+	}
```
dd32: This could just be made an abstract class, and this left as a function signature.

```suggestion
	/**
	 * Check if the given slug is a valid item.
	 *
	 * Must be defined in the child class.
	 */
	public function check_slug( $param );
```

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
dd32: It's worth including; it's effectively saying "Is this a string? And it's got text in it? And we're able to find a plugin post for it?" rather than expecting that `get_plugin_post()` properly handles `$slug = ''`.

That being said; I don't think it's needed at all, as the `slug` param should only be coming from the URL rewrite, which should always be a non-zero-length string?

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
```diff
@@ -0,0 +1,191 @@
...
+		$suggest_string = '';
+
+		unset( $suggestion_links[ $current_locale ] );
+
+		if ( ! empty( $suggestion_links ) ) {
+			$output_locale = key( $suggestion_links );
+			switch_to_locale( $output_locale );
+			$suggest_string = sprintf(
+				// translators: %s: List of links to plugin directory in other locales.
+				__( 'The plugin directory is also available in %s.', 'wporg' ),
```
dd32, line 64: I'm wondering if it's worthwhile making these available in the plugin directory translate project via `wporg-plugins` textdomain. It's not impossible to make that happen.

mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace WordPressdotorg\MU_Plugins\REST_API;
+
+/**
+ * Base_Locale_Banner_Controller
+ */
+class Base_Locale_Banner_Controller extends \WP_REST_Controller {
```
dd32:
```suggestion
abstract class Base_Locale_Banner_Controller extends \WP_REST_Controller {
```

To go with the later suggestion.

mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
```diff
@@ -0,0 +1,78 @@
...
+			'/' . $this->rest_base . '/(?P<slug>[^/]+)/',
+			array(
+				'methods' => \WP_REST_Server::READABLE,
+				'callback' => array( $this, 'get_response_for_item' ),
+				'args' => array(
+					'debug' => array(
+						'sanitize_callback' => function( $param ) {
+							return (bool) $param;
+						},
+					),
```
dd32: This can just use the inbuilt boolean sanitization I think? 
```suggestion
					'debug' => array(
						'type' => 'boolean'
					),
```

dd32 (approved):

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
ryelle: That's fair, I copied this one from [the plugin directory code](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/class-base.php#L62), but it's true that it won't be empty due to the rewrite regex. I've removed it to simplify the check.

mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
ryelle: Switched to an abstract class in https://github.com/WordPress/wporg-mu-plugins/pull/609/commits/2246fd26664f848eeb90f63c85fd2900e36d1b71

mu-plugins/rest-api/endpoints/class-wporg-base-locale-banner-controller.php
ryelle: Oh yeah, I forgot there's built-in checking for basic types. Done in https://github.com/WordPress/wporg-mu-plugins/pull/609/commits/05c7766df287a3018aa5139b8519b7517034e675.

mu-plugins/rest-api/endpoints/class-wporg-themes-locale-banner-controller.php
ryelle: The values from `get_all_valid_locales` are already checked and filtered out if they return false, so we know this will be OK.

mu-plugins/rest-api/endpoints/class-wporg-themes-locale-banner-controller.php
ryelle: Same as above, the results from `get_translated_locales` are filtered out if they don't have a GP_Locale response.

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
ryelle: To be honest I copied this from the current code, and I'm not totally sure when it would be false— I would _assume_ all rosetta sites correspond to a GP locale.

Well, it looks like it's false on `https://test.wordpress.org/plugins/wp-json/wporg-plugins/v1/locale-banner/recipe-block/`, so there's at least one place (test's locale uses `de`, but it should be `de_DE`). I don't know if any non-test sites trigger it.

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
ryelle: I have no strong feelings either way, but I don't know how to do that if it's more complicated than just changing the textdomain.

ryelle:
I've made the requested updates and I think it's good to merge now, but I'll wait until tomorrow in case there's any more feedback or comment replies.

mu-plugins/rest-api/endpoints/class-wporg-plugins-locale-banner-controller.php
dd32: After looking into this, I don't think it's worth it.

It'd require updating `translate/bin/generate-pot.sh` to create a duplicate pot file and then merge them, etc...

The only question I would make instead then is, `would it be worthwhile moving the child-classes of the abstract to be within the relevant directory code bases`, right now, I don't think it's worth it. If translators get confused about the string being missing, we can see about changes later on.

I think this is fine to merge @ryelle.

PR #616, opened 2024-05-31
Tools: Update packages

### LINKED ISSUE #608: Query Filters: Icons on wrong side on RTL sites
On RTL sites, the checkmarks are still on the left, when they should be before the text, on the right. Also, the down chevrons are on the right, when they should be after the text on the left. This might be fixed by adding an RTL CSS process step, which is currently missing.

Screenshot 2024-05-09 at 2 41 55 PM

ryelle:
Additionally, this fixes #608, as wp-scripts 27.9 introduces an RTLCSS build step

PR just to ensure the github workflows succeed.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -422,25 +422,27 @@
         },
         {
             "name": "nikic/php-parser",
-            "version": "v4.17.1",
+            "version": "v5.0.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/nikic/PHP-Parser.git",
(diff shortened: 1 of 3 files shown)


adamwoodnz:
So does this mean that if we update repos to 27.9 we can remove the custom rtl build config? [Example](https://github.com/WordPress/Learn/blob/trunk/wp-content/themes/pub/wporg-learn-2024/webpack.config.js).

ryelle:
Yep! but don't update to 28+, there are breaking changes there.

PR #618, opened 2024-06-04
Fixed missing quotes on the code is poetry src

SteveJonesDev:
This PR fixes the missing quotes on the "Code is Poetry" image src attribute in the footer. This was identified when running https://wordpress.org/ through the Equalize Digital Accessibility Checker plugin. The missing quotes cause false positives in this plugin.

Screenshot 2024-06-04 at 12 43 05 PM


ryelle (approved):
Thanks for the fix!

Issue #614, opened 2024-05-24
Remove links to https://mercantile.wordpress.org/ on WordPress.org


joemcgill:
Originally reported by [Dan Soschin](https://profiles.wordpress.org/dansoschin/) on https://core.trac.wordpress.org/ticket/61292.

> On WordPress.org, we have links to the Mercantile Store (swag store) located in the main navigation (About >> Swag Store) and the footer (4th column, 4th item).
> 
> The Mercantile store is currently closed and has no plans to reopen. (A separate discussion on that may occur in the future.) However, to support its closure, we should remove those two links to reduce inbound traffic.

Looks like [removing these lines](https://github.com/WordPress/wporg-mu-plugins/blob/4ab57d37d930c8b8526458fc510568d0fb0050f8/mu-plugins/blocks/global-header-footer/blocks.php#L577-L581) might do the job.

adamwoodnz:
There are also links in the [About](https://wordpress.org/about) page navigation, which I will also remove, issue created: https://github.com/WordPress/wporg-main-2022/issues/443

![Screenshot 2024-05-27 at 9 50 47 AM](https://github.com/WordPress/wporg-mu-plugins/assets/1017872/a26a10c8-3693-4247-8f38-93749e621bcf)

![Screenshot 2024-05-27 at 9 51 00 AM](https://github.com/WordPress/wporg-mu-plugins/assets/1017872/85bb651f-f8fd-4737-937e-a490c914a5a9)

adamwoodnz:
@WordPress/meta-design it's unfortunate that this leaves a hole in the footer links, anything we can do here?

![Screenshot 2024-05-27 at 10 31 31 AM](https://github.com/WordPress/wporg-mu-plugins/assets/1017872/2e12abf2-d349-46fe-91d6-a77111af1d87)

adamwoodnz:
We have a redirect set up that takes visitors from https://wordpress.org/about/swag to the mercantile store, what should happen instead @dansoschin?

jasmussen:
Last we curated there were many options we omitted for the sake of making room. The conversation is somewhere in the archive, but hard to find. Perhaps @thetinyl remembers?

adamwoodnz:
I've removed the links while we wait for answers on the above.

courtneyr-dev:
> @WordPress/meta-design it's unfortunate that this leaves a hole in the footer links, anything we can do here?
> 
> 
> 
[...]

PR #619, opened 2024-06-04
Add Five for the Future link to global footer

adamwoodnz:
See #614 

Replaces the Swag store links which have been removed while the store is closed.

Screenshot 2024-06-05 at 10 21 26 AM


thetinyl:
Five for the Future is already present in the global header under "Community". I don't think the "Swag" link in the global header needs to be replaced, personally. Since it's a dropdown, it doesn't look like a column that's come up short (compared to the footer, for example).

adamwoodnz:
> Five for the Future is already present in the global header under "Community". I don't think the "Swag" link in the global header needs to be replaced, personally. Since it's a dropdown, it doesn't look like a column that's come up short (compared to the footer, for example).

Thanks @thetinyl, nice catch. Now removed from header.

PR #620, opened 2024-06-05
Add tag name configurability to global footer block

adamwoodnz:
See https://github.com/WordPress/Learn/issues/2501 and https://wordpress.slack.com/archives/C01KC5VDQBC/p1717552327365119?thread_ts=1717455021.461299&cid=C01KC5VDQBC

On Learn there is a site footer above the global footer. This is currently between the `main` and the `footer` landmark provided by the global footer block. For accessibility reasons there shouldn't be content between these landmarks.

This PR adds the ability for the tagName to be configured so that the global footer can be nested within a footer element and maintain valid HTML.

Can be tested using https://github.com/WordPress/Learn/pull/2502

There should be no effect in other themes as the default value for the `tagName` attribute is set to `footer`.

### Before

Screenshot 2024-06-05 at 2 38 30 PM

### After 

Screenshot 2024-06-05 at 2 52 30 PM

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -847,8 +847,10 @@ function render_global_footer( $attributes, $content, $block ) {
 	$wrapper_attributes = get_block_wrapper_attributes(
 		array( 'class' => 'global-footer wp-block-group' )
 	);
+	$tag_name = $attributes['tagName'] ?? 'footer';
 	return sprintf(
-		'<footer %1$s>%2$s</footer>%3$s',
+		'<%1$s %2$s>%3$s</%1$s>%4$s',
+		$tag_name,
 		$wrapper_attributes,
 		$markup,
 		$footer_markup,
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -847,8 +847,10 @@ function render_global_footer( $attributes, $content, $block ) {
 	$wrapper_attributes = get_block_wrapper_attributes(
 		array( 'class' => 'global-footer wp-block-group' )
 	);
+	$tag_name = $attributes['tagName'] ?? 'footer';
```
ryelle: Do you think this should be validated somehow, maybe one of a list of possible tag names? I guess it would just be `div` or `footer`, in reality.

adamwoodnz: Yeah that's a good idea imo.

adamwoodnz: Added in [433b5c7](https://github.com/WordPress/wporg-mu-plugins/pull/620/commits/433b5c71c80ecadfd8eb9d8f6aa01271a17fd9ce)

PR #624, opened 2024-06-10
Maintain sidebar container compatibility with 3 column layout

adamwoodnz:
The parent theme 3 column layout has been updated for compatibility with Gutenberg 18.5.

This change makes the scrollbar container compatible. The logic is simplified so that the padding and margin aren't factors.

See https://github.com/WordPress/wporg-parent-2021/issues/139

Test using https://github.com/WordPress/wporg-documentation-2022/pull/97 and https://github.com/WordPress/wporg-developer/pull/522


ryelle (approved):
Alignment looks good when scrolling, section jumping still works as expected, and footer-collision is working 👍🏻

PR #626, opened 2024-06-13
Get the locale of current site and send it to lang-guess

### LINKED ISSUE #2428: Add locale suggestion banner
Add a locale suggestion banner to the header and front page header.

outdoor2kode:
Closes https://github.com/WordPress/Learn/issues/2428

The locale information for Learn comes from the query parameter "locale" rather than the URL subdomain, so this PR retrieves the locale and sends it to lang guess for processing.

lang-guess corresponding tweaks: [meta.trac.wordpress.org/ticket/7675](https://meta.trac.wordpress.org/ticket/7675)

### DIFF
--- a/mu-plugins/blocks/language-suggest/language-suggest.php
+++ b/mu-plugins/blocks/language-suggest/language-suggest.php
@@ -27,3 +27,20 @@ function language_suggest_block_init() {
 	);
 }
 add_action( 'init', __NAMESPACE__ . '\language_suggest_block_init' );
+
+function language_suggest_enqueue_scripts() {
+    wp_enqueue_script(
+        'language-suggest-front',
+        plugins_url( 'src/front.js', __FILE__ ),
+        array(),
+        null,
+        true
+    );
+
+    wp_add_inline_script(
+        'language-suggest-front',
+        'var LanguageSuggestData = ' . json_encode( array( 'locale' => get_locale() ) ) . ';',
+        'before'
+    );
+}
+add_action( 'plugins_loaded', __NAMESPACE__ . '\language_suggest_enqueue_scripts' );
--- a/mu-plugins/blocks/language-suggest/language-suggest.php
+++ b/mu-plugins/blocks/language-suggest/language-suggest.php
@@ -39,7 +39,7 @@ function language_suggest_enqueue_scripts() {
 
     wp_add_inline_script(
         'language-suggest-front',
-        'var languageSuggestData = ' . json_encode( array( 'locale' => get_locale() ) ) . ';',
+        'var languageSuggestData = ' . wp_json_encode( array( 'locale' => get_locale() ) ) . ';',
         'before'
     );
 }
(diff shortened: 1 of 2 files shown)


mu-plugins/blocks/language-suggest/src/front.js
```diff
@@ -1,14 +1,17 @@
...
 
-	const endpoint =
-		container.dataset.endpoint ||
-		'https://wordpress.org/lang-guess/lang-guess-ajax.php?uri=' +
-			encodeURIComponent( window.location.pathname );
+	const endpoint = new URL(
+		container.dataset.endpoint || 'https://wordpress.org/lang-guess/lang-guess-ajax.php'
+	);
+	endpoint.searchParams.set( 'uri', encodeURIComponent( window.location.pathname ) );
+	endpoint.searchParams.set( 'locale', encodeURIComponent( LanguageSuggestData.locale ) );
```
dd32: I think `set()` already encodes, but as none of the characters need encoding, I think this is safe to include.

mu-plugins/blocks/language-suggest/language-suggest.php
```diff
@@ -27,3 +27,20 @@ function language_suggest_block_init() {
...
+        'language-suggest-front',
+        plugins_url( 'src/front.js', __FILE__ ),
+        array(),
+        null,
+        true
+    );
+
+    wp_add_inline_script(
+        'language-suggest-front',
+        'var LanguageSuggestData = ' . json_encode( array( 'locale' => get_locale() ) ) . ';',
```
dd32: Probably best to use `wp_json_encode()` here just for sanity sake.
```suggestion
        'var LanguageSuggestData = ' . wp_json_encode( array( 'locale' => get_locale() ) ) . ';',
```

JS variables are usually camelCase, but I don't care enough to suggest it to be `languageSuggestData`.

dd32 (approved):

mu-plugins/blocks/language-suggest/src/front.js
outdoor2kode: yeah, you're right, none of the chars need it and `set()` already does enough encoding.
removed: 17fcac1

mu-plugins/blocks/language-suggest/language-suggest.php
outdoor2kode: Thanks. Should be camelCase, missed that.
`wp_json_encode`: 020a5cd
camelCase: 20258e3

ryelle:
It looks like this triggers a notice now, I'm seeing this on my local env:

> Notice: Function wp_enqueue_script was called incorrectly. Scripts and styles should not be registered or enqueued until the wp_enqueue_scripts, admin_enqueue_scripts, or login_enqueue_scripts hooks. This notice was triggered by the language-suggest-front handle. Please see Debugging in WordPress for more information.

PR #628, opened 2024-06-17
Language suggestion: Prevent loading script when not needed

ryelle:
See #626. That PR registered and enqueued the viewScript so that it appeared on every single page, regardless of whether the block was used. Instead, this attaches the data to the viewScript itself, so that it is rendered when the block loads the viewScript.

This also fixes the notice I mentioned in https://github.com/WordPress/wporg-mu-plugins/pull/626#issuecomment-2174428957; and fixes a missing translation context value in the block style.

To test

- Load a page with the language suggest banner
- You should see `language-suggest/build/front.js` in the network tab
- View source, and `languageSuggestData` should be defined
- No JS errors or warnings
- Load a page without the language suggest banner
- No `language-suggest/build/front.js` in the network tab or page source

### DIFF
--- a/mu-plugins/blocks/language-suggest/language-suggest.php
+++ b/mu-plugins/blocks/language-suggest/language-suggest.php
@@ -28,19 +28,14 @@ function language_suggest_block_init() {
 }
 add_action( 'init', __NAMESPACE__ . '\language_suggest_block_init' );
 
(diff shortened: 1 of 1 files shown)


outdoor2kode (approved):
Thanks for the fix!! It looks good on network tab and while sandboxed.

PR #630, opened 2024-06-18
Fonts: Add Noto Serif JP, faux `cjk` subset

### LINKED ISSUE #427: How to Implement Web fonts for a specific language for a specific locale
related #266 

Regarding design coordination with Japanese members, we would like to implement Japanese web fonts, as we cannot reproduce them with the fonts currently used on the site.

We need "Noto Serif JP" ( https://fonts.google.com/noto/specimen/Noto+Serif+JP ) weights 400, 600 (regular, semibold),
but, you know, Japanese font data is huge so that we might need the subset data.
If we can, I believe it's better way to use the dynamic subsetting.

Then, I would like to discuss with you what methods are realistic and possible.

I don't think Google's dynamic subsetting (or API) is available.
So, as for Japanese only, since Noto Serif JP has subset data prepared by individuals, I'm thinking of possibility using that as an alternative.
https://github.com/ixkaito/NotoSerifJP-subset

Thanks!

mentioned  @nukaga , @t-hamano

ryelle:
This PR adds the Noto Serif JP font from [noto-cjk](https://github.com/notofonts/noto-cjk/tree/main/Serif#downloading-noto-serif-cjk), using "Region Specific Subset OTFs Japanese (日本語)".

I've decided to use this version, instead of the subset version suggested in https://github.com/WordPress/wporg-main-2022/issues/427, because this is a variable font. This means we have access to all weights (just like EB Garamond) without having to load multiple font files. This version has also been updated since the subset versions were made.

The font file has been added to the global-fonts CSS, so that w.org sites can use this with `font-family: "Noto Serif JP";`. The font can also be marked for preloading with `global_fonts_preload( 'Noto Serif JP', 'cjk' );`.

Followup PRs to the parent theme and main theme are in progress to actually use the font.

Fixes https://github.com/WordPress/wporg-main-2022/issues/427

### DIFF
--- a/mu-plugins/global-fonts/style.css
+++ b/mu-plugins/global-fonts/style.css
@@ -1244,5 +1244,5 @@
 	font-weight: 200 900;
 	font-style: normal;
 	font-display: swap;
-	src: url(./Noto Serif/NotoSerifJP-cjk.woff2) format("woff2");
+	src: url(./NotoSerif/NotoSerifJP-cjk.woff2) format("woff2");
 }
--- a/mu-plugins/global-fonts/style.css
+++ b/mu-plugins/global-fonts/style.css
@@ -1233,3 +1233,16 @@
 	src: url(./IBMPlexSans/IBMPlexSans-SemiBold-arrows.woff2) format("woff2");
 	unicode-range: U+2190-2199;
 }
+
+/*------------------------------*
+ *  Noto Serif JP               *
+ *------------------------------*/
+
+/* cjk */
+@font-face {
+	font-family: "Noto Serif JP";
+	font-weight: 200 900;
+	font-style: normal;
+	font-display: swap;
+	src: url(./Noto Serif/NotoSerifJP-cjk.woff2) format("woff2");
+}
--- a/mu-plugins/global-fonts/index.php
+++ b/mu-plugins/global-fonts/index.php
@@ -135,7 +135,7 @@ function get_font_url( $font, $subset ) {
 	$lower_font   = strtolower( trim( $font ) );
 	$lower_subset = strtolower( trim( $subset ) );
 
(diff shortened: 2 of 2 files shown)


t-hamano:
Thank you for addressing this issue!

I tested this PR by doing the following:

- Checkout this branch locally and built it
- Added the following code to my theme:
  ```php
  function enqueue_cjk_fonts() {
  	global_fonts_preload( 'Noto Serif JP', 'cjk' );
  	wp_enqueue_style( 'wporg-global-fonts' );
  }
  add_action( 'init', 'enqueue_cjk_fonts' );
  ```

The font is preloaded, but the font family is not applied.

image

I investigated the reason and found that it might be because the folder contains a space.

This is because the space is encoded in the font URL:

```
<link rel='preload' href='http://localhost:8888/wp-content/mu-plugins/global-fonts/Noto%20Serif/NotoSerifJP-cjk.woff2' as='font' crossorigin='crossorigin' type='font/woff2' />
```

But in the stylesheet, the space remains a space.

As a test, I changed the directory name from `Noto Serif` to `NotoSerif` and removed the spaces from the relevant code, and the font family was applied correctly.

image

mu-plugins/global-fonts/style.css
```diff
@@ -1233,3 +1233,16 @@
...
+ *  Noto Serif JP               *
+ *------------------------------*/
+
+/* cjk */
+@font-face {
+	font-family: "Noto Serif JP";
+	font-weight: 200 900;
+	font-style: normal;
+	font-display: swap;
+	src: url(./Noto Serif/NotoSerifJP-cjk.woff2) format("woff2");
```
outdoor2kode: I think it's the same as @t-hamano's comment. I have to set this to 
```
./Noto\ Serif/NotoSerifJP-cjk.woff2
``` 
to make it work. Otherwise, the font isn't applied.

ryelle:
Aha, I had Noto Serif JP installed locally, so I missed the issue. I've renamed the folder to `NotoSerif` to match the other folders, which lack spaces, and it's working correctly now.

PR #633, opened 2024-06-24
Move and adjust ratings-bars and ratings-stars.

StevenDufresne:
We have ratings components on wp.org in there places:
 - Themes Single ([Example](https://wordpress.org/support/theme/twentytwentyfour/reviews/?filter=4))
 - Plugins Single ([Example](https://wordpress.org/plugins/gutenberg/))
 - Support Forums ([Reviews](https://wordpress.org/support/plugin/gutenberg/reviews/?filter=4))

This PR moves blocks added to https://github.com/WordPress/wporg-theme-directory to this project for use across the site.

## Approach
I decided to use block.json to expose the attributes and custom filters for themes to use to populate the values. I'm not 100% sure it's the best approach. I considered using [render_block_data](https://developer.wordpress.org/reference/hooks/render_block_data/) instead of a custom filter but thought this was more maintainable in the future and we use custom filters for other blocks. I can be convinced otherwise.

## wporg/ratings-stars 

```js
<!-- wp:wporg/ratings-stars /-->
```

### Attributes
| Attribute | Type |
|--------|--------|
| rating | number| 

## wporg/ratings-bars 

```js
<!-- wp:wporg/ratings-bars /-->
```

### Attributes
| Attribute | Type | Description |
|--------|--------|--------|
| ratingsCount | number |  The total of ratings, must match sum of all values in `ratings` |
| ratings | array(5) | Rating count. The array must have 5 items. [1] = 1 star, [5] = 5 star.
| slug | string | The slug of the theme or plugin. |
| supportUrl | string | Base url for support forums. |

### Filter Example

```php
use function WordPressdotorg\Theme\Theme_Directory_2024\{get_support_url, get_theme_information};

/**
 * Update ratings blocks with real rating data.
 *
 * @param array $data    Rating data.
 * @param int   $post_id Current post.
 *
 * @return array
 */
function set_rating_data( $data, $post_id ) {
	$theme = get_theme_information( $post_id );
	if ( ! $theme ) {
		return $data;
	}

	return array(
		'rating' => $theme->rating,
		'ratingsCount' => $theme->num_ratings,
		'ratings' => $theme->ratings,
		'supportUrl' => get_support_url( $theme->slug . '/reviews/' ),
	);
}
add_filter( 'wporg_ratings_data', __NAMESPACE__ . '\set_rating_data', 10, 2 );
```

## Testing this PR
You can test this PR in conjunction with https://github.com/WordPress/wporg-theme-directory/pull/146.

_That environment doesn't include the `ratings` table by default so it make makes sense to test in a sandbox._

1. Load this PR in your theme directory environment on [try/refactor-rating-stars](https://github.com/WordPress/wporg-theme-directory/tree/try/refactor-rating-stars)[try/refactor-rating-stars](https://github.com/WordPress/wporg-theme-directory/tree/try/refactor-rating-stars) 
2. Navigate to a theme, the reviews should be present


mu-plugins/blocks/ratings-bars/render.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+$attributes = apply_filters( 'wporg_block_ratings_bars', $attributes );
```
ryelle:
```suggestion
$attributes = apply_filters( 'wporg_ratings_bars_attributes', $attributes );
```

The rough naming scheme I've been using is snake-case block name for a prefix (so `wporg/ratings-bars` becomes `wporg_ratings_bars`) and a descriptor of the filter value (in this case, `attributes`).

Is filtering all of the attributes is necessary here? Other things like styles, alignment, etc will come through as well. Do you think we'll ever want to have a static `<!-- wp:wporg/ratings-stars {"rating":80} /-->` block? (other block, but easier to write 🙂 )

I would also give this a doc block following [the hooks doc standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters), you can see an example on [the wporg_query_filter_options_* filter](https://github.com/WordPress/wporg-mu-plugins/blob/652c3585fcc901116263eaf580753d916d6bf3ea/mu-plugins/blocks/query-filter/render.php#L6-L29).

ryelle: The filter should also pass back the post ID from the block context ($block->context['postId']), just to ensure the filtered ratings always pull from the correct post.

mu-plugins/blocks/ratings-bars/src/block.json
```diff
@@ -0,0 +1,32 @@
...
+			"default": []
+		},
+		"slug": {
+			"type": "string"
+		},
+		"support_url": {
+			"type": "string"
+		}
+	},
+	"editorScript": "file:./index.js",
```
ryelle, line 15:
```suggestion
	"usesContext": [ "postId" ],
	"editorScript": "file:./index.js",
```

Add the context back so that you can pass it through to the filter.

mu-plugins/blocks/ratings-bars/src/block.json
```diff
@@ -0,0 +1,32 @@
...
+			"type": "array",
+			"default": []
+		},
+		"slug": {
+			"type": "string"
+		},
+		"support_url": {
+			"type": "string"
+		}
+	},
```
ryelle: Typically block attributes use camelCase, so this should be `supportUrl` & `numRatings` (maybe `ratingsCount` or `totalRatings` would be a little more understandable? not a strong opinion, I know this reflects the theme object).

mu-plugins/blocks/ratings-stars/render.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+$attributes = apply_filters( 'wporg_block_ratings_stars', $attributes );
```
ryelle: Pretty much all the same comments apply to this block too— naming scheme, whether this should filter all of `$attributes` or just a placeholder value (like query total & query filters do), and it should also get the post ID from context to pass back.

mu-plugins/blocks/ratings-stars/index.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Block Name: Ratings (stars)
+ * Description: The avgerage rating of this theme displayed as stars.
```
ryelle:
```suggestion
 * Description: The average rating of this theme displayed as stars.
```
Pretty sure this typo's my fault  😆

ryelle (changes requested):
The blocks do work, but I have some feedback on the filter process. In the query total & query filter blocks, the filter runs over a placeholder value so the filter only applies to that item, I think an approach like that would be better here.

mu-plugins/blocks/ratings-bars/render.php
StevenDufresne: > Is filtering all of the attributes is necessary here?

I would prefer to only filter the block attributes but I don't see a way to do that without forcing consumers to add multiple filters for each parameter. Is there a different approach I should be considering?

StevenDufresne: > Do you think we'll ever want to have a static  block?

I'm not sure. If the answer is no, how would you architect this differently?

ryelle: It seemed easier to show than describe, so my approach is here: https://github.com/WordPress/wporg-mu-plugins/compare/try/rating_blocks...try/rating_blocks-alt

- The filter runs over an empty array, which is populated by the hooked function in the child themes, so it doesn't need to run over the classnames, etc.
- I used `wp_parse_args` to ensure the resulting array had the right formatting, so then we can just check a few things to make sure the filter did provide data.
- It is required to provide the data via filter, if no hooked function returns the data, the block will not render.
- Removed the `slug` arg, and built the support URL in the hooked function, ex `wordpress.org/support/theme/SLUG/reviews/`. This + `add_query_arg` create the URL in the render function. It would be nice to replace this with the block binding, but that's not available to custom blocks yet.
    - I thought about leaving this as an attribute, but think it makes sense to build it dynamically like this.
- Use one filter for both blocks. I don't feel _very_ strongly about this, but they use a similar format, so it seems easier in the child theme to only provide the ratings data once.

Now, the child theme hook would look like this:

```php
use function WordPressdotorg\Theme\Theme_Directory_2024\{get_support_url, get_theme_information};

/**
 * Update ratings blocks with real rating data.
 *
 * @param array $data    Rating data.
 * @param int   $post_id Current post.
 *
 * @return array
 */
function set_rating_data( $data, $post_id ) {
	$theme = get_theme_information( $post_id );
	if ( ! $theme ) {
		return $data;
	}

	return array(
		'rating' => $theme->rating,
		'ratingsCount' => $theme->num_ratings,
		'ratings' => $theme->ratings,
		'supportUrl' => get_support_url( $theme->slug . '/reviews/' ),
	);
}
add_filter( 'wporg_ratings_data', __NAMESPACE__ . '\set_rating_data', 10, 2 );
```

Feel free to use any of that branch's code that you want, but that's my "how I'd do it" answer 🙂

StevenDufresne: I see. Yeah that seems logical to me. I don't have much of a preference here. If you think it's better, we can go with that.

ryelle (approved):
Looks good now, one last little 🧹 cleanup request before merging — the attributes can be removed from `block.json`.

PR #637, opened 2024-07-07
Site Breadcrumbs: Return empty string when breadcrumbs is an empty array

outdoor2kode:
In the design of most wporg sites, breadcrumbs are not needed when the level is less than 3. Therefore, a filter is usually applied to the breadcrumbs block to make the breadcrumbs an empty array. If this early return is not implemented, it would still return a div that occupies space.

See https://github.com/WordPress/Learn/pull/2598/commits/6fa130a1dd222cf020da700f0afb9b65cef33643


adamwoodnz (approved):
I haven't tested, but this makes sense to me 👍

PR #638, opened 2024-07-10
Gutenberg tweaks: Remove suffix from RTL styles

ryelle:
CSS registered from block.json are incorrectly configured with a suffix for RTL sites, but the files don't use a suffix. This prevents the files from being replaced correctly, causing visual issues on RTL sites.

The "simple" fix here is to just wipe out the suffix on all wporg-added styles. Checking for the `rtl` extra data should also prevent replacing anything that isn't using the block.json registration. We don't use `.min` or separate minified files anywhere (IIRC), so this shouldn't have any negative impacts.

See https://core.trac.wordpress.org/ticket/61625.

| | Before | After |
|---|---|---|
| Query filters | Screenshot 2024-07-10 at 5 03 45 PM | Screenshot 2024-07-10 at 5 02 27 PM |
| Local nav | Screen Shot 2024-07-10 at 17 04 08 | Screen Shot 2024-07-10 at 17 04 43 |

**To test**

This is easiest to test on a sandbox, so that you can view a real RTL site. When applied, this should fix a lot of the issues flagged on https://github.com/WordPress/Learn/issues/2483.


adamwoodnz (approved):
Tested in sandbox and query filters and local nav LGTM, great sleuthing.

PR #641, opened 2024-07-18
Local navigation block: Hide page title before collapsing menu

### LINKED ISSUE #640: Local Navigation: Prevent menu from collapsing when page title is long
See https://github.com/WordPress/Learn/issues/2622 — When the site + page title is long, the menu collapses to prevent the items from colliding or wrapping. This is not ideal, the navigation should stay visible as the primary focus of this bar.

> Another option could be to simply _drop_ the page title when it would collide, so that only "Learn WordPress" is there:
>
> https://github.com/WordPress/Learn/assets/541093/f28937eb-6cf2-4829-af57-0271f994e12e

ryelle:
Fixes #640, see See https://github.com/WordPress/Learn/issues/2622. When elements (site title, page title, navigation) are too long, they can overlap or cause wrapping issues. In https://github.com/WordPress/wporg-mu-plugins/pull/573, the navigation menu collapsed when they would wrap, but in some cases this can cause the navigation to be hidden on more screen-sizes than is ideal. Instead, the navigation menu should be prioritized, and the page title can be hidden first.

This PR updates the logic, so that the page title is hidden first, extending the screen sizes that will see the correct desktop version of the nav bar. If the menu will still collide with the site title, it does collapse down into the chevron dropdown, but this is only at some screen sizes, since most site titles are short.

**Screenshots**

Screen recording of shrinking the screen from 1440px, with an exaggerated page title and menu length. The page title disappears around 1300px, and while the menu stays inline until around 970px (where it would otherwise run into the site title). Previously, the menu would collapse at that 1300 width. (Note that these breakpoints are relative to the page content, not universal)

https://github.com/user-attachments/assets/e5c47d7c-b332-45fd-842b-f415e2ce444f

| | Before | After |
|---|---|---|
| 1310px | Screen Shot 2024-07-18 at 17 14 13 | same |
| 1300px | Screen Shot 2024-07-18 at 17 19 44 | Screen Shot 2024-07-18 at 17 14 17 |

On a real page, the Learn menu is now visible here, where previously it collapsed on a larger screen size.

Screen Shot 2024-07-18 at 17 42 50

**To test:**

- View a site with a "L2B" page header — has the site title and a page title.
- Scroll down so you can see the page title (easier to test)
- Shrink your screen down, the page title should disappear before the menu collapses into a chevron
- Nothing should wrap to a second line
- The mobile menu kicks in at 600px, and there should be no change there

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -185,7 +185,7 @@
 		}
 	}
 
-	@media ( max-width: 550px ) {
+	@media ( max-width: 599px ) {
 		& .wp-block-group p:not(.wp-block-site-title) {
 			display: none;
 		}
--- a/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
+++ b/mu-plugins/blocks/local-navigation-bar/postcss/style.pcss
@@ -341,6 +341,12 @@
 		display: none;
 	}
 
+	&.wporg-hide-page-title {
+		& .wp-block-group p:not(.wp-block-site-title) {
+			display: none;
+		}
+	}
+
 	&.wporg-show-collapsed-nav {
 		& .wp-block-navigation {
 			display: none;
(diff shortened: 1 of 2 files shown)


fcoveram (approved):
From the screenshots, it looks great to me 🚀

adamwoodnz (approved):
LGTM

I did find one odd point between 550px and 599px where the site title is covered by the mobile button, but this appears to be existing, and I guess only happens with this really long title. It is fairly graceful.

Screenshot 2024-07-22 at 9 30 43 PM

ryelle:
> I did find one odd point between 550px and 599px where the site title is covered by the mobile button, but this appears to be existing, and I guess only happens with this really long title. It is fairly graceful.

Yeah, I avoided changing anything under 600px, but it is a little strange that the title appears for just that 49px width breakpoint. I've bumped it up to 599 so the page title stays hidden.

PR #649, opened 2024-09-03
Global header, local nav: Make the logo bigger

### LINKED ISSUE #646: Main navigation bar: Make the logo bigger
The W logo in the main WordPress.org navigation bar serves both as a branding element for the project, and an anchor for navigating to the homepage. At 27x27px it's a tad small. Here's a proposal to scale that up to 30x30px to give it a little more presence.

before after

Here's an overlay showing before / after (in blue):

W symbol

Overall it's not that much larger, but it's enough to make it feel slightly more substantial. The mockups in WordPress/wporg-mu-plugins#647 also include this.

ryelle:
Fixes #646. Adjust the spacing around the logo and other elements to use a calculated value from the nav bar height, so the extra media queries can be removed. This also adjusts the size of the logo from ~27px to 30px, [making the logo bigger](https://images.app.goo.gl/Hz8tyJR5omQZPQCL7).

**Screenshots**

| Before | After |
|---|---|
| logo-before-top | logo-after-top |
| logo-before-scroll | logo-after-scroll |
| logo-before-mobile | logo-after-mobile |
| logo-before-download | logo-after-download |

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/header/header.pcss
@@ -31,11 +31,6 @@ html {
 		padding-left: var(--wp--style--block-gap);
 		padding-right: var(--wp--style--block-gap);
 		margin: 0;
-
-		@media (--tablet) {
-			padding-top: 33px;
-			padding-bottom: 33px;
-		}
 	}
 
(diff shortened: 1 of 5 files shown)


jasmussen:
Nice, ship it!

PR #650, opened 2024-09-10
New block: Chapter List

ryelle:
This ports [the "Chapter list" block from Developer](https://github.com/WordPress/wporg-developer/tree/trunk/source/wp-content/themes/wporg-developer-2023/src/chapter-list) to the mu-plugins so it can be used on other handbook sites (Five for the Future, in this case).

The only change to the code was setting up some default CSS, so the styles in the child theme.json can be removed.

See https://github.com/WordPress/wporg-developer/pull/539

**Screenshots**

There should be no visual changes.

Screen Shot 2024-09-10 at 17 07 02

Screen Shot 2024-09-10 at 17 14 08

**To test**

1. Apply this PR, build files
2. Apply the wporg-developer PR https://github.com/WordPress/wporg-developer/pull/539
3. OR, test with Five for the Future, if you have handbook content set up locally
4. View a handbook page
5. The "Chapter list" left sidebar should appear
6. It should work like production— a list of page links, with the current item bold
7. It should scroll with the page content, unless it's longer than the footer in which case it will scroll internally

### DIFF
--- a/mu-plugins/blocks/chapter-list/class-chapter-walker.php
+++ b/mu-plugins/blocks/chapter-list/class-chapter-walker.php
@@ -1,0 +1,95 @@
+<?php
+/**
+ * Custom walker for chapter block page list.
+ *
+ * Identical to `Walker_Page` except with a `walk()` that ignores orphaned
+ * pages, which are pages with an ancestor that is not published.
+ */
+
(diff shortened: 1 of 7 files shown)


adamwoodnz (approved):
Tested on local developer site and LGTM.

Haven't tested 5ftf.

PR #651, opened 2024-09-10
Add a shortcode for nocookie youtube embeds

ryelle:
See https://github.com/WordPress/wporg-main-2022/issues/493#issuecomment-2316230298 — We'd like to use the "privacy-enhanced" youtube embeds, but the domain used is not supported by the core embeds. I've put this code in wporg-mu-plugins because I'll also want to use it on WCUS's livestream pages; and I can see needing it on other pages.

Hopefully [the core ticket](https://core.trac.wordpress.org/ticket/44610) will be fixed at some point, and we can phase this out.

```
[youtube-nocookie title="Showcase Day video player"]https://www.youtube-nocookie.com/embed/PM2Ov8atqBw?si=4YwCx25TQfdWPcW_[/youtube-nocookie]
```

Screenshot 2024-09-10 at 6 34 43 PM

**To test**

- Find a youtube video
- Click share, embed to get to the iframe code
- Select "Enable privacy-enhanced mode."

Screenshot 2024-09-10 at 6 32 06 PM

- From there, copy out the `src` URL
- Create or edit a post, add the shortcode block
- Add the shortcode: `[youtube-nocookie]COPIED URL[/youtube-nocookie]`
- Optionally set height, width, and title attributes

Try other URLs, it should only create the iframe for youtube-nocookie.com domains.

### DIFF
--- a/mu-plugins/plugin-tweaks/youtube-shortcode.php
+++ b/mu-plugins/plugin-tweaks/youtube-shortcode.php
@@ -19,11 +19,10 @@ add_shortcode( 'youtube-nocookie', __NAMESPACE__ . '\render' );
  *
  * @param array  $attr    Shortcode attributes array, can be empty if the original arguments string cannot be parsed.
  * @param string $content Content inside shortcode tags.
- * @param string $tag     Shortcode name.
  *
  * @return string HTML code for iframe embed.
  */
-function render( $attr, $content, $tag ) {
+function render( $attr, $content ) {
 	// Short out early if the content is not a valid URL.
 	// Returns null if content is not a URL at all.
 	$host = wp_parse_url( $content, PHP_URL_HOST );
--- a/mu-plugins/plugin-tweaks/youtube-shortcode.php
+++ b/mu-plugins/plugin-tweaks/youtube-shortcode.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Add a shortcode for nocookie youtube embeds.
+ *
+ * These are not supported in the core embed block or shortcode.
+ * See https://core.trac.wordpress.org/ticket/44610
+ */
+
+namespace WordPressdotorg\MU_Plugins\Plugin_Tweaks\Youtube_Shortcode;
+
+add_shortcode( 'youtube-nocookie', __NAMESPACE__ . '\render' );
+
+/**
+ * Render the youtube iframe.
+ *
+ * The shortcode content is the URL, checked against a safelist of
+ * `youtube-nocookie` domains. Attributes can be `width`, `height`,
+ * and `title.
+ *
+ * @param array  $attr    Shortcode attributes array, can be empty if the original arguments string cannot be parsed.
+ * @param string $content Content inside shortcode tags.
+ * @param string $tag     Shortcode name.
+ *
+ * @return string HTML code for iframe embed.
+ */
+function render( $attr, $content, $tag ) {
+	// Short out early if the content is not a valid URL.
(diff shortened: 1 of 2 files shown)


mu-plugins/plugin-tweaks/youtube-shortcode.php
```diff
@@ -0,0 +1,55 @@
...
+	$html_attrs = '';
+	foreach ( $args as $name => $value ) {
+		if ( $value ) {
+			$html_attrs .= $name . '="' . esc_attr( $value ) . '" ';
+		}
+	}
+
+	return sprintf(
+		// `allow` settings copied from youtube-provided embed code.
+		'<iframe style="aspect-ratio: 16/9;" src="%s" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen %s></iframe>',
```
StevenDufresne, line 50: Minor. Should this use [numbered placeholders](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#best-practices-for-writing-strings:~:text=If%20you%20have%20more,not%20what%20we%20want.)?

mu-plugins/plugin-tweaks/youtube-shortcode.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Render the youtube iframe.
+ *
+ * The shortcode content is the URL, checked against a safelist of
+ * `youtube-nocookie` domains. Attributes can be `width`, `height`,
+ * and `title.
+ *
+ * @param array  $attr    Shortcode attributes array, can be empty if the original arguments string cannot be parsed.
+ * @param string $content Content inside shortcode tags.
+ * @param string $tag     Shortcode name.
```
StevenDufresne: [Core uses the parameter name](https://developer.wordpress.org/reference/functions/shortcode_atts/#:~:text=in%20shortcode%20tag.-,%24shortcode,The%20name%20of%20the%20shortcode%2C%20provided%20for%20context%20to%20enable%20filtering,-Default%3A) `$shortcode` for this. Should we also follow that standard?

StevenDufresne: I also don't see that parameter being used.

mu-plugins/plugin-tweaks/youtube-shortcode.php
StevenDufresne, line 50: Would enforcing `16/9` but allowing users to set width/height be an issue for editors?

mu-plugins/plugin-tweaks/youtube-shortcode.php
ryelle: Core uses `$tag` elsewhere, so I don't think it's really a standard. This parameter is only useful if we're using the same callback for different shortcodes, which doesn't apply here. I'll remove it.

mu-plugins/plugin-tweaks/youtube-shortcode.php
ryelle: Good question— I assumed you would only ever really set one of them, but you can set both and it will override the aspect ratio.

For example, `[youtube-nocookie height=300 width=300…` is square:

Screenshot 2024-09-11 at 11 27 22 AM

ryelle:
Going to merge this so I can build out the WordCamp pages, but we can iterate if needed.

PR #655, opened 2024-10-03
Link Wrapper: Default to the permalink if the url isn't defined

adamwoodnz:
This enables the block to be used in query loops, for example the Make homepage team grid.

Uses server side rendering so that the permalink can be easily looked up.

### DIFF
--- a/mu-plugins/blocks/link-wrapper/src/block.json
+++ b/mu-plugins/blocks/link-wrapper/src/block.json
@@ -30,6 +30,9 @@
 				"padding": true
 			}
 		},
+		"dimensions": {
+			"minHeight": true
+		},
 		"typography": {
 			"fontSize": true,
 			"lineHeight": true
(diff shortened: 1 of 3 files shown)


mu-plugins/blocks/link-wrapper/src/block.json
```diff
@@ -30,6 +30,9 @@
 				"padding": true
 			}
 		},
+		"dimensions": {
+			"minHeight": true
```
adamwoodnz, line 35: Allow configuring the wrapper to fill the grid item.

ryelle:
I think changing the `save` function like this will cause invalid block errors in the editor. But the render function would still work, `$content` should be the link markup. Maybe that + [WP_HTML_Tag_Processor](https://developer.wordpress.org/reference/classes/wp_html_tag_processor/) to update the href attribute if no `url`?

I've [done similar](https://github.com/WordPress/wporg-theme-directory/blob/f52089c06b6cdb60fc62e6f7d7a26fceca754025/source/wp-content/themes/wporg-themes-2024/inc/block-config.php#L289-L298) with just `str_replace` on the theme directory, so this would definitely be useful 🙂

PR #656, opened 2024-10-04
Output message when no ratings exist.

StevenDufresne:
See: https://github.com/WordPress/wordpress.org/pull/380

This block doesn't have any view when no ratings exist.

| Before | After |
|--------|--------|
| Screenshot 2024-10-07 at 10 38 53 AM | Screenshot 2024-10-07 at 10 38 43 AM |

### DIFF
--- a/mu-plugins/blocks/ratings-bars/render.php
+++ b/mu-plugins/blocks/ratings-bars/render.php
@@ -34,7 +34,7 @@ $defaults = array(
 $data = wp_parse_args( $data, $defaults );
 
 if ( empty( $data['ratings'] ) || empty( $data['ratingsCount'] ) ) {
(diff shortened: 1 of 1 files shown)


StevenDufresne:
Looking at our language, we also appear to use `Ratings` and `Reviews` interchangeably. I wonder if we want to update that.

ryelle:
Currently plugins with no reviews say "This plugin has not been rated yet." ([example](https://wordpress.org/plugins/magical-blocks/)). Could be generalized to "This has not been rated yet." ?

That said, I think we generally use "reviews" for the whole "unit of feedback", text and number, while ratings are just the numbers. If not, I think we could standardize on that. So this string could be "No reviews have been submitted yet", with the "Add my review" link under it.

PR #657, opened 2024-10-10
Global header: Implement design updates

### LINKED ISSUE #572: Consistency tweaks in Global nav
As proposed in #535, the design introduced involves changes in both local and global nav. For the latter, the changes are general spacings to reach consistency across the two components, and on mobile, the menu has new text styles that make group indentation clearer.

Here is an image of global nav menu in both desktop and mobile views.

Menu popover of Global nav component for each site group

Here is an image of the main components of Global Nav.

Figma components with their variants of Global nav

---

All changes are [documented](https://www.figma.com/file/VO03eHjoYJrX1aLisReB8q/Simpler-navigation?type=design&node-id=1069-6032&mode=dev) in the Design page of ["Simpler navigation" file](https://www.figma.com/file/VO03eHjoYJrX1aLisReB8q/Simpler-navigation?node-id=1%3A2193&mode=dev).

ryelle:
This applies the style changes suggested in #572 for the global header. The header height is decreased (and any alignment issues fixed), the active style is more subtle, and the mobile menu is more consistent.

The menu items with child menus (Extend, Learn, etc) are now a lighter grey, but they are still focusable & clickable — this is core behavior for "click to open" submenus in the overlay menu. Nothing happens if you click them. This could be an issue for Gutenberg, but there's nothing we can do here (I tried a few things).

I simplified & fixed the modal view javascript now that the navigation block uses the Interactivity API instead of the micromodal library, so opening the search overlay will close an open menu overlay and vice versa.

The new design removed the side border on active items on mobile (see the 4th row of screenshots), which means the only indicator of current page is color. We need to add some other non-color effect here. @WordPress/meta-design 

Fixes #572

**Screenshots**

The desktop screens are all neutral, no interactions. The mobile views were opened by keyboard, so the first element is focused. For News, it's also the current page.

| Before | After |
|---|---|
| before news desktop | after news desktop
| before showcase desktop | after showcase desktop |
| before about desktop| after about desktop |
| before news mobile | after news mobile |
| before about mobile | after about mobile |

**Video for focus & dropdown styles**

https://github.com/user-attachments/assets/0fc2da01-a94a-4175-9d23-2fcbf17dc8fc

https://github.com/user-attachments/assets/97f15185-2b15-4dda-848d-f20450c4f0b1

**To test**

- Check out & build this branch
- Use this with any of the other environments
- See the header is shorter now
- Make sure nothing looks broken

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/view.js
+++ b/mu-plugins/blocks/global-header-footer/js/view.js
@@ -193,29 +193,16 @@
 			closeSearchButton.setAttribute( 'aria-label', labels.closeSearchLabel || 'Close Search' );
 		}
 
-		// Watch for the `has-modal-open` class to be removed, and remove the global class too.
-		// This works as a callback to be fired when the global header modals are closed, as
-		// they're attached when each modal opens.
-		const modalCloseObserver = new window.MutationObserver( ( mutationList, observer ) => {
-			for ( const mutation of mutationList ) {
-				const { attributeName, type, target } = mutation;
-				if ( type === 'attributes' && attributeName === 'class' ) {
-					if ( ! target.classList.contains( 'has-modal-open' ) ) {
-						target.classList.remove( 'has-global-modal-open' );
-					}
-				}
-			}
-			// Remove the observer to prevent recursion. This will be re-attached when the modal is opened.
-			observer.disconnect();
-		} );
-
-		const openButtons = document.querySelectorAll( '[data-micromodal-trigger]' );
+		const openButtons = document.querySelectorAll(
+			'.global-header__navigation [data-wp-on-async--click="actions.openMenuOnClick"],' +
+			'.global-header__search [data-wp-on-async--click="actions.openMenuOnClick"]'
+		);
 		openButtons.forEach( function ( button ) {
 			// When any open menu button is clicked, find any existing close buttons and click them.
 			button.addEventListener( 'click', function ( event ) {
-				const thisModal = event.target.getAttribute( 'data-micromodal-trigger' );
+				const thisModal = event.target.parentNode.querySelector( '[data-wp-class--has-modal-open]' ).id;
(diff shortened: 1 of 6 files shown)


mu-plugins/blocks/global-header-footer/js/view.js
```diff
@@ -193,44 +193,36 @@
...
-					}
-				}
-			}
-			// Remove the observer to prevent recursion. This will be re-attached when the modal is opened.
-			observer.disconnect();
-		} );
-
-		const openButtons = document.querySelectorAll( '[data-micromodal-trigger]' );
+		const openButtons = document.querySelectorAll(
+			'.global-header__navigation [data-wp-on-async--click="actions.openMenuOnClick"],' +
```
StevenDufresne, line 197: Do we need to rely on the Interactivity API attributes? Is it fair to assume a button in the menu is a dropdown?

ryelle: This is a way to differentiate open and close buttons, if we went for just "button" it would catch both.

jasmussen:
Looks great! If @fcoveram has a chance to look, it would be nice, but no blocker.

fcoveram:
Sorry for arriving late here. I was thinking of bringing the border line back. Here is a mockup of the border idea with visual tweaks.

Mockups of menu items in large and small breakpoints

If you all think the above looks good, I can create the issue and point out the new documentation in the Design Library.

ryelle:
> If you all think the above looks good, I can create the issue and point out the new documentation in the Design Library.

Works for me. When you create the issue, can you write out exactly what "with visual tweaks" means— just adding the border, or did other spacing/sizes change, that kind of thing. Thanks!

fcoveram:
I created the ticket #666 (👹) asking for implementing the idea dropped in my previous comment

PR #658, opened 2024-10-20
Render Callout blocks as Notice blocks

### LINKED ISSUE #374: Make blogs: Convert callout blocks to notice blocks
The Make blogs have old style callout blocks:

These need to display as notice blocks. We added a [mu-plugin to convert callout shortcodes to notices](https://github.com/WordPress/wporg-mu-plugins/pull/337), perhaps we can use a similar approach, or failing that, add these [fallback styles](https://github.com/WordPress/wporg-developer/pull/429) to the parent theme.

This should also address the [styles of the callout blocks on Learn](https://github.com/WordPress/Learn/issues/2739#issuecomment-2272886063).

adamwoodnz:
Builds on previous work to render Callout shortcodes as Notice blocks.

This enhancement saves admins of sites like the Make blogs and Learn WordPress from having to update all the existing Callouts in content to match the redesigned UI.

Closes https://github.com/WordPress/wordpress.org/issues/374

Note that we added [style overrides in Developer Resources](https://github.com/WordPress/wporg-developer/blob/034722c393dc768ad00b78887b04d781d87d6f67/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L579), and these will still be required as the callouts have been parsed from GitHub markdown and exist as post content, not blocks.

### Screenshots from [Make/Support](https://make.wordpress.org/support/?new-theme=1)

| Before | After |
|-|-|
| Screenshot 2024-10-21 at 10 56 25 AM | Screenshot 2024-10-21 at 10 57 18 AM |

| Admin | Frontend |
|-|-|
| Screenshot 2024-10-21 at 11 25 22 AM | Screenshot 2024-10-21 at 11 25 32 AM |

### Testing

1. Load this branch in a local environment or sandbox
2. Add Callout blocks to a page or post. In the admin the block should display as a Callout.
3. On the frontend the block should render as a Notice, with all inner blocks rendered as expected.

### DIFF
--- a/mu-plugins/blocks/notice/index.php
+++ b/mu-plugins/blocks/notice/index.php
@@ -9,6 +9,8 @@ namespace WordPressdotorg\MU_Plugins\Notice;
 // Run after `WPorg_Handbook_Callout_Boxes` registers the shortcodes.
 add_action( 'init', __NAMESPACE__ . '\init', 11 );
 
+add_filter( 'pre_render_block', __NAMESPACE__ . '\render_callout_block_as_notice', 11, 2 );
+
 /**
  * Registers the block using the metadata loaded from the `block.json` file.
  * Behind the scenes, it registers also all assets so they can be enqueued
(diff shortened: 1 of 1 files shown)


ryelle:
Good approach 👍🏻 

Very minor, but with the old theme there's no spacing between notices. This isn't an issue with the new theme, so this can probably just be considered in-progress with the switch to the new theme.

Screenshot 2024-10-21 at 5 11 12 PM

I also noticed extra space at the end of the notices, it looks like o2 is injecting a script tag, so the p:last-child rule which would usually prevent the margin is not applying:

Screenshot 2024-10-21 at 5 47 37 PM

adamwoodnz:
> o2 is injecting a script tag, so the p:last-child rule which would usually prevent the margin is not applying

Nice catch, fixed in [141e450](https://github.com/WordPress/wporg-mu-plugins/pull/658/commits/141e450fa83cbc7a370d497f3bf0913d25f98899)

PR #670, opened 2024-11-19
Show a more accurate star rating.

StevenDufresne:
This PR removes rounding for display averages as discussed in https://github.com/WordPress/wporg-theme-directory/issues/147#issuecomment-2194747159 and https://github.com/WordPress/wordpress.org/pull/408#issuecomment-2448457010. This is particularly important for plugins as most plugins fall within the 4-5 star ranges and rounding makes it difficult to compare plugins by ratings.

## Themes
| Before | After |
|--------|--------|
| Screenshot 2024-11-19 at 12 36 50 PM | Screenshot 2024-11-19 at 12 37 06 PM | 
| Screenshot 2024-11-19 at 12 36 39 PM | Screenshot 2024-11-19 at 12 36 20 PM | 

## Plugins
| Before | After |
|--------|--------|
| Screenshot 2024-11-19 at 12 37 44 PM | Screenshot 2024-11-19 at 12 37 19 PM |
| Screenshot 2024-11-19 at 12 37 37 PM | Screenshot 2024-11-19 at 12 37 26 PM |


mu-plugins/blocks/ratings-stars/render.php
```diff
@@ -30,11 +30,11 @@
 	<div class="wporg-ratings-stars__icons">
 		<?php
 
-		$display_rating = round( $data['rating'] / 10 ) * 0.5;
+		$display_rating = $data['rating'] / 10 * 0.5;
 		for ( $i = 0; $i < 5; $i++ ) {
 			if ( $i + 1 <= $display_rating ) {
 				echo '<svg class="is-star-filled" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><path d="M11.776 4.454a.25.25 0 01.448 0l2.069 4.192a.25.25 0 00.188.137l4.626.672a.25.25 0 01.139.426l-3.348 3.263a.25.25 0 00-.072.222l.79 4.607a.25.25 0 01-.362.263l-4.138-2.175a.25.25 0 00-.232 0l-4.138 2.175a.25.25 0 01-.363-.263l.79-4.607a.25.25 0 00-.071-.222L4.754 9.881a.25.25 0 01.139-.426l4.626-.672a.25.25 0 00.188-.137l2.069-4.192z"></path></svg>';
-			} else if ( $i + 0.5 === $display_rating ) {
+			} else if ( $i + 0.5 <= $display_rating && $i + 1 > $display_rating ) {
```
outdoor2kode: Do we need to check `$i + 1 > $display_rating` here? 
The first if condition checks `if ($i + 1) <= $display_rating`. If this condition fails, it implies that `$i + 1 > $display_rating` is already true.

mu-plugins/blocks/ratings-stars/render.php
```diff
@@ -30,11 +30,11 @@
 	<div class="wporg-ratings-stars__icons">
 		<?php
 
-		$display_rating = round( $data['rating'] / 10 ) * 0.5;
+		$display_rating = $data['rating'] / 10 * 0.5;
```
outdoor2kode: I think we can directly divide by 20, as it’s just converting a 0–100 range into 0–5.

StevenDufresne: Yes!

mu-plugins/blocks/ratings-stars/render.php
StevenDufresne: Yes! (AI is nice, but I'm falling out of love with it)

ryelle:
This does change the rounding of values so that a 3.8 used to render as 4 stars, but now renders as 3½, which also means that only plugins with a 5 will get the 5 stars. 

Twenty Twenty-Four with a 3.8 rating:

| Before | After |
|---|---|
| Screenshot 2024-11-21 at 4 42 13 PM | Screenshot 2024-11-21 at 4 42 16 PM |]

The card view also currently uses a rounded view (it still uses `Template::get_star_rating`), so as this is, the stars render differently:

| Card view | Single view |
|---|---|
| Screenshot 2024-11-21 at 4 46 13 PM | Screenshot 2024-11-21 at 4 46 25 PM |

ryelle:
The change in https://github.com/WordPress/wporg-mu-plugins/pull/670/commits/43bb8a55b4735ddb800b5f2384f596d1bf31b0bd brings back the previous rounding code for the stars themselves, but calculates a separate unrounded display rating:

Screenshot 2024-11-21 at 4 56 43 PM

StevenDufresne:
Makes sense. Thanks for the help kelly.

PR #682, opened 2025-04-14
Global Header/Navigation: Switch to observer for menu open & close actions

### LINKED ISSUE #678: Bug in Mobile Layout – Navigation Menu Behavior*
When scrolling the page so that only part of the header is visible and then tapping the main navigation menu, the menu appears below the admin toolbar instead of aligned with the header.

If the user then taps an empty area on the screen, the main header begins to behave like a sticky header on mobile.

At this point, if the user opens the toggle menu from the now-"sticky" header, the menu appears detached from the header, resulting in a broken layout.

### **Issue 1: Sticky Header Behavior After Toolbar Click**

**Steps to reproduce:**
1. On the mobile layout, click the main menu toggle button.
2. Then tap the profile toolbar.

**Bug behavior:**
- The main header begins to behave like a *sticky header*, remaining fixed at the top of the viewport.
- The subheader shifts upward to occupy the original space of the main header.

> 📝 *This behavior does not appear to be intended and may be triggered by layout reflows or class toggling after the toolbar interaction.*

---

### **Issue 2: Menu Detaches from Subheader on Scroll**

**Steps to reproduce:**
1. Scroll down the page in mobile view.
2. Click the subheader’s menu toggle.

**Bug behavior:**
- The expanded subheader menu appears *visually detached* from the subheader.

---

### **Issue 3: Combined Layout Shift and Detached Menu**

**Steps to reproduce:**
1. Trigger the bug from **Issue 1** (main header becomes sticky, subheader shifts).
2. Expand the **main header’s menu** using its toggle.
3. Expand the **subheader’s menu** using its toggle.
4. Collapse the **main header’s menu**.

**Bug behavior:**
- The subheader menu remains open while the main header menu is collapsed.
- At this point, the subheader menu scrolls independently from its parent header, appearing detached from the layout.
- The header-menu structure becomes fragmented and causes user confusion.

> 📝 *This compound issue likely arises due to improper DOM hierarchy or inconsistent positioning logic when both headers and menus are simultaneously active.*

Image

Image

### LINKED ISSUE #681: Global submenu incorrectly positioned with scroll.
Originally filed as https://meta.trac.wordpress.org/ticket/7940 by @suzuki3932

> When I scroll the page and click the arrow under the two lines, the menu position is bugged. (only smartphone)
>
> Browser: Chrome
> Menu: Submenu
> incorrectly: I think it's designed not to scroll

See below video, duplicatable with Chrome Devtools.
To trigger, scroll page such that the submenu expander is the top visible element on the page.

https://github.com/user-attachments/assets/272a0771-314c-44db-96fa-182b4495c51a

ryelle:
Fixes #678. The global header nav adds a class when opened, `has-global-modal-open`, but this was only being removed when the nav was closed by button click. This changes the way _that_ class is removed by watching the `has-modal-open` class, which the core block adds, so that the custom class is always removed when the nav modal closes.

I then used the same observer to fix #681, by simply scrolling to the top of the page when a nav is opened. I did try to fix the alignment via CSS, but it's fixed to the viewport position. Switching to `absolute` + some other tweaks worked, but then the bottom was cut off.

Screen Shot 2025-04-14 at 18 42 58

So I'm on the fence about whether this is too hacky — it does make the screen jump, but then everything is positioned perfectly… Here's an example:

https://github.com/user-attachments/assets/c98b6494-4262-4fcd-b7c2-9091cdfbaa3d

**To test**

Global header fix:
1. Use a viewport < 890px
2. Open the global header menu
3. Hit escape
4. It should close the menu, and behave like before the menu was open
    - The local navigation bar should still be visible
    - The global header should not be sticky

Detached menu fix:
1. Use a viewport < 600px
2. Scroll down slightly
3. Open the global header or local navigation dropdown
4. It should scroll back to the top of the page
5. The open menu's close button should be correctly aligned with the previous open button
6. None of the page itself should be visible

This issue ^ appears on both the global header menus and local nav menus, so you can test both.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/js/view.js
+++ b/mu-plugins/blocks/global-header-footer/js/view.js
@@ -1,3 +1,4 @@
+/* global MutationObserver */
 /**
  * File wporg-global-header-script.js.
  *
@@ -195,7 +196,7 @@
 
 		const openButtons = document.querySelectorAll(
 			'.global-header__navigation [data-wp-on-async--click="actions.openMenuOnClick"],' +
-			'.global-header__search [data-wp-on-async--click="actions.openMenuOnClick"]'
+				'.global-header__search [data-wp-on-async--click="actions.openMenuOnClick"]'
 		);
 		openButtons.forEach( function ( button ) {
 			// When any open menu button is clicked, find any existing close buttons and click them.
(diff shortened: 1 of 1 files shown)


Jiwoon-Kim:
I was thinking, would it be possible to implement a scroll-responsive sticky header for this feature? the header could hide when the user scrolls down and reappear when scrolling up. This approach could help improve the user experience by minimizing the header's space when not needed, while still providing easy access when the user needs it.

Also, how about implementing the menu in an off-canvas style? This could provide a clean and modern design, saving space while offering easy access to the navigation items. The menu could slide in from the side when triggered (via a button or swipe), and close when clicked outside or when an item is selected.

Would love to hear thoughts on this approach—let me know if it's feasible and if you think it could be added!

ryelle:
We've gone through a few iterations on the sticky menu idea to land where we are today. You can see some of the historical discussion here https://github.com/WordPress/wporg-mu-plugins/issues/465#issuecomment-1743859797.

> Also, how about implementing the menu in an off-canvas style? 

The header and menu we have now is done mostly with core blocks (with some adjustments for design), so unless this can be done with core blocks, I don't think it's likely to change. Smaller fixes for the current header would be more likely to be implemented. However, if you'd like to propose something, you could [start a discussion](https://github.com/WordPress/wporg-mu-plugins/discussions).

dd32:
> So I'm on the fence about whether this is too hacky — it does make the screen jump, but then everything is positioned perfectly… Here's an example:

This does feels too hacky.. and the screen jump isn't amazing IMHO.. however, the positioning and visual is more important than the minor screen jump and the code seems like it's _relatively_ simple, and I'm not really familiar with MutationObserver..

Just noting that I've seen this, am going to double-check it by testing, and will merge+deploy.

PR #684, opened 2025-05-05
Fonts: Add Noto Serif SC, KR, and subsetting tool for CJK 

ryelle:
See https://github.com/WordPress/wporg-parent-2021/issues/149 — This PR adds Noto Serif SC (Simplified Chinese), following the other additions of Noto for Japanese and Central Kurdish. I've also added a filter when calling to preload fonts, this will avoid excess locale conditionals in the blocks file.

On its own, this PR should not have any noticeable effect, except that Noto Kufi will not be preloaded. See https://github.com/WordPress/wporg-parent-2021/pull/173 for actually loading Noto Serif SC, and test instructions.


mu-plugins/global-fonts/index.php
```diff
@@ -257,9 +261,6 @@ function get_font_url( $font, $subset ) {
 
 	$filepath = $font_folder . $font_file_name . $lower_subset . '.woff2';
 	if ( ! file_exists( __DIR__ . '/' . $filepath ) ) {
-		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
-			trigger_error( sprintf( 'Requested font file %s does not exist.', esc_html( $filepath ) ), E_USER_WARNING );
-		}
```
ryelle, line 262: This is not a helpful error message. Ideally it would be flagged to the translators, because it seems to trigger most often when they've translated the subset into their language, but instead we can silently ignore it. The only effect of ignoring it is that the font is not _pre_loaded, but it is still loaded.

mu-plugins/global-fonts/NotoSerif/NotoSerifSC-cjk.woff2
devhaozi, line 1: This font file is large. Can use a tool like https://chinese-font.netlify.app/en/online-split/ to slice it to increase loading speed if possible.

ryelle: Thanks for pointing out that package — I've tried it out a little and it creates a lot of smaller font files, which would not be great for managing the fonts. I'll do some more testing and see if we can find a balance.

devhaozi: > Thanks for pointing out that package — I've tried it out a little and it creates a lot of smaller font files, which would not be great for managing the fonts. I'll do some more testing and see if we can find a balance.

Documentation is at https://github.com/KonghaYao/cn-font-split/tree/release/packages/ffi-js#%E5%AE%8C%E6%95%B4%E7%9A%84%E5%8F%AF%E4%BC%A0%E5%85%A5%E5%8F%82%E6%95%B0
`chunkSize` and `maxAllowSubsetsCount` parameters seem to control the subpacket size

dd32:
@ryelle Would you be able to double check the combined PR is correct? Visually it seems to have merged correctly.. but would like another set of eyes before I go to trunk

ryelle:
@dd32 Yep, this looks good to merge — it will need to be deployed along with parent & main theme updates: https://github.com/WordPress/wporg-parent-2021/pull/173, https://github.com/WordPress/wporg-main-2022/pull/560.

PR #686, opened 2025-05-08
Add font subsetting tool for CJK

ryelle:
See https://github.com/WordPress/wporg-mu-plugins/pull/684#pullrequestreview-2816914388 — The fonts for CJK languages can be quite large due to the characters needed, even compressed to woff2. However, they can be subset so that browsers only load the fragmented file that contains the relevant characters for a page. We have a tool already for subsetting, but it requires identifying the unicode character ranges, and that's a lot more complicated in CJK languages. Instead, @devhaozi recommended [cn-font-split](https://github.com/KonghaYao/cn-font-split/), which can split a font based on character analysis of Chinese, Japanese, and Korean text. It also has controls to optimize font fragments vs size, though they were not consistent so it was a bit trial and error to land here.

This adds a new script `bin/font-subset-cjk.js`, which like `bin/font-subset.js` runs over source font files to create subsetted fonts + a `style.css` file, which lists out the fonts and respective unicode sets. This creates around 20 subset files for each source. Then in `mu-plugins/global-fonts/style.css` we can load the generated style.css. The browser will read that, and when Noto Serif SC is called to render "初识" it will load the file with those characters in it.

I'm very surprised at how much this saves, but it makes sense. These pages are using Inter & system font fallbacks for the body font, so the only time Noto Serif is loaded is for headings.

Average font loading, tested across 9 representative pages:

| Language | Before | After | Average subset files loaded per page |
|---|---|---|---|
| Simplified Chinese | 11.12 MB | 1.44 MB | 2.7 |
| Japanese | 6.05 MB | 2.51 MB | 4.2 |
| Korean | 5.94 MB | 1.57 MB | 3.0 |

For reference, the English homepage is 192 KB, so it's still larger, but not as bad. It's possible we could improve that more by further tweaking the chunk size, but then we start seeing the subset count skyrocket… and it still needs to be maintainable.

The only downside is we can't preload the font anymore, since we don't know ahead of time what files will be used, but given the size improvement, this seems worth it.

**To test**

If you want to test the build process:

* Run `yarn`
* Add `{"type": "module"}` to package.json
* Run `yarn font-subset-cjk`

It will show output about generating the files. Some of the font files will change, and the date in the `style.css` will update. You can discard these changes. Theoretically we shouldn't need to run this very often.

To test the loading:

* Load up this PR and https://github.com/WordPress/wporg-parent-2021/pull/174
* Update the preloading code in the parent theme to prevent loading the full fonts
    ```
    if ( 'ja' === get_locale() || 'ko_KR' === get_locale()  || 'zh_CN' === get_locale() ) {
        $font_pair = [ false, false ];
    } else {
    ```
* Set your [site's language to Korean, Japanese, or Chinese (China)](https://github.com/WordPress/wporg-main-2022/?tab=readme-ov-file#working-on-non-english-sites)
* View any page, it should preload the relevant Noto Serif fragments (view source to see)
* It should not load the full Noto Serif font
* Headings should be using Noto Serif

Note that this builds on #684 & #685 so that I can subset JP, KR, and SC at once.


bin/font-subset-cjk.js
```diff
@@ -0,0 +1,43 @@
...
+	const fontName = fontNames[ i ];
+	const fontPath = path.resolve( `./fonts/${ fontName }.ttf` );
+	const inputBuffer = new Uint8Array( readFileSync( fontPath ).buffer );
+	const outputDir = `./mu-plugins/global-fonts/NotoSerif/${ fontName }`;
+
+	await fontSplit( {
+		input: inputBuffer,
+		outDir: outputDir,
+		// This seems to be a maximum, not average. Subset files won't exceed this size.
+		chunkSize: 5 * 1024 * 1024, // 10MB
```
devhaozi: `5 * 1024 * 1024` seems to be 5MB, not 10MB.

devhaozi:
Note that `cn-font-split` has not yet completed algorithm optimization for languages ​​other than Simplified Chinese, and perhaps the size can be further reduced in the future.
See https://github.com/KonghaYao/cn-font-split/issues/156

bin/font-subset-cjk.js
ryelle: I was changing this a lot trying to identify how it controlled the output, must have missed that latest change. Thanks!
```suggestion
		chunkSize: 5 * 1024 * 1024, // 5MB
```

ryelle:
Oh good to know. That's probably why Japanese had a higher rate of subset files loaded. The Korean subsetting seems to be at least a little smart; there's a hangul syllable frequency reference and it does seem to put the common hangul into one file—though it probably didn't optimize hanja.

Jiwoon-Kim:
@ryelle In Korea, most users enjoy very fast internet speeds — typically 500 Mbps to 1 Gbps on wired connections, and 5G is widely available for mobile devices. As a result, even with a modest increase in font file sizes, performance concerns would be minimal for most users here.

PR #689, opened 2025-05-12
Expand / Include all of the WordPress social media accounts

dd32:
#687 added the Mastodon account to get it verified, but we have a lot more accounts on every other platform that has been asked to be listed.

@eidolonnight @bjmcsherry have asked to increase this to all of the accounts.

These include:
[X](https://twitter.com/WordPress/)
[Bluesky](https://bsky.app/profile/wordpress.org)
[Mastodon](https://mastodon.world/@WordPress)
[Threads](https://www.threads.net/@wordpress)
[Facebook](https://www.facebook.com/WordPress/)
[Instagram](https://www.instagram.com/wordpress/)
[Linkedin](https://www.linkedin.com/company/wordpress/)
[TikTok](https://www.tiktok.com/@wordpress)
[YouTube](https://www.youtube.com/wordpress)
[Tumblr](https://wordpress.tumblr.com/)

The proposed layout would be:
> X  |  Bluesky  |  Mastodon  |  Threads  |  Facebook  |  Instagram  |  LinkedIn  |  TikTok  |  YouTube  |  Tumblr

The visual representation of these gets a bit busy, but not overly bad in my opinion.
There's some responsive work that could/should be done here, although this pre-dates this change, this just makes it a little more obvious.

<table>
<tr><td colspan=2>Screenshot 2025-05-12 at 12 24 59 pm</td></tr>
<tr><td>Screenshot 2025-05-12 at 12 26 00 pm</td><td>Screenshot 2025-05-12 at 12 26 18 pm</td></tr>
<tr><td>Screenshot 2025-05-12 at 12 26 18 pm</td><td>Screenshot 2025-05-12 at 12 26 44 pm</td></tr>
</table>

### DIFF
--- a/mu-plugins/blocks/global-header-footer/footer.php
+++ b/mu-plugins/blocks/global-header-footer/footer.php
@@ -149,12 +149,16 @@ $code_is_poetry_src = isset( $attributes['textColor'] ) && str_contains( $attrib
 
 	<!-- wp:social-links {"className":"is-style-logos-only"} -->
 	<ul class="wp-block-social-links is-style-logos-only">
-		<!-- wp:social-link {"url":"https://www.facebook.com/WordPress/","service":"facebook","label":"<?php echo esc_html_x( 'Visit our Facebook page', 'Menu item title', 'wporg' ); ?>"} /-->
 		<!-- wp:social-link {"url":"https://www.x.com/WordPress","service":"x","label":"<?php echo esc_html_x( 'Visit our X (formerly Twitter) account', 'Menu item title', 'wporg' ); ?>"} /-->
+		<!-- wp:social-link {"url":"https://bsky.app/profile/wordpress.org","service":"bluesky","label":"<?php echo esc_html_x( 'Visit our Bluesky account', 'Menu item title', 'wporg' ); ?>"} /-->
 		<!-- wp:social-link {"url":"https://mastodon.world/@WordPress","service":"mastodon","rel":"me","label":"<?php echo esc_html_x( 'Visit our Mastodon account', 'Menu item title', 'wporg' ); ?>"} /-->
+		<!-- wp:social-link {"url":"https://www.threads.net/@wordpress","service":"threads","label":"<?php echo esc_html_x( 'Visit our Threads account', 'Menu item title', 'wporg' ); ?>"} /-->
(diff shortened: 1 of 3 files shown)


eidolonnight:
> There's some responsive work that could/should be done here, although this pre-dates this change, this just makes it a little more obvious.

Can this be addressed prior to commit? Some of these layouts are pretty messy.

If not, do we need to select a subset of platforms for now?

eidolonnight (approved):
Links are all correct, and any alignment issues do not represent a regression from current behavior.

ryelle:
I happened to see this and took a stab at fixing up the alignments. It's still a little odd when just one or two items wrap, but overall better spacing & alignment across screen sizes.

| Size | Main | Rosetta |
|---|---|---|
| 450px | after-en-450 | after-es-450 |
| 600px | after-en-600 | after-es-600 |
| 960px | after-en-960 | after-es-960 |
| 1600px | after-en-1600 | after-es-1600 |

dd32:
Thank you @ryelle! You're so much better at these styles than I am! I hadn't yet attempted it, but I'm almost certain it wouldn't have looked that good without you

I had wondered if we should drop the mobile-only `(w)` and use the full `(w) WordPress.org` always such that the 450px/600px viewports are more visually balanced, so that it doesn't look like the plain WordPress logo is misaligned with the social icons.. since currently it blends in and looks like one of them.

And yes, I agree, it's a little odd looking when only some icons wrap, I was originally going to see if there was a flexbox `text-wrap: balance` that would work here, but I imagine not.

ryelle:
> since currently it blends in and looks like one of them.

Yeah, I ended up adding a larger space between so it doesn't get close enough to look like yet another social icon. The 600px screenshot is about as close as they'll get, I think.

> And yes, I agree, it's a little odd looking when only some icons wrap, I was originally going to see if there was a flexbox text-wrap: balance that would work here, but I imagine not.

I was hoping the same, but that's not how flex works unfortunately. Maybe switching to a float layout would let us use `balance`, but I'm not actually sure. And I don't think we want to deal with floats.

zokhshop11:
https://www.tiktok.com/@zokhshop_?p=N

zokhshop11:
#zokhshop #fyp #foryou #viral #joran #kekili #tiktok #tren #challenge #dance #comedy #music #love #feldakahangbarat

zokhshop11:
[https://www.tiktok.com/@zokhshop_?p=N](url)

PR #697, opened 2025-06-12
Footer: Ensure that the Code is Poetry logo is visible when Gutenberg is loaded.

dd32:
See https://meta.trac.wordpress.org/ticket/8005

### DIFF
--- a/mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
+++ b/mu-plugins/blocks/global-header-footer/postcss/footer/logos.pcss
@@ -31,6 +31,7 @@
 	& .wp-block-image {
 		text-align: left;
 		margin: 0;
+		position: unset;
 
 		& a {
 			display: block;


iamsandeepdahiya:
Thank you for working on this. Trac Ticket: https://meta.trac.wordpress.org/ticket/8005

One more doubt that I have. The page https://wordpress.org/support/view/all-topics/ doesn’t load the block editor (Gutenberg) or any post editor, because it's not a post editing screen. So, any issue due to Gutenberg loading is strange. Isn't it?

dd32:
> doesn’t load the block editor (Gutenberg) or any post editor, because it's not a post editing screen. So, any issue due to Gutenberg loading is strange. Isn't it?

It is, but it's because of the [blocks-everywhere plugin](https://wordpress.org/plugins/blocks-everywhere/) in use on the site. I assume that It's loading the JS and CSS in preparation to upgrade any textareas to an editor interface.

PR #706, opened 2025-09-09
Events: Ensure that online events never have a lat/lon returned.

dd32:
This is to ensure that it defaults to the group location.

See https://wordpress.slack.com/archives/C02RQC7RN/p1757258790929699

tldr: Online events are sometimes having their lat/lon set to the Tuvalu coordinates, instead of being set to the group location. This resolves that, while also handling the case of hybrid meetups that have an online and in-person component.


dd32:
Example.

| Before | After |
| --- | --- |
| hartford-dashboard|Screenshot 2025-09-09 at 1 33 08 pm |

PR #714, opened 2025-11-25
Allow future-scheduled posts in REST API export endpoint

dd32:
Includes future-dated posts/pages in the wporg_export content, allowing it to be sync'd to GitHub and local environments ahead of it's release.

Draft and Pending-review content is not included.

This is technically a information disclosure, because the pages/posts aren't published yet, but for the purposes of the sites that use this export functionality, having future dated items made public via local environments is not a great concern (to me).

Creating this as a PR for discussion first.


dd32:
This would be helpful for things like the upcoming release pages:
https://github.com/WordPress/wporg-main-2022/pull/581

PR #719, opened 2025-12-28
Notice block: Migrate to use inner blocks

t-hamano:
The `multiline` prop of the RichText` component is deprecated. As a result, the following warning will be displayed in the browser console:

```
installHook.js:1 wp.blockEditor.RichText multiline prop is deprecated since version 6.1 and will be removed in version 6.3. Please use nested blocks (InnerBlocks) instead. See: https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/
```

To avoid this, migrate from the RichText component to Inner Blocks.

### Testing Instructions 

- Check out the trunk branch and insert some Notice blocks with some variations.
- Check out this branch and reload your browser.
  - You should see `Updated Block: wporg/notice` in your browser console.
  - All content in the Notice block should be preserved.

### DIFF
--- a/mu-plugins/blocks/notice/src/block.json
+++ b/mu-plugins/blocks/notice/src/block.json
@@ -26,12 +26,6 @@
 			"enum": [ "alert", "info", "warning", "tip", "success" ]
 		}
 	},
-	"example": {
-		"attributes": {
-			"type": "alert",
-			"content": "<p>This is an alert notice.</p>"
-		}
-	},
 	"textdomain": "wporg",
 	"editorScript": "file:./index.js",
 	"style": "file:./style.css"
--- a/mu-plugins/blocks/notice/src/block.json
+++ b/mu-plugins/blocks/notice/src/block.json
@@ -20,12 +20,6 @@
 		}
 	},
 	"attributes": {
-		"content": {
-			"type": "string",
-			"source": "html",
-			"selector": ".wp-block-wporg-notice__content",
-			"multiline": "p"
-		},
 		"type": {
 			"type": "string",
 			"default": "tip",
@@ -34,9 +28,9 @@
 	},
 	"example": {
 		"attributes": {
-            "type": "alert",
+			"type": "alert",
 			"content": "<p>This is an alert notice.</p>"
-        }
+		}
 	},
 	"textdomain": "wporg",
 	"editorScript": "file:./index.js",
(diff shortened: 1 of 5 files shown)


mu-plugins/blocks/notice/src/block.json
```diff
@@ -34,9 +28,9 @@
 	},
 	"example": {
 		"attributes": {
-            "type": "alert",
+			"type": "alert",
 			"content": "<p>This is an alert notice.</p>"
-        }
+		}
 	},
```
ryelle: In my testing, the block example is empty now:

Image

ryelle (commented):
Thanks for tackling this! It's working well for blocks created in the editor, but another use of this block is server-side - replacing the callout shortcodes & block. Could you check that as well?

An example you could look at is [the PHP Inline Docs page](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/) (see the [shortcodes in the markdown content](https://github.com/WordPress/wpcs-docs/blob/master/inline-documentation-standards/php.md)).

The block markup is generated here:

https://github.com/WordPress/wporg-mu-plugins/blob/433e9cc12ce5394c7c74fae31bb88554b6872d98/mu-plugins/blocks/notice/index.php#L48-L71

mu-plugins/blocks/notice/src/block.json
t-hamano: Fixed in 7202287b08acd35e9020bd79749eb76595bc92ee

image

t-hamano:
@ryelle Thanks for the review!

> replacing the callout shortcodes & block. Could you check that as well?

This works just as it did before.

## Shortcode

### Editor

image

```html
<!-- wp:shortcode -->
[info]Testing info message[/info]
<!-- /wp:shortcode -->

<!-- wp:shortcode -->
[tip]Testing tip message[/tip]
<!-- /wp:shortcode -->
```

### Frontend

image

## Block

### Editor

```html
<!-- wp:wporg/callout -->
<div class="callout callout-info">
	<!-- wp:paragraph -->
	<p>Paragraph Block</p>
	<!-- /wp:paragraph -->
	<!-- wp:list -->
	<ul class="wp-block-list">
		<!-- wp:list-item -->
		<li>List Item Block</li>
		<!-- /wp:list-item -->
		<!-- wp:list-item -->
		<li>List Item Block</li>
		<!-- /wp:list-item -->
	</ul>
	<!-- /wp:list -->
	<!-- wp:paragraph -->
	<p>Paragraph Block</p>
	<!-- /wp:paragraph -->
</div>
<!-- /wp:wporg/callout -->
```

### Frontend

image

PR #720, opened 2025-12-28
Bump block API version to 3

t-hamano:
See:

- https://github.com/WordPress/gutenberg/issues/73138
- [Migrating Blocks for iframe Editor Compatibility – Block Editor Handbook | Developer.WordPress.org](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/block-migration-for-iframe-editor-compatibility/)

As you can see from the resources above, in WordPress 7.0, the post editor is planned to run as an iframe regardless of the block's API version.

As of WordPress 6.9, a browser warning will be displayed if a block with a block API version lower than 2 is registered:

```
The block "wporg/chapter-list" is registered with API version 2 or lower. This means that the post editor may work as a non-iframe editor.
Since all editors are planned to work as iframes in the future, set the `apiVersion` field to 3 and test the block inside the iframe editor.
See: https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/#version-3-wordpress-6-3
```

Therefore, we should upgrade all blocks to version 3 beforehand and test them to ensure they work correctly in the iframe editor.

### Testing Instructions

- Verify that the browser warning about the API version disappears.
- Verify that the editor is running as an iframe. In your browser console, search for the following element: `iframe[name="editor-canvas"]`
- Smoke test blocks:
  - It should work as before.
  - The styles should be applied correctly.

The API version for the following three features remains 2, but these are not blocks in the first place, and block.json is provided for convenience only, so they are not relevant to the purpose of this PR.

- wporg/horizontal-slider
- wporg/screenshot-preview
- wporg/time


ryelle:
The version warnings are gone for these blocks 👍🏻  However, I can't test the iframe yet since other blocks in the individual themes still use v2.

Are there any v2 → v3 frontend changes I should check?

t-hamano:
> However, I can't test the iframe yet since other blocks in the individual themes still use v2.

Can you tell me what those v2 blocks are? I'd love to check them out too.

Also, if you want to test these blocks in an iframe editor, you can test in the following environments, as the editor will run as an iframe in both cases. These editors are also accessible from the Classic Editor.

- Post editor > Enable Mobile or Tablet preview
- Appearance > Editor > Patterns > Create new pattern

PR #732, opened 2026-03-09
Query Filter: Guard against missing key attribute

dd32:
## Summary

- Adds an early return in the query-filter block render when the `key` attribute is not set, preventing an `Undefined array key "key"` PHP warning.
- The guard is placed before the docblock and `apply_filters` call, matching the existing early-return pattern used for missing options.

## Test plan

- Verify that query-filter blocks with a `key` attribute still render correctly.
- Verify that query-filter blocks without a `key` attribute render nothing and produce no PHP warnings.

### DIFF
--- a/mu-plugins/blocks/query-filter/render.php
+++ b/mu-plugins/blocks/query-filter/render.php
@@ -3,6 +3,10 @@
  * Render the query filter.
  */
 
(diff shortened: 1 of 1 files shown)


dd32:
This happens due to people adding it via the site editor without the key, mostly in test templates, it's easier to just not render when used incorrectly.

PR #738, opened 2026-04-15
Fix hamburger icon in local-navigation-bar — replace arrow path with …

chauhanraj754:
## Summary
Fixes the hamburger menu toggle icon in the Photo Directory 
mobile navigation bar (≤599px viewport).

## Problem
The first SVG `<path>` in the menu toggle button was drawing 
a chevron/arrow shape instead of a horizontal line, making 
the hamburger icon appear incorrectly on mobile.

## Fix
Replaced the incorrect arrow path with the correct horizontal 
line path to match the other two lines in the icon.

## Before
issue

## After
fix

## Trac ticket
https://meta.trac.wordpress.org/ticket/8227

### DIFF
--- a/mu-plugins/blocks/local-navigation-bar/index.php
+++ b/mu-plugins/blocks/local-navigation-bar/index.php
@@ -147,6 +147,8 @@ function customize_navigation_block_icon( $block_content ) {
 			)
 		)
 	) {
+		$open_icon_replaced = false;
+
 		if (
 			$tag_processor->next_tag(
 				array(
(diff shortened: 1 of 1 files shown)


coffee2code:
Thanks for the PR! You are correct about where the issue originates, but the fix isn't quite what you proposed.

The true issue is that the function (`customize_navigation_block_icon()`) didn't fully replace the hamburger menu SVG with the caret/chevron as it intended. It replaced the path for the top line of the hamburger icon with the caret. But the bottom two lines of the hamburger icon weren't removed, which caused the reported issue. Removing the paths for the rest of the original hamburger icon is the true fix here.

Fix incoming.

PR #740, opened 2026-05-13
Utilities: Add Rate_Limit helper class

dd32:
Adds a generic `wp_cache`-backed rate limiter under `WordPressdotorg\MU_Plugins\Utilities`. Consumers can use it to cap an event at N occurrences per interval without bespoke caching code.

## Why

Several spots across the WordPress.org / WordCamp ecosystem need to throttle background events (failed-API retries, abuse counters, etc.) and currently roll their own `wp_cache_add`/`wp_cache_incr` pairs. This consolidates the pattern.

## How

On WordPress.org infrastructure the cache group is registered via `wp_cache_add_no_remote_groups()` so counters stay local to each Memcached server. This keeps the limit per-DC, avoids cross-DC replication, and ensures a fault in one DC will not disable the gated behavior everywhere. The constructor flag `$no_mc_remote` defaults to true and the call is feature-detected so the class still works on environments without that drop-in.

The class is namespaced and autoloaded via the existing `Autoload\register_class_path()` registration in `loader.php` — no entry needed in the eager-require list.

## Usage

```php
use WordPressdotorg\MU_Plugins\Utilities\Rate_Limit;

$ratelimit = new Rate_Limit( 'bozo_behavior' );

if ( $ratelimit->is_ok() ) {
    // ...do the thing...
    if ( is_wp_error( $whatever ) ) {
        $ratelimit->bump();
        return;
    }
}
```

## PHP version

Targets PHP 8.1+: typed properties, `readonly $cache_key`, scalar param and return types, `private const` visibility, `int|bool` union return on `bump()`.

## Test plan

- [ ] `composer run lint` clean against the new file
- [ ] Verify counter increments and respects TTL on a wp-env install with the object-cache drop-in
- [ ] Verify the class loads via autoloader without an explicit `require_once`
- [ ] Confirm `wp_cache_add_no_remote_groups()` path is exercised on production (graceful fallback when the function is absent has been tested locally)

### DIFF
--- a/mu-plugins/utilities/class-rate-limit.php
+++ b/mu-plugins/utilities/class-rate-limit.php
@@ -1,0 +1,153 @@
+<?php
+namespace WordPressdotorg\MU_Plugins\Utilities;
+
+defined( 'WPINC' ) || die();
+
+/**
+ * Class Rate_Limit
+ *
+ * Helps limiting an event to $limit occurrences per $interval seconds, by using wp_cache.
+ *
+ * On WordPress.org's infrastructure, the cache group is registered as a "no-remote" group via
+ * `wp_cache_add_no_remote_groups()` so that counters stay local to each Memcached server. This
+ * keeps the limit per-DC, avoiding cross-DC replication and ensuring a fault on one DC doesn't
+ * disable functionality everywhere.
+ *
+ * Usage:
+ *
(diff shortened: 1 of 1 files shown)


dd32:
Note: This isn't available for WordCamp at present, due to lack of Memcache in use there.

PR #741, opened 2026-06-04
Build Tooling: Migrate from Yarn to npm

derpaschi:
Standardizes the project on a single Node package manager by replacing Yarn (Classic) with npm. The repository already shipped npm-oriented developer docs and a partly-npm `build.yml` workflow, so this removes the remaining Yarn usage and the mixed-tooling ambiguity.

## Why

The project carried two package managers at once: a `yarn.lock`, `yarn`-based CI in `linters.yml`/`tests.yml`, and a `setup:tools` script invoking `yarn install` — while the README documented `npm run …` and `build.yml` already used `npm install`. A single package manager removes the lockfile-drift risk of two resolvers, matches the existing documentation, and lowers the barrier for new contributors.

## What changes

- Remove `yarn.lock`; add `package-lock.json` (generated with Node 20 / npm 10, per `.nvmrc`).
- `package.json`: `setup:tools` now runs `npm install`; add a `preinstall` guard (`npx --yes only-allow npm`) so the project can't be installed with Yarn/pnpm again.
- CI (`build.yml`, `linters.yml`, `tests.yml`): switch `cache: yarn` → `cache: npm` and replace `yarn …` invocations with their `npm run …` / `npx` equivalents.
- `bin/font-subset-cjk.js`: update the `yarn` references in the usage comments to npm.
- `README.md`: drop the documented `npm run build:rtl` command, which has no matching script (`npm run build` already produces `style-rtl.css`).

No application/runtime code is touched, and no dependency versions change: a fresh npm resolve produces the same lint-relevant tree (`@wordpress/eslint-plugin@18.1.0`, `prettier@3.0.3`) as the previous `yarn.lock`.

## Test plan

- [x] `npm install` succeeds and generates `package-lock.json`; the `preinstall` guard blocks `yarn install`
- [x] `npm run build` builds all block styles including RTL
- [x] `npm run lint:css` passes
- [x] `npm run lint:js` shows only pre-existing findings (verified identical to a `yarn lint:js` run on `trunk`)
- [x] CI green on this PR (lint + tests workflows on the npm path)


obenland:
Suggestion (non-blocking): consider `npm ci` instead of `npm install` in CI.

`tests.yml` and `build.yml` both run `npm install`, which can silently mutate the now-committed `package-lock.json`. With a lockfile present, `npm ci` is the standard CI choice — it's reproducible and fails loudly on lockfile drift rather than rewriting it.

(`setup:tools` should stay `npm install` since it's also run locally for first-time setup.)

obenland:
Suggestion (non-blocking): redundant install in `tests.yml`.

The job runs `npm run setup:tools` (which already runs `npm install`) and then `npm install` again as a separate step. This is a pre-existing pattern carried over from the Yarn setup — not a regression — but the duplicate install could be collapsed now that you're touching this.

PR #751, opened 2026-08-07
Global header/footer: Restore Swag to the footer, tidy the header link

j111q:
## What & why

The [Mercantile swag store](https://mercantile.wordpress.org/) has reopened, and Matt asked that it be linked from the main .org navigation. It's already reachable in the header's About menu (last changed in #739); this PR tidies that link and restores it to the footer. Both links open the store in a **new tab** — it's a separate site — with screen-reader-only "opens in a new tab" text for accessibility.

### Footer — restore Swag to the "Get Involved" column

Swag used to hold the footer's **4th column, 4th item**. The history:

- **#615** removed it from header and footer because the store was closed (#614 — *"to reduce inbound traffic"* while it had *"no plans to reopen"*).
- That left, in the author's words, *"a hole in the footer links."*
- **#619** filled the hole with **Five for the Future**, explicitly *"to replace the Swag store links which have been removed while the store is closed."*

The store has reopened, so this reverses that swap. **Five for the Future keeps its own place in the header's About menu.** Swag carries the `↗` to match the store link in the header's About menu and flag the jump to the separate store.

| Get Involved (before) | Get Involved (after) |
|---|---|
| Get Involved | Get Involved |
| Events | Events |
| Donate ↗ | Donate ↗ |
| **Five for the Future** | **Swag ↗** |

### Header — promoted out of the About menu

- Renamed "Swag Store ↗" → "Swag ↗": shorter, matches the `wordpress.org/about/swag` redirect and the store's own name, and consistent with the other external dropdown links (Job Board ↗, Openverse ↗).
- `Swag Store` was under `About`, but with this PR, `Swag` will sit next to it at the top-level menu

## Notes

- We are promoting `Swag` to a top-level menu item after `About`. BUT, I do want to acknowledge that a top-level item should _ideally_ behave like its peers — shared header, an active/selected state on the destination — and the store right now is a separate site with its own chrome and no global nav. Doing that properly means getting Mercantile to adopt the global header, plus a more holistic look at the global nav to include the store. I will take a look at that for a better long-term solution, outside this PR. 

- Separately, [jobs.wordpress.net](https://jobs.wordpress.net/) was flagged as another surface whose footer could be aligned with the global footer; out of scope here.

### DIFF
--- a/mu-plugins/blocks/global-header-footer/blocks.php
+++ b/mu-plugins/blocks/global-header-footer/blocks.php
@@ -608,12 +608,12 @@ function get_global_menu_items() {
 					'url'   => 'https://jobs.wordpress.net/',
 					'type'  => 'custom',
 				),
-				array(
-					'title' => esc_html_x( 'Swag Store ↗︎', 'Menu item title', 'wporg' ),
-					'url'   => 'https://mercantile.wordpress.org/',
-					'type'  => 'custom',
-				),
 			),
+		),
+		array(
+			'title' => esc_html_x( 'Swag ↗︎', 'Menu item title', 'wporg' ),
+			'url'   => 'https://mercantile.wordpress.org/',
+			'type'  => 'custom',
 		),
 	);
 
--- a/mu-plugins/blocks/global-header-footer/footer.php
+++ b/mu-plugins/blocks/global-header-footer/footer.php
@@ -92,7 +92,7 @@ $code_is_poetry_src = isset( $attributes['textColor'] ) && str_contains( $attrib
 		<li><a href="https://wordpressfoundation.org/donate/"><?php echo esc_html_x( 'Donate ↗', 'Menu item title', 'wporg' ); ?></a></li>
 		<!-- /wp:list-item -->
 		<!-- wp:list-item -->
-		<li><a href="https://wordpress.org/five-for-the-future/"><?php echo esc_html_x( 'Five for the Future', 'Menu item title', 'wporg' ); ?></a></li>
+		<li><a href="https://mercantile.wordpress.org/"><?php echo esc_html_x( 'Swag ↗', 'Menu item title', 'wporg' ); ?></a></li>
 		<!-- /wp:list-item -->
 	</ul>
 	<!-- /wp:list -->


mu-plugins/blocks/global-header-footer/blocks.php
```diff
@@ -609,9 +609,10 @@ function get_global_menu_items() {
 					'type'  => 'custom',
 				),
 				array(
-					'title' => esc_html_x( 'Swag Store ↗︎', 'Menu item title', 'wporg' ),
-					'url'   => 'https://mercantile.wordpress.org/',
-					'type'  => 'custom',
+					'title'         => esc_html_x( 'Swag ↗︎', 'Menu item title', 'wporg' ),
+					'url'           => 'https://mercantile.wordpress.org/',
+					'type'          => 'custom',
+					'opensInNewTab' => true,
```
obenland: Let's have it open in the existing tab. No other link in header or footer currently opens in a new tab as it's an a11y best practice to avoid that.

mu-plugins/blocks/global-header-footer/header.php
```diff
@@ -36,13 +36,24 @@ function recursive_menu( $menu_item, $top_level = true ) {
 	$has_submenu = ! empty( $menu_item['submenu'] );
 
 	if ( ! $has_submenu ) {
+		$label            = $menu_item['title'];
+		$opens_in_new_tab = '';
+
+		if ( ! empty( $menu_item['opensInNewTab'] ) ) {
+			$opens_in_new_tab = ',"opensInNewTab":true';
+			// The visible "↗" is hidden from assistive tech by `add_aria_hidden_to_arrows()`,
+			// so announce the new tab with screen-reader-only text instead.
```
obenland: This will probably be removed with the removal of the open-in-new-tab functionality, but if it ever comes up again, let's stick with [php docs standards for multiline comments](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments) or limit comments to a single line.

obenland (changes requested):

mu-plugins/blocks/global-header-footer/blocks.php
j111q: Changed this 👍

mu-plugins/blocks/global-header-footer/header.php
j111q: Oh great to know for the future, thank you! Yes this has been dropped.

j111q:
Thank you for review @obenland! Updated:
- `Swag` now opens in the same tab (no more new-tab behavior in header and footer), 
- And moved `Swag` to a top-level item at the end of the row, after `About` (per convo with @richtabor @fcoveram ) 

Ready for another look when you are able 🙏

obenland (approved):

PR #769, opened 2026-09-04
Meetup OAuth: take the authorization code from the caller

bor0:
`get_oauth_token()` read `$_GET['code']` itself. It now takes the code as a parameter and reads nothing from the request — an authorization code arrives on a redirect, so the caller that handled that redirect is the one positioned to say it belongs to an authorization it started. The site option fallback is unchanged.

The reconnection notice can't build a usable authorize link for the same reason, so the URL it prints is filterable via `meetup_oauth2_authorize_url` and no longer carries a `state`.

This is the "separate coordinated change" #775 left a note for when it wrapped that block in `phpcs:disable WordPress.Security.NonceVerification.Recommended`. With the request read gone the suppression is no longer needed, so it goes too.

`wcorg-meetup-oauth.php` in WordPress/wordcamp.org is the only consumer of this flow and pairs with this change: WordPress/wordcamp.org#2002. That side queues the code in the site option as well as passing it, and the client reads that option the same way before and after this change, so the two can land in either order.

Note this also covers `wcpt`'s `update_meetup_data()`, which instantiates `Meetup_Client` during `save_post` and could previously pick a code out of the request on any admin page load.

### Testing

`phpunit/test-meetup-oauth2-client.php` pins where the code comes from: a code on the request is never spent, a caller-supplied one is used and beats the stored option, and the site option remains the fallback when the caller passes nothing. Token requests are intercepted through `pre_http_request`, so nothing leaves the process.

- `run-tests` green: `OK (53 tests, 147 assertions)`.
- `phpcs` clean on both files against trunk's current ruleset.
- `Meetup_Client::get_oauth_token()` and `reset_oauth_token()` both call with no arguments and fall through to the site option exactly as before.

https://claude.ai/code/session_013V9sVz7atsstmmD8gN39Df

### DIFF
--- a/mu-plugins/utilities/class-meetup-oauth2-client.php
+++ b/mu-plugins/utilities/class-meetup-oauth2-client.php
@@ -207,9 +207,13 @@ class Meetup_OAuth2_Client extends API_Client {
 	 *
 	 * This stores a successfully retrieved token array to the database for repeated use, until it expires.
 	 *
+	 * @param string $auth_code Optional. An authorization code that the caller has already established
+	 *                          belongs to an authorization it started itself. Default empty, which falls
+	 *                          back to the code stored in the site option.
+	 *
 	 * @return string
 	 */
-	public function get_oauth_token() {
+	public function get_oauth_token( $auth_code = '' ) {
 		if ( $this->oauth_token && ! $this->is_expired_token( $this->oauth_token ) ) {
 			return $this->oauth_token['access_token'];
 		}
@@ -226,15 +230,13 @@ class Meetup_OAuth2_Client extends API_Client {
 
 		// If it's not a valid token, or the refresh token wasn't valid, check to see if we're able to fetch a new one.
 		if ( ! $valid ) {
-			$auth_code = get_site_option( self::SITE_OPTION_KEY_AUTHORIZATION, false );
-
-			// The token is stored network-wide, so binding it is a network administrator's call.
-			if ( isset( $_GET['code'], $_GET['state'] )
-				&& 'meetup-oauth' === $_GET['state']
-				&& is_admin()
-				&& current_user_can( 'manage_network_options' )
-			) {
-				$auth_code = sanitize_text_field( wp_unslash( $_GET['code'] ) );
+			/*
+			 * Nothing is read from the request here. An authorization code arrives on a redirect from
+			 * Meetup, which can't carry a nonce, so only a caller that minted and matched a `state` of its
+			 * own is in a position to say the code belongs to an authorization it started.
+			 */
+			if ( ! is_string( $auth_code ) || ! $auth_code ) {
+				$auth_code = get_site_option( self::SITE_OPTION_KEY_AUTHORIZATION, false );
 			}
 
 			if ( $auth_code ) {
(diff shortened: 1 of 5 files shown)


mu-plugins/utilities/class-meetup-oauth2-client.php
```diff
@@ -226,15 +230,13 @@ public function get_oauth_token() {
...
-				&& 'meetup-oauth' === $_GET['state']
-				&& is_admin()
-				&& current_user_can( 'manage_network_options' )
-			) {
-				$auth_code = sanitize_text_field( wp_unslash( $_GET['code'] ) );
+			/*
+			 * Nothing is read from the request here. An authorization code arrives on a redirect from
+			 * Meetup, which can't carry a nonce, so only a caller that minted and matched a `state` of its
+			 * own is in a position to say the code belongs to an authorization it started.
+			 */
```
obenland:
```suggestion
```
I'd probably just remove that.

mu-plugins/utilities/class-meetup-oauth2-client.php
```diff
@@ -248,15 +250,33 @@ public function get_oauth_token() {
...
+				add_query_arg(
+					array(
+						// `add_query_arg()` doesn't encode, so the values are encoded on the way in.
+						'client_id'     => rawurlencode( self::CONSUMER_KEY ),
+						'response_type' => 'code',
+						'redirect_uri'  => rawurlencode( self::REDIRECT_URI ),
+					),
+					self::URL_AUTHORIZE
+				)
+			);
```
obenland: I love the hook docs, thanks. Maybe the value could be separated out ahead of the filter, to make it easier to see what's going on when reading the code.

obenland (commented):
Thanks, I only have minor formatting suggestions.

mu-plugins/utilities/class-meetup-oauth2-client.php
bor0: Removed in da458ae.

mu-plugins/utilities/class-meetup-oauth2-client.php
bor0: Done in da458ae — `$authorize_url` is built first, then passed through the filter on its own line.

bor0:
Both suggestions applied in da458ae, thanks.

Two other things since your review:

- Added `phpunit/test-meetup-oauth2-client.php` covering where the code comes from — request ignored, caller-supplied used, caller wins over the stored option, option still the fallback. My earlier claim in the description that nothing covered `utilities/` was simply wrong; I'd checked the wrong directory.
- The PR was open rather than draft when you reviewed it — that was unintended on my part. It's a draft now until `run-tests` is green.

Current CI state: the last `run-tests` failure is infrastructure, not the change — the wp-env image build 404'd fetching `sudo` from `deb.debian.org` before any test ran. The run before it caught a real path bug in my test file (wp-env mounts `mu-plugins` and `phpunit` at different roots), fixed in fc8b41f. So the new test still hasn't actually executed anywhere yet — I'll re-run once the mirror recovers and won't call it verified until it passes.

bor0:
**On the red CI — it isn't this branch.**

I re-ran run `33866556101`, which is commit `f127e36` on this branch *before* the test file existed and which passed on 2026-09-04. Re-run today it fails identically:

```
E: Failed to fetch http://deb.debian.org/debian-security/pool/updates/main/s/sudo/sudo_1.9.5p2-3+deb11u4_amd64.deb  404  Not Found
target tests-wordpress: failed to solve: process "/bin/sh -c apt-get -qy install sudo" did not complete successfully: exit code: 100
```

Same code, same step (`Build blocks and start the WordPress environment`), green three days ago. Debian has superseded `deb11u4` in bullseye-security and `Tests-WordPress.Dockerfile:31` runs `apt-get -qy install sudo` against a stale index with no preceding `apt-get update`, so it asks for a package no longer in the pool. That will hit every PR in this repo, not just this one.

**On the new test**, since CI can't run it: I executed it against this branch's actual class in the wordcamp.org phpunit environment — 4 tests, 6 assertions, all passing. That validates the interception, the reflection reset and the assertions, but it ran on PHP 8.4 / multisite rather than the 7.4 / single-site this repo pins, so it is not a substitute for `run-tests` here. I've kept it 7.4-compatible (no `??`, no arrow functions, explicit `setAccessible()`), but I'd treat it as verified only once CI can actually build again.

Ready for review otherwise.

bor0:
### Status

Both of your suggestions are applied in `da458ae` — the explanatory block above the fallback is gone, and `$authorize_url` is now built before the `apply_filters()` call rather than inline.

Rebuilt on current `trunk` (2026-09-17). Conflicts came from #770 and #775 landing in the meantime:

- The CI-fix commits this branch had merged in are now on `trunk` via #770, so they've dropped out of the diff. It's back to just the two Meetup files.
- #775 wrapped the `$_GET` read in `phpcs:disable WordPress.Security.NonceVerification.Recommended` with a note that nonce hardening needed a separate coordinated change. This is that change, so the read and its suppression both go.
- Took `trunk`'s version of `test-encryption.php` wholesale — it already carries the two fixes that rode along here, and has since renamed the key helper.
- One `phpcbf` fixup for the tightened multi-line call-signature sniff.

`run-tests` is green: `OK (53 tests, 147 assertions)`, including the four new ones. Earlier I flagged the new test as provisional because it had only ever run in the wordcamp.org environment — that no longer applies, it now runs here.

*(I'll keep editing this comment rather than adding new ones.)*

obenland (approved):

PR #771, opened 2026-09-08
Modal: expand shortcodes in the label before escaping

mcliwanow:
The download button on https://wordpress.org/download/ has read "Download WordPress [latest_version]" since September 4 (Meta Trac #8143, reopened: https://meta.trac.wordpress.org/ticket/8143).

The download pattern carries that shortcode in the block's `label` attribute. Since wporg-parent-2021 moved pattern shortcodes from the rendered output to the pattern source (WordPress/wporg-parent-2021#183 and #184), block attributes are not reached anymore: `do_shortcode()` skips HTML comments, and block delimiters are HTML comments. The `href` attribute kept working only because #755 expands it in `render.php` itself.

This does the same for `label`. Expand it, drop anything the expansion emitted so `the_content` can't run it a second time, then `wp_kses_post()` and `esc_attr()` as before. The dialog's `aria-label` from #767 reads the same attribute, so it picks up the expanded value too.

## Tests

New `phpunit/test-modal.php` renders the block through `do_blocks()` and checks the button text, the `aria-label`, the `href`, that markup in the label is still filtered after the pass, and that a shortcode emitted by the expansion is dropped. On trunk all four fail, on this branch the suite is 30/30.

I grepped the block delimiters in wporg-main-2022 and in the themes in the wordpress.org repo for shortcodes. The download pattern's modal label is the only one, so nothing else is waiting on this.

## Testing it in a browser

The bug only shows when the modal is rendered from a template or a pattern. In post content `the_content` runs `do_shortcode()` over the output and hides it, so the page has to come from a block template. With `wp-env start` running from this repo:

1. Stand in for the theme shortcodes with a throwaway mu-plugin (don't commit it):

    ```php
    // mu-plugins/zz-demo-shortcodes.php
    <?php
    add_shortcode( 'latest_version', fn() => '7.1' );
    add_shortcode( 'download_link', fn() => 'https://wordpress.org/latest.zip' );
    ```

2. Create a `page-download` template holding the modal, and a page that uses it:

    ```sh
    npx wp-env run cli wp post create --post_type=wp_template --post_status=publish --post_name=page-download --post_title='Page: Download' --porcelain --post_content='<!-- wp:wporg/modal {"href":"[download_link]","label":"Download WordPress [latest_version]"} --><!-- wp:paragraph --><p>Howdy!</p><!-- /wp:paragraph --><!-- /wp:wporg/modal -->'
    npx wp-env run cli wp post term set <template id> wp_theme twentytwentyfive
    npx wp-env run cli wp post create --post_type=page --post_status=publish --post_name=download --post_title=Download --porcelain
    ```

3. Open `http://localhost:8888/?page_id=<page id>`. On trunk the button reads "Download WordPress [latest_version]", on this branch "Download WordPress 7.1". The `aria-label` on the `role="dialog"` element changes the same way.

If you swap `render.php` between trunk and the branch with the env running, give opcache a couple of seconds before reloading, it served me the old file once.

### DIFF
--- a/mu-plugins/blocks/modal/render.php
+++ b/mu-plugins/blocks/modal/render.php
@@ -14,7 +14,8 @@ $style .= get_style_decl_from_attr( $attributes, 'overlayColor' );
 $style .= get_style_decl_from_attr( $attributes, 'closeButtonColor' );
 
 // Expand first, then drop any shortcode the expansion emitted; it would otherwise expand again later.
-$href = strip_shortcodes( do_shortcode( $attributes['href'] ?? '' ) );
+$href  = strip_shortcodes( do_shortcode( $attributes['href'] ?? '' ) );
+$label = strip_shortcodes( do_shortcode( $attributes['label'] ) );
 
 $button_class = 'wp-block-button';
 if ( ! empty( $attributes['buttonStyle'] ) ) {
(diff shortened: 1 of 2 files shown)


pkevan (approved):

nikvalani96:
Thank you for fixing and deploying this! I’ve verified that the button now displays “Download WordPress 7.1” correctly. 🙌

PR #777, opened 2026-09-16
Global Footer: tidy the logo, trademark and social links layout

jamesckemp:
Tidies the bottom of the global footer so the logo, trademark notice and social links read as one block, and gives "Code is Poetry" its own strip at the very bottom.

## What changed

- **Code is Poetry** moves into a full-width strip at the bottom of the footer, centred, in a darker shade of the footer colour (the same `nero` used for header hovers) and matching the global header's 60px height. The graphic is now inlined from a single SVG that inherits the light text colour, replacing the two fixed-colour variants and the dependency on s.w.org. The translated text fallback used on non-English locales drops from 20px to 18px so its cap height matches the graphic.
- **Trademark notice** is now a paragraph block that sits 4px below the WordPress.org logo. It uses the footer link size (14px from the tablet breakpoint, 16px below) and the existing `--wp-global-header--text-color--light` colour, which is also what Code is Poetry now uses, so the two match and both adapt to light and dark footer backgrounds. That colour is charcoal-5 on the dark footer, 5.27:1 against the footer and 5.82:1 against the strip.
- **Social icons** drop from 28px to 24px with the row gap halved to 10px. Between the tablet and desktop breakpoints (890 to 1151px) they split into two even rows of five, centred on the logo block, instead of leaving a single icon orphaned. Above 1152px they sit on one row, centred on the logo. Below 890px they flow naturally.
- **Spacing above the logo row** is 44px on mobile, matching the footer's top padding, and 60px from the tablet breakpoint.
- **Full WordPress.org logo** is used at every width. The small W mark was only needed when the row also had to fit the centred Code is Poetry image, and the mark is still used by the header and local navigation bar.

No new breakpoints: everything uses the shared `--tablet` and `--desktop` custom media.

## Screenshots

| Width | Before | After |
| --- | --- | --- |
| 390px | before-390 | after-390 |
| 1000px | before-1000 | after-1000 |
| 1280px | before-1280 | after-1280 |

## Testing

- `npm run lint:css`, `composer run lint` and `npm run build` pass.
- Checked on a local plugin directory site at 390, 459, 480, 600, 889, 890, 1000, 1125, 1152, 1280 and 1396px.
- The white footer variant and the translated text fallback (Spanish and Japanese) were checked by emulating their classes and markup in the browser. The white variant inherits the charcoal text colour for the graphic.

## Notes for deploy

The footer markup changed, so the Trac and Codex templates will need re-syncing after this ships (see the block README).


obenland:
How would you feel about giving the Code is Poetry its own paragraph and keep it centered between the links and the WordPress logo? It would be a smaller visual change while keeping the alignment and spacing fixes

jamesckemp:
@obenland the main problem I found with that was that it felt very off balance in practise. The other option I considered was pulling it onto its own line like so

image

(I think we might want to increase the brightness of it in this scenario)

obenland:
Oh, because it would sit above this gap between logo and social links. I like your idea about pulling it into its own line

jamesckemp:
@obenland I just pushed an update based on our convo above and updated the screenshots in the PR description. Is this looking more appropriate? It also accounts for the translated "Code is Poetry" text for non-"en" languages.

obenland:
Sorry, this is getting nit-picky, just curious how strongly you feel about the dividing line between the links and the logos. Might be cleaner to leave it off?

jamesckemp:
@obenland I like nitpicky! Not overly opinionated on it either way. Let me share the screenshots of what it'd look like before updating the PR

image

image

image

What do you think?

obenland:
Yeah I'd go without the line

jamesckemp:
@obenland PR updated, and screenshots refreshed in the main description!

jamesckemp:
Hi @obenland, just wanted to check in on this one. Anything else you feel needs amending?

